PR #8, opened 2022-07-13
Feature/db health check

mhuseinov:
- typerom 
  - orm.config.ts
  - readme.md
- health check endpoint
  - db read/write
  - api
  - healthcheck migration
- verbose flag for Nestjs


alcs-api/src/main.ts
```diff
@@ -12,7 +12,7 @@ async function bootstrap() {
   // fastify
   const app = await NestFactory.create<NestFastifyApplication>(
     AppModule,
-    new FastifyAdapter({ logger: true }),
+    new FastifyAdapter({ logger: JSON.parse(config.get<string>('VERBOSE')) }),
```
dhaselhan, line 15: JSON.parse here seems a little odd, is there a reason this isn't a boolean in the config and get<boolean> here?

alcs-api/src/healthcheck/healthcheck.entity.ts
```diff
@@ -0,0 +1,10 @@
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+
+@Entity()
+export class HealthCheck {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column({ default: BigInt(new Date().getTime()).toString(), type: 'bigint' })
+  UpdateDate: string;
```
dhaselhan, line 9: Should this be updateDate instead of UpdateDate?

alcs-api/src/main.ts
mhuseinov: initially I had issue with boolean, even like this:  config.get<boolean>('VERBOSE'). Anyway, it is not relevant anymore, since I've changed it to LOG_LEVEL instead

dhaselhan (approved):
LGTM 👍

PR #11, opened 2022-07-15
Add Keycloak AuthGuard

dhaselhan:
* Add user entity to back it
* Add tests and config

### DIFF
--- a/alcs-api/src/common/authorization/role.guard.spec.ts
+++ b/alcs-api/src/common/authorization/role.guard.spec.ts
@@ -1,0 +1,143 @@
+import { ExecutionContext } from '@nestjs/common';
+import { Reflector } from '@nestjs/core';
+import { Test, TestingModule } from '@nestjs/testing';
+import { createMock } from '@golevelup/nestjs-testing';
+import {
+  KEYCLOAK_CONNECT_OPTIONS,
+  KEYCLOAK_INSTANCE,
+  KEYCLOAK_LOGGER,
+} from 'nest-keycloak-connect';
+import { KeycloakMultiTenantService } from 'nest-keycloak-connect/services/keycloak-multitenant.service';
+import { UserService } from '../../user/user.service';
+import { RoleGuard } from './role.guard';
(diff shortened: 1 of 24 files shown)


alcs-api/src/common/authorization/role.guard.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+        user: {
+          email: 'test@test.com',
+          email_verified: false,
+        },
+      }),
+    } as any);
+
+    guard.keyCloakGuard = createMock<KeyCloakRoleGuard>();
+    guard.keyCloakGuard.canActivate = jest.fn(async () => true);
+    reflector.get = jest.fn((): any => ['admin']);
```
mhuseinov, line 81: is it possible to declare some constant or enum in common for 'admin'?

alcs-api/src/common/authorization/role.guard.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+  it('should accept if users has any of the roles', async () => {
+    const mockContext = createMock<ExecutionContext>();
+    mockContext.switchToHttp.mockReturnValue({
+      getRequest: () => ({
+        user: {
+          email: 'test@test.com',
+          email_verified: true,
+        },
+      }),
+    } as any);
```
mhuseinov, line 115: is it worth moving this into private method in scope of this file to avoid code climate complains of to lengthy functions?

alcs-api/src/common/authorization/role.guard.spec.ts
dhaselhan: Great idea, Ill make an enum for AUTH_ROLES in the AuthGuard :)

alcs-api/src/common/authorization/role.guard.spec.ts
dhaselhan: Moved this into a const and re-used it across a few tests to DRY up the code

mhuseinov (approved):

PR #13, opened 2022-07-15
Application, ApplicationStatus API

mhuseinov:
- Application CRUD API
- ApplicationStatus CRUD API
- unit tests
- coverage configuration improvement


alcs-api/src/application/application.controller.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    applicationService = module.get<ApplicationService>(ApplicationService);
+    controller = module.get<ApplicationController>(ApplicationController);
+  });
+
+  it('should be defined', () => {
+    expect(controller).toBeDefined();
+  });
+
+  it('can delete', async () => {
```
dhaselhan, line 52: Discussion: Should all test cases start with should? As in should call applicationService delete when deleting or something?

alcs-api/src/common/utils/test-helpers/mockTypes.ts
```diff
@@ -30,9 +33,13 @@ export const mockKeyCloakProviders = [
   {
     provide: KEYCLOAK_LOGGER,
     useValue: {
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
```
dhaselhan, line 36: Good catch! I have now remembered to turn on eslint

dhaselhan (approved):
LGTM 👍

PR #14, opened 2022-07-18
Add skeleton login system

dhaselhan:


alcs-frontend/src/app/shared/footer/footer.component.html
```diff
@@ -0,0 +1 @@
+<p>©BC Government 2022</p>
```
mhuseinov, line 1: maybe something like  new Date().getFullYear()

dhaselhan: Done changed to be dynamic

mhuseinov (approved):
🔥

PR #16, opened 2022-07-19
Add Board + Backing Changes

dhaselhan:
* Add board component from POC
* Add update method to AppService
* Combine ApplicationServiceModule and ApplicationModule


alcs-api/src/application/application-status/application-status.service.ts
```diff
@@ -31,6 +31,15 @@ export class ApplicationStatusService {
     return await this.applicationStatusRepository.save(applicationEntity);
   }
 
+  async fetchStatusId(statusCode: string) {
```
mhuseinov, line 34: lets return the whole object here, so if ever need other properties we have everything in place

alcs-api/src/application/application.controller.ts
```diff
@@ -1,48 +1,87 @@
...
+    };
+  }
+
+  @Patch()
+  async update(
+    @Body() application: ApplicationPartialDto,
   ): Promise<ApplicationDto> {
-    const app = await this.applicationService.createOrUpdate(application);
+    let statusId;
+    if (application.status) {
```
mhuseinov, line 47: do we want to check if application status changed?

alcs-frontend/src/app/features/admin/admin.component.ts
```diff
@@ -1,15 +1,67 @@
...
   }
 
+  onDropped($event: { id: string; status: string }) {
+    this.applicationService
+      .updateApplication({
+        fileNumber: $event.id,
+        status: $event.status,
+      })
+      .then((r) => {
+        //TODO: Move this to a toast
```
mhuseinov, line 55: 👍

alcs-frontend/src/app/shared/drag-drop-board/drag-drop-item.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface DragDropItem {
```
mhuseinov, line 1: not in scope of this PR, I wonder if we will need to capture who moved item between columns?

dhaselhan: I think we will, having a last updated by is useful for Applications. We should check with product if they want a full history a la Jira style or just tracking the last person to make changes.

alcs-api/src/application/application.service.ts
```diff
@@ -17,37 +17,52 @@ export class ApplicationService {
...
     // TODO replace with AutoMapper
     applicationEntity = applicationEntity ?? new Application();
-    applicationEntity.number = application.number;
+    applicationEntity.fileNumber = application.fileNumber;
     applicationEntity.title = application.title;
     applicationEntity.body = application.body;
 
-    if (application.statusId) {
-      applicationEntity.statusId = application.statusId;
+    return await this.applicationRepository.save(applicationEntity);
```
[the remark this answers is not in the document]
dhaselhan: Ill move the check to the controller and make it createOrUpdate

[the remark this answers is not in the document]
dhaselhan: Done!

alcs-api/src/application/application-status/application-status.service.ts
dhaselhan: Done!

alcs-api/src/application/application.controller.ts
dhaselhan: Maybe? I'm not quite sure what you are suggesting. Like throw an error if it's the same status?

mhuseinov: I mean like skip the status update flow if it is the same

alcs-api/src/main.ts
```diff
@@ -55,14 +54,17 @@ const registerGlobalFilters = (app: NestFastifyApplication) => {
 };
 
 const registerCors = (app: NestFastifyApplication) => {
-  app.enableCors({
-    origin: [
-      config.get<string>('BASE_URL'),
-      config.get<string>('KEYCLOAK.AUTH_SERVER'),
-    ],
-  });
+  app.enableCors();
```
mhuseinov, line 57: my guess that this will break swagger authorization

alcs-api/src/application/application.controller.ts
dhaselhan: ah yes! Ill add that in

alcs-api/src/main.ts
dhaselhan: Fixed, add FRONTEND_ROOT to config and added these back

alcs-api/src/application/application.controller.ts
dhaselhan: Done!

mhuseinov (approved):
💯

PR #22, opened 2022-07-22
Auth Updates

dhaselhan:
* Switch to offline verification
* Verify token before sending user to authorized frontend
* Add Redis Service + Redis to docker compose
* Change auth guard to use roles from Keycloak
* Add in programmed KeyCloak Roles


alcs-api/docker-compose.yml
```diff
@@ -26,6 +27,7 @@ services:
       - postgres:/var/lib/postgres
     networks: 
       - alcs
+
   keycloak:
```
mhuseinov, line 31: since we switched to government sso do we need this local keycloak?

mhuseinov (approved):

alcs-api/docker-compose.yml
dhaselhan: Nope, I am happy to remove it

dhaselhan: Done! Also removed admin-db

PR #23, opened 2022-07-25
Feature/expanded card view

mhuseinov:
- added veiw-card component
- updated user entity
- added assignee on application
- unit tests for api

PS:
view card is WIP, will be implemented in next mrs


alcs-api/src/common/authorization/authorization.service.ts
```diff
@@ -86,4 +108,34 @@ export class AuthorizationService {
...
+      email: user.email,
+      name: user.name,
+      displayName: user.display_name,
+      identityProvider: user.identity_provider,
+      preferredUsername: user.preferred_username,
+      givenName: user.given_name,
+      familyName: user.family_name,
+      idirUserGuid: user.idir_user_guid,
+      idirUserName: user.idir_username,
+    } as CreateOrUpdateUserDto;
```
dhaselhan, line 126: You shouldn't need the as here if you are mapping correctly, this is kinda forcing it

alcs-api/src/common/authorization/authorization.service.ts
```diff
@@ -86,4 +108,34 @@ export class AuthorizationService {
...
+  }
+
+  private async registerUser(payload: Buffer) {
+    const decodedToken = JSON.parse(
+      Buffer.from(payload).toString(),
+    ) as UserFromToken;
+    this.logger.debug(decodedToken);
```
dhaselhan, line 133: Do you want to leave this debug in?

alcs-api/src/user/user.controller.ts
```diff
@@ -26,13 +26,6 @@ export class UserController {
     return this.userService.createUser(dto);
```
dhaselhan, line 26: Should we also delete create user since it should only happen from SSO now?

alcs-api/src/common/authorization/authorization.service.ts
mhuseinov: yep, I assume we will have 'error' or 'warning' level on deployed env so this should not cause any issues

alcs-api/src/user/user.controller.ts
mhuseinov: that is a good question. We do not have any requirements on supporting something other than SSO so I think yes. I can get rid of it in my next pr

alcs-api/src/common/authorization/authorization.service.ts
mhuseinov: the difference is in CamelCase and SnakeCase. after casting object remained empty (only default fields populated)

alcs-api/src/common/authorization/authorization.service.ts
dhaselhan: Sounds good!

alcs-api/src/user/user.controller.ts
dhaselhan: Sounds good

alcs-api/src/common/authorization/authorization.service.ts
mhuseinov: done

dhaselhan (approved):
LGTM 👍

PR #24, opened 2022-07-25
Add History to Application, improve Board UI

dhaselhan:
* Add migration to create initial statuses for applications
* Add nest-cls so we can use request scoped storage easily to track user
* Add subscriber to listen to Application changes and create a history entity when status changes
* Add tests for supporting classes
* Add required roles to User Controller


alcs-api/src/application/application.entity.ts
```diff
@@ -1,23 +1,23 @@
...
   title: string;
 
   @Column()
   body: string;
 
   @Column({
     type: 'uuid',
-    nullable: false,
-    default: 'e0083fa2-9457-433b-b711-9344e1e3fd48',
+    default: 'e6ddd1af-1cb9-4e45-962a-92e8d532b149',
```
mhuseinov, line 20: I thinks we've updated this value 3 times already 😄 Migrations...

alcs-api/src/common/entities/base.entity.ts
```diff
@@ -12,14 +12,12 @@ import {
 
 export const getTimestampColumnsOptions = (): ColumnOptions => {
   return {
-    type: 'timestamp',
+    type: 'timestamptz',
```
mhuseinov, line 15: 👍

alcs-api/src/main.ts
```diff
@@ -78,6 +78,9 @@ async function bootstrap() {
     new FastifyAdapter({
       logger: { level: config.get<string>('LOG_LEVEL') },
     }),
+    {
+      logger: [config.get<LogLevel>('LOG_LEVEL')],
+    },
```
mhuseinov, line 83: hmm., 2 loggers?

alcs-api/src/providers/typeorm/migrations/1657753458415-healthcheck.ts
```diff
@@ -1,15 +0,0 @@
-import { MigrationInterface, QueryRunner } from 'typeorm';
```
mhuseinov, line 1: I assume since we have enough examples of migrations this one will be included to the initial migration, correct?

alcs-api/src/main.ts
dhaselhan: Yes, the first option only seems to set the Fastify logging level, but does not set the NestJS logger level, Not entirely sure why it's separate?

alcs-api/src/providers/typeorm/migrations/1657753458415-healthcheck.ts
dhaselhan: Yes, we will have to remember to add this back into the initial migration when it gets setup

alcs-api/src/main.ts
dhaselhan: Even worse, you see in my MR that even setting the logging level here doesn't seem to impact keycloak even when it has useNestJsLogger:true, so I had to manually set its log level also?

mhuseinov (commented):
💯

mhuseinov (approved):

PR #25, opened 2022-07-25
Add UpdatedBy Subscriber

dhaselhan:
* will set updatedBy on all entities that have the column


alcs-api/src/common/entities/updated-by.subscriber.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(
+    dataSource: DataSource,
+    private readonly cls: ClsService,
+    private userService: UserService,
+  ) {
+    dataSource.subscribers.push(this);
+  }
+
+  async beforeInsert(event: UpdateEvent<any>) {
+    if ('auditCreatedBy' in event.entity) {
```
mhuseinov, line 23: I wish we could do something like 'Base.auditCreatedBy.name'

alcs-api/src/common/entities/updated-by.subscriber.ts
```diff
@@ -0,0 +1,57 @@
...
+  async beforeUpdate(event: UpdateEvent<any>) {
+    if ('auditUpdatedBy' in event.entity) {
+      const userEmail = this.cls.get('userEmail');
+      if (userEmail) {
+        const user = await this.userService.getUser(userEmail);
+        if (!user) {
+          throw new Error(
+            'User not found from token! Has their email changed?',
+          );
+        }
```
mhuseinov, line 49: looks like this check/retrieval can be moved into private function

alcs-api/src/common/entities/updated-by.subscriber.ts
dhaselhan: Since we can't assume that every Entity extends base, this is the 'correct' way.

alcs-api/src/common/entities/updated-by.subscriber.ts
dhaselhan: Totally, good call, it is repeated.

mhuseinov: also I would add email itself into error message

alcs-api/src/common/entities/updated-by.subscriber.ts
mhuseinov: I'm more from perspective of 'auditCreatedBy' not to be a constant, but that is just typescript limitation

mhuseinov (approved):
🔥

PR #31, opened 2022-07-28
Add Paused Feature

dhaselhan:
* Add new ApplicationTimeTrackingService to track active and paused days
* Add to controller and enrich returned Applications with paused and active day counts

### DIFF
--- a/alcs-api/src/application/application-paused.service.spec.ts
+++ b/alcs-api/src/application/application-paused.service.spec.ts
@@ -1,0 +1,100 @@
+import { createMock, DeepMocked } from '@golevelup/nestjs-testing';
+import { Test, TestingModule } from '@nestjs/testing';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+import { BusinessDayModule } from '../providers/business-days/business-day.module';
+import { BusinessDayService } from '../providers/business-days/business-day.service';
+import { ApplicationPaused } from './application-paused.entity';
+import { ApplicationTimeTrackingService } from './application-time-tracking.service';
+import { Application } from './application.entity';
+
+describe('ApplicationTimeTrackingService', () => {
+  let applicationPausedService: ApplicationTimeTrackingService;
+  let mockBusinessDayService: DeepMocked<BusinessDayService>;
+  let applicationRepositoryMock: DeepMocked<Repository<ApplicationPaused>>;
+
+  beforeEach(async () => {
+    applicationRepositoryMock = createMock<Repository<ApplicationPaused>>();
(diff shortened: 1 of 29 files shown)


alcs-api/src/application/application-time-tracking.service.ts
```diff
@@ -0,0 +1,80 @@
...
+      });
+    });
+
+    return resultMap;
+  }
+
+  private async getApplicationPausedTimes(applicationUuids: string[]) {
+    const pausedTimes = (await this.applicationPausedRepository.query(
+      `
+        SELECT
```
mhuseinov, line 46: Is it here as a temporary solution? If yes ignore, otherwise
what if you create a function on db level and then call it here? This way we can keep this things separate and later test them.

alcs-api/src/application/application-paused.service.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { createMock, DeepMocked } from '@golevelup/nestjs-testing';
```
mhuseinov, line 1: maybe rename file to application-time-tracking.service.spec.ts?

dhaselhan: Good catch, definitely should be fixed!

mhuseinov (approved):

alcs-api/src/application/application-time-tracking.service.ts
dhaselhan: Done! Moved to a function and added a migration for it

alcs-api/src/application/application-paused.service.spec.ts
dhaselhan: Done!

alcs-api/src/providers/typeorm/migrations/1659032043316-paused_calculation_function.ts
```diff
@@ -0,0 +1,41 @@
...
+              start_date,
+              application_uuid) AS paused
+      GROUP BY
+          application_uuid $$
+      LANGUAGE SQL;
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`delete from public.application_status`);
```
mhuseinov, line 39: DROP FUNCTION IF EXISTS calculate_paused_time;  or 
DROP FUNCTION IF EXISTS calculate_paused_time(p_ids uuid[]);

I recommend second option since it will take care of overloads

dhaselhan: Thanks for the fix! Added

PR #32, opened 2022-07-28
Add Toast Service + Bug Fixes

dhaselhan:
* Set refresh token into refresh token instead of access token
* Remove update false flag from auditUpdatedAt so we can update the value
* Set AuditUpdateAt manually for Update query

### DIFF
--- a/alcs-frontend/src/app/services/toast/toast.service.ts
+++ b/alcs-frontend/src/app/services/toast/toast.service.ts
@@ -25,7 +25,6 @@ export class ToastService {
 
   showErrorToast(text: string, action: string = '') {
     this.snackBar.open(text, action, {
-      duration: this.TOAST_DURATION,
       panelClass: 'error',
     });
   }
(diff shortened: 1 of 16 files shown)


alcs-frontend/src/app/services/toast/toast.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  showWarningToast(text: string, action: string = '') {
+    this.snackBar.open(text, action, {
+      duration: this.TOAST_DURATION,
+      panelClass: 'warning',
+    });
+  }
+
+  showErrorToast(text: string, action: string = '') {
```
mhuseinov, line 26: I assume this will be changed to manual close in next tickets?

mhuseinov (approved):

alcs-frontend/src/app/services/toast/toast.service.ts
dhaselhan: ah good point, Ill do it now since its one line so we dont forget

dhaselhan: Done!

PR #34, opened 2022-07-29
Feature/nestjs automapper

mhuseinov:
- updates for User mapping
- application mapping
- application status mapping


dhaselhan (approved):
LGMT 👍

PR #35, opened 2022-07-29
Add Application Type, make application type and applicant editable

dhaselhan:


alcs-api/src/application/application.controller.ts
```diff
@@ -53,22 +57,30 @@ export class ApplicationController {
...
   }
 
   @Patch()
   @UserRoles(...ANY_AUTH_ROLE)
   async update(
-    @Body() application: ApplicationPartialDto,
+    @Body() application: ApplicationUpdateDto,
   ): Promise<ApplicationDto> {
     const existingApplication = await this.applicationService.get(
       application.fileNumber,
```
mhuseinov, line 86: I wonder if it worth adding the application number into ServiceValidationException:

`throw new ServiceValidationException(`File not found ${application.fileNumber}`);`

alcs-api/src/common/automapper/application.automapper.profile.ts
```diff
@@ -53,10 +68,13 @@ export class ApplicationProfile extends AutomapperProfile {
...
-              ad.status,
-            );
-            return status;
+            return await this.applicationStatusService.fetchStatus(ad.status);
+          }),
+        ),
+        forMember(
+          async (a) => a.type,
+          mapFrom(async (ad) => {
+            return await this.applicationTypeService.get(ad.status);
```
mhuseinov, line 77: should it be ad.type?

alcs-api/src/application/application.controller.ts
```diff
@@ -53,22 +57,30 @@ export class ApplicationController {
...
-    const app = await this.applicationService.createOrUpdate(entity);
+  async create(
+    @Body() application: CreateApplicationDto,
+  ): Promise<ApplicationDto> {
+    const type = await this.applicationTypeService.get(application.type);
+    const app = await this.applicationService.createOrUpdate({
+      ...application,
+      type,
+      typeUuid: type.uuid,
+      title: 'Do we still need title?',
```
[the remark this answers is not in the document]
dhaselhan: This is a product question, I realized we don't use title in the UI currently but wasn't 100% sure if it could be removed.

alcs-api/src/application/application.controller.ts
dhaselhan: Good suggestion, added!

alcs-api/src/common/automapper/application.automapper.profile.ts
dhaselhan: Yes it should be!

dhaselhan: Fixed

mhuseinov (approved):
🔥

PR #37, opened 2022-08-02
Update App Type Short Codes

dhaselhan:
* Update App Type Short codes to match provided requirements

### DIFF
--- a/alcs-api/src/providers/typeorm/migrations/1659047420726-seed_application_types.ts
+++ b/alcs-api/src/providers/typeorm/migrations/1659047420726-seed_application_types.ts
@@ -4,15 +4,15 @@ export class seedApplicationTypes1659047420726 implements MigrationInterface {
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(`INSERT INTO "public"."application_type" 
       ("uuid", "audit_deleted_date_at", "audit_created_at", "audit_updated_at", "audit_created_by", "audit_updated_by", "code", "description", "label", "short_label", "background_color", "text_color") VALUES
-      ('1a8d99d2-1601-4af2-afb1-f5429b32be5a', NULL, now(), NULL, 'alcs-api', NULL, 'POFO', 'Placement of Fill Only', 'Placement of Fill', 'Fill', '#b2ff59', '#000'),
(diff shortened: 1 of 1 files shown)


alcs-api/src/providers/typeorm/migrations/1659047420726-seed_application_types.ts
```diff
@@ -4,15 +4,15 @@ export class seedApplicationTypes1659047420726 implements MigrationInterface {
...
+      ('9f444184-09be-4a6e-8095-c702cc4681c7', NULL, now(), NULL, 'alcs-api', NULL, 'SUBD', 'Subdivision Permit', 'Subdivison', 'SD', '#c12bdb', '#fff'),
+      ('dc07b1e7-2f58-43ab-817f-7ef83d9d35b0', NULL, now(), NULL, 'alcs-api', NULL, 'EXCL', 'Exclusion Permit', 'Exclusion', 'EXC', '#45f4f4', '#000'),
+      ('e272645d-dcd2-405b-8caf-f94a17500f14', NULL, now(), NULL, 'alcs-api', NULL, 'TURP', 'Transportation, Utility, Trail Permits', 'Transportation, Utility, or Recreational Trail', 'TUR', '#0a1e8d', '#fff'),
       ('ee666f9f-4d94-44dd-b081-535c7aeb50c8', NULL, now(), NULL, 'alcs-api', NULL, 'NFUP', 'Non-Farm Use Permit', 'Non-Farm Use', 'NFU', '#228820', '#fff'),
-      ('f985b157-cc10-4475-8504-06279a2cd73d', NULL, now(), NULL, 'alcs-api', NULL, 'NARS', 'Non-Adhering Residential Use', 'Non-Adhering Residential Use', 'NA Res Use', '#454545', '#fff');
+      ('f985b157-cc10-4475-8504-06279a2cd73d', NULL, now(), NULL, 'alcs-api', NULL, 'NARU', 'Non-Adhering Residential Use', 'Non-Adhering Residential Use', 'NARU', '#454545', '#fff');
```
mhuseinov, line 15: I'm wondering if we want to use 'ON CONFLICT'? so this will be safe to run multiple times?

dhaselhan: I'm not sure to be honest. Do you want me to change it?

mhuseinov: I think if we end up updating this further then it is not a bad idea. Let's hold of till we get deployed db and starting from that point we will right migrations in safe mode

mhuseinov (approved):
🎯

PR #39, opened 2022-08-03
Add CHES Email Service

dhaselhan:
* Add email service using CHES

### DIFF
--- a/alcs-api/src/providers/email/email.service.ts
+++ b/alcs-api/src/providers/email/email.service.ts
@@ -1,0 +1,94 @@
+import { HttpService } from '@nestjs/axios';
+import { Inject, Injectable, Logger } from '@nestjs/common';
+import { firstValueFrom } from 'rxjs';
+import { CONFIG_TOKEN, IConfig } from '../../common/config/config.module';
+
+@Injectable()
+export class EmailService {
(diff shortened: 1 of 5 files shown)


alcs-api/src/providers/email/email.service.ts
```diff
@@ -0,0 +1,94 @@
...
+          {
+            headers: {
+              Authorization: `Bearer ${token}`,
+            },
+          },
+        ),
+      );
+      this.logger.debug({ to, from, subject }, `Email sent`);
+    } catch (e) {
+      this.logger.error(e, 'Failed to Send Email');
```
mhuseinov, line 91: I wonder if we need to record parameters of the sending email into the log or somewhere else. If we have some kind of UI from CHES that is similar to sendgrid then we do nod need it, otherwise I would probably think about having some place to validate on how emails are performing: sent, failed etc.

mhuseinov (approved):
🎯

alcs-api/src/providers/email/email.service.ts
dhaselhan: CHES does not provide any kinda of UI, they do have an endpoint to check on an email after it is sent but its one email. This error should contain the entire request/response to the service.

PR #43, opened 2022-08-05
Refactor Codes into own service

dhaselhan:
* Reduce duplicated code services into dedicated code service
* Make a single http call to get all codes


alcs-api/src/application/application-code/application-code.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    @InjectRepository(ApplicationDecisionMaker)
+    private decisionMakerRepository: Repository<ApplicationDecisionMaker>,
+  ) {}
+
+  async getAllCodes() {
+    const values = await Promise.all([
+      this.typeRepository.find(),
+      this.statusRepository.find(),
+      this.decisionMakerRepository.find(),
+    ]);
```
pdittaro, line 24: So two part mapping to get codes? One to stick it in the service and one to explicitly expose it on the controller?

dhaselhan: Yes. Services should be the ones interacting with the DB and controllers handling mapping to the front end.

dhaselhan: I could probably cache the ones that are static, e.g. everything but status, and load them once when the service starts.

mhuseinov (approved):
👍

PR #47, opened 2022-08-09
Prevent card from going to decision release when its paused

dhaselhan:
* Simple front end rule for now until we get more rule requirements
* Fix error toast to have dismiss button by default, change button font color to grey

### DIFF
--- a/alcs-frontend/src/app/shared/drag-drop-board/drag-drop-board.component.ts
+++ b/alcs-frontend/src/app/shared/drag-drop-board/drag-drop-board.component.ts
@@ -35,7 +35,6 @@ export class DragDropBoardComponent {
       return;
     }
 
-    debugger;
     if (targetColumn.status === 'RELE' && selectedCard.paused) {
       this.toastService.showErrorToast(
         "Paused cards cannot be moved to the 'Decision Released' column. Please unpause the card first and try again"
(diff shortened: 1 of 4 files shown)


alcs-frontend/src/app/shared/drag-drop-board/drag-drop-board.component.ts
```diff
@@ -32,6 +35,14 @@ export class DragDropBoardComponent {
       return;
     }
 
+    debugger;
```
mhuseinov, line 38: we probably do no need this

alcs-frontend/src/app/shared/drag-drop-board/drag-drop-board.component.ts
```diff
@@ -32,6 +35,14 @@ export class DragDropBoardComponent {
...
+    if (targetColumn.status === 'RELE' && selectedCard.paused) {
+      this.toastService.showErrorToast(
+        "Paused cards cannot be moved to the 'Decision Released' column. Please unpause the card first and try again"
```
mhuseinov, line 41: is there any enum that contains description "Decision Released" ?

dhaselhan: I could add a subscriber to the applicationStatus from the service, look up the value from there using the code, and use that label?

alcs-frontend/src/app/shared/drag-drop-board/drag-drop-board.component.ts
dhaselhan: OOF, removed.

mhuseinov (approved):
🎯

PR #49, opened 2022-08-09
Add support for BCeID

dhaselhan:
* Add new page for when user logins without any roles in the system
* Remove auth requirement for logout
* Add new columns for BCeID types and re-type tokens
* Move card components into Board folder


alcs-api/src/common/authorization/authorization.service.ts
```diff
@@ -110,31 +131,40 @@ export class AuthorizationService {
   }
 
   // FIXME: this will be replaced with automapper
```
mhuseinov, line 133: is it a good time to address this?

alcs-api/src/logout/logout.controller.ts
```diff
@@ -1,19 +1,14 @@
...
 
 @ApiOAuth2(config.get<string[]>('KEYCLOAK.SCOPES'))
 @Controller('logout')
-@UseGuards(RoleGuard)
 export class LogoutController {
   constructor(@Inject(CONFIG_TOKEN) private config: IConfig) {}
 
   @Get()
-  @UserRoles(...ANY_AUTH_ROLE)
   async logout() {
```
mhuseinov, line 12: jsut curious what happens if not authenticated user tries to log out?

alcs-api/src/user/user.dto.ts
```diff
@@ -24,11 +20,15 @@ export class CreateOrUpdateUserDto {
 
   @AutoMap()
   @IsString()
-  givenName: string;
+  name?: string;
+
+  @AutoMap()
+  @IsString()
+  givenName?: string;
```
mhuseinov, line 27: why these one optional? this maybe an issue for mentions

mhuseinov: I see, bceid does not have those.

mhuseinov (approved):

alcs-api/src/user/user.dto.ts
dhaselhan: yes, BCeID only has a few fields, we might need to think about how to do initials and other things differently

alcs-api/src/common/authorization/authorization.service.ts
dhaselhan: I think that since the field names don't match, as in we have to change from snake_case to camelCase, I'll remove this comment.

alcs-api/src/logout/logout.controller.ts
dhaselhan: Just tested and it behaves just like as if you were logged in and goes to the redirect URI, so basically takes you to the login page for keycloak.

alcs-api/src/common/authorization/authorization.service.ts
dhaselhan: TODO removed.

PR #50, opened 2022-08-09
Add High Priority Feature

dhaselhan:
* Add toggleable priority to cards
* Add confirmation dialog service for easily producing confirmation dialogs
* Fix toast error styling for smaller messages
*

### DIFF
--- a/alcs-api/src/application/application.service.ts
+++ b/alcs-api/src/application/application.service.ts
@@ -46,31 +46,20 @@ export class ApplicationService {
   async createOrUpdate(
     application: Partial<Application>,
   ): Promise<Application> {
-    let applicationEntity = await this.applicationRepository.findOne({
+    const existingApplication = await this.applicationRepository.findOne({
       where: { fileNumber: application.fileNumber },
     });
 
-    // TODO: replace with AutoMapper
-    applicationEntity = applicationEntity ?? new Application();
-    applicationEntity.fileNumber = application.fileNumber;
-    applicationEntity.applicant = application.applicant;
-    applicationEntity.status = application.status;
-    applicationEntity.statusUuid = application.statusUuid;
-    applicationEntity.type = application.type;
-    applicationEntity.typeUuid = application.typeUuid;
-    applicationEntity.assigneeUuid = application.assigneeUuid;
-    applicationEntity.paused = application.paused;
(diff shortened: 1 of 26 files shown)


alcs-api/src/application/application.service.ts
```diff
@@ -46,31 +46,20 @@ export class ApplicationService {
   async createOrUpdate(
     application: Partial<Application>,
   ): Promise<Application> {
-    let applicationEntity = await this.applicationRepository.findOne({
+    const existingApplication = await this.applicationRepository.findOne({
       where: { fileNumber: application.fileNumber },
     });
 
-    // TODO: replace with AutoMapper
```
mhuseinov, line 53: 👍  nice cleanup

alcs-frontend/src/app/features/card-detail-dialog/card-detail-dialog.component.html
```diff
@@ -1,9 +1,23 @@
...
+    </span>
+  </div>
+  <div class="center">
+    <div class="card-state">{{ currentCard.paused ? 'Paused' : 'Active' }}</div>
+    <div>
+      <button (click)="onTogglePriority()" mat-icon-button>
+        <span
+          [ngClass]="{
+            'high-priority': currentCard.highPriority,
+            'low-priority': !currentCard.highPriority
```
mhuseinov, line 15: should this be just regular priority?

mhuseinov: Is there a chance of getting more than 2 priorities? If yes maybe a priority code will work better here

mhuseinov (approved):

alcs-frontend/src/app/features/card-detail-dialog/card-detail-dialog.component.html
dhaselhan: Renamed the css class to empty-priority and filled-priority to better match what they mean

alcs-api/src/application/application.service.ts
dhaselhan: I started to add AutoMapper but realized it was from Application Entity to Application Entity to Object.assign was all that was needed.

PR #53, opened 2022-08-15
Move off synchronize

dhaselhan:
* Set synhronize false
* Generate initial migration
* Update CLI config to use snake case

### DIFF
--- a/alcs-api/src/providers/typeorm/migrations/1658160165558-migrations.ts
+++ b/alcs-api/src/providers/typeorm/migrations/1658160165558-migrations.ts
@@ -1,0 +1,115 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class migrations1658160165558 implements MigrationInterface {
+  name = 'migrations1658160165558';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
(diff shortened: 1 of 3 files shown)


alcs-api/src/providers/typeorm/migrations/1658160165558-migrations.ts
```diff
@@ -0,0 +1,115 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
pdittaro, line 1: Does this timestamp need to be modified in order to run first in the order?

alcs-api/src/providers/typeorm/orm.config.ts
```diff
@@ -15,7 +15,7 @@ export const getTypeOrmModuleOptions = (
   entities: [join(__dirname, '**', '*.entity.{ts,js}')],
   /* Note : it is unsafe to use synchronize: true for schema synchronization
```
pdittaro, line 16: Can probably remove this comment as well.

pdittaro (approved):

alcs-api/src/providers/typeorm/orm.config.ts
dhaselhan: Removed

PR #56, opened 2022-08-16
Initial Docker-compose

pdittaro:
Created a docker-compose and associated Dockerfiles to create a containerized local development environment that will mirror the deployed environment.

Switched the 'default.json' values in the API config folders to work with docker (since local developers will be overriding with `development.json` values for the most part).


mhuseinov (approved):
💯

PR #60, opened 2022-08-16
Add multi board ability

dhaselhan:


alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -31,7 +34,26 @@ export class BoardComponent implements OnInit {
...
+    });
+
+    this.applicationService.$applicationDecisionMakers.subscribe((dms) => {
+      this.decisionMakers = dms;
+      const decisionMaker = dms.find((dm) => dm.code === this.decisionMakerCode);
+      if (decisionMaker) {
+        this.boardTitle = decisionMaker.label;
+      } else {
+        this.boardTitle = 'All Applications';
+      }
```
mhuseinov, line 54: in this case, I would prefer ternary condition or just set the `this.boardTitle='All Applications'` by default in declaration

mhuseinov (approved):

alcs-frontend/src/app/features/board/board.component.ts
dhaselhan: Good point! I moved it to initialize as All Applications instead.

PR #64, opened 2022-08-19
Favourite Boards

mhuseinov:
* favourite board functionality
* user settings
* unit tests

### DIFF
--- a/alcs-api/src/user/user.dto.ts
+++ b/alcs-api/src/user/user.dto.ts
@@ -1,7 +1,11 @@
 import { AutoMap } from '@automapper/classes';
-import { IsEmail, IsString } from 'class-validator';
+import { IsEmail, IsOptional, IsString } from 'class-validator';
 
 export class CreateOrUpdateUserDto {
+  @AutoMap()
+  @IsOptional()
+  uuid?: string;
+
   @AutoMap()
   @IsEmail()
   email: string;
--- a/alcs-api/src/user/user.dto.ts
+++ b/alcs-api/src/user/user.dto.ts
@@ -66,10 +66,6 @@ export class CreateOrUpdateUserDto {
 export class UserDto extends CreateOrUpdateUserDto {
   @AutoMap()
   @IsString()
-  uuid: string;
-
-  @AutoMap()
-  @IsString()
   initials?: string;
 
   @AutoMap()
--- a/alcs-api/src/user/user.dto.ts
+++ b/alcs-api/src/user/user.dto.ts
@@ -1,5 +1,6 @@
 import { AutoMap } from '@automapper/classes';
 import { IsEmail, IsOptional, IsString } from 'class-validator';
+import { UserSettings } from './user.entity';
 
 export class CreateOrUpdateUserDto {
   @AutoMap()
@@ -49,6 +50,9 @@ export class CreateOrUpdateUserDto {
   @AutoMap()
   @IsString()
   bceidUserName?: string;
+
+  @AutoMap()
+  settings?: UserSettings;
 }
 
 export class UserDto extends CreateOrUpdateUserDto {
--- a/alcs-api/src/user/user.dto.ts
+++ b/alcs-api/src/user/user.dto.ts
@@ -1,6 +1,10 @@
 import { AutoMap } from '@automapper/classes';
-import { IsEmail, IsOptional, IsString } from 'class-validator';
-import { UserSettings } from './user.entity';
+import { Type } from 'class-transformer';
+import { IsDefined, IsEmail, IsOptional, IsString } from 'class-validator';
+
+export class UserSettingsDto {
+  favoriteBoards: string[];
+}
 
 export class CreateOrUpdateUserDto {
   @AutoMap()
@@ -45,14 +49,18 @@ export class CreateOrUpdateUserDto {
 
   @AutoMap()
   @IsString()
+  @IsOptional()
   bceidGuid?: string;
 
   @AutoMap()
   @IsString()
+  @IsOptional()
   bceidUserName?: string;
 
   @AutoMap()
-  settings?: UserSettings;
+  @Type(() => UserSettingsDto)
+  @IsDefined()
+  settings?: UserSettingsDto;
 }
 
 export class UserDto extends CreateOrUpdateUserDto {
--- a/alcs-api/src/user/user.service.ts
+++ b/alcs-api/src/user/user.service.ts
@@ -3,6 +3,7 @@ import { InjectMapper } from '@automapper/nestjs';
 import { Injectable, Logger } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
+import { ServiceNotFoundException } from '../common/exceptions/base.exception';
 import { CreateOrUpdateUserDto, UserDto } from './user.dto';
 import { User } from './user.entity';
 
@@ -35,7 +36,9 @@ export class UserService {
     const existingUser = await this.getUser(email);
 
     if (!existingUser) {
-      throw new Error('User not found');
+      throw new ServiceNotFoundException(
+        `User with provided email not found ${email}`,
+      );
     }
 
     return this.userRepository.softRemove(existingUser);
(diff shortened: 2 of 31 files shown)


alcs-api/src/user/user.dto.ts
```diff
@@ -1,7 +1,16 @@
...
-import { IsEmail, IsString } from 'class-validator';
+import { Type } from 'class-transformer';
+import { IsDefined, IsEmail, IsOptional, IsString } from 'class-validator';
+
+export class UserSettingsDto {
+  favoriteBoards: string[];
+}
 
 export class CreateOrUpdateUserDto {
+  @AutoMap()
```
dhaselhan, line 10: Since a user can't change uuid, and the lookup is by email, then this doesn't need to be here.

alcs-frontend/src/app/shared/header/header.component.scss
```diff
@@ -22,17 +24,43 @@
...
+  color: #d3d3d3;
+  
+  mat-icon {
+    margin-bottom: 5px;
+  }
+}
+
+.favorite-btn.gold {
+  color: #ffd700;
+  visibility: visible !important;
```
dhaselhan, line 54: If you change the class to:
.board-item .favorite-btn.gold you shouldn't need important

alcs-frontend/src/app/shared/header/header.component.ts
```diff
@@ -12,27 +16,109 @@ import { AuthenticationService } from '../../services/authentication/authenticat
...
+  onSelectBoard(dmCode: string) {
+    this.router.navigateByUrl(`/board/${dmCode}`);
   }
 
   onLogout() {
     this.authService.logout();
   }
+
+  private updateFavoriteBoardsList(dm: ApplicationDecisionMakerDto) {
+    if (!this.currentUserProfile) {
```
dhaselhan, line 88: The parent function has already run this check and its only called from that one function, can this check be removed?

alcs-frontend/src/app/shared/header/header.component.ts
```diff
@@ -12,27 +16,109 @@ import { AuthenticationService } from '../../services/authentication/authenticat
...
+      return;
+    }
+
+    if (!this.currentUserProfile?.settings?.favoriteBoards) {
```
dhaselhan, line 92: Should we add a default of empty array in the migration? That way we wouldn't need to worry about checking if it exists or not.

alcs-frontend/src/app/shared/header/header.component.ts
```diff
@@ -12,27 +16,109 @@ import { AuthenticationService } from '../../services/authentication/authenticat
...
+    if (!this.currentUserProfile) {
+      return;
+    }
+
+    this.updateFavoriteBoardsList(dm);
+
+    try {
+      await this.userService.updateUser(this.currentUserProfile);
+      await this.userService.fetchUsers();
+      this.applicationService.$applicationDecisionMakers.next(this.decisionMakers);
```
dhaselhan, line 119: Only the application service should be updating its decision makers. By doing this the decision makers in the application service will not match the decision makers in the observable.

alcs-frontend/src/app/shared/header/header.component.ts
```diff
@@ -12,27 +16,109 @@ import { AuthenticationService } from '../../services/authentication/authenticat
...
+    this.userService.$users.subscribe((users) => {
+      this.currentUserProfile = users.find((u) => u.email === this.currentUser.email);
     });
+
+    this.$sortedDecisionMakers = this.applicationService.$applicationDecisionMakers.pipe(
+      map((dms) => this.transformDecisionMakers(dms))
+    );
+  }
+
+  private transformDecisionMakers(dms: ApplicationDecisionMakerDto[]) {
```
dhaselhan, line 51: Since favourites also needs to be indicated on the board component, should this enrichment be done in the application service?

The application service could have the subscribe and this logic, this would mean that isFavourite is accessible in all components instead of just the header.

alcs-frontend/src/app/shared/header/header.component.ts
```diff
@@ -12,27 +16,109 @@ import { AuthenticationService } from '../../services/authentication/authenticat
...
+    this.authService.isAuthenticated.subscribe((isAuthenticated) => {
+      if (isAuthenticated) {
+        this.currentUser = this.authService.getCurrentUser();
+
+        this.userService.fetchUsers();
+        this.applicationService.setup();
+      }
+    });
+
+    this.userService.$users.subscribe((users) => {
```
dhaselhan, line 42: Should we add a $userProfile to the userService that publishes the current users profile?

alcs-api/src/user/user.service.ts
```diff
@@ -48,4 +51,24 @@ export class UserService {
       },
     });
   }
+
+  async getUserByUuid(uuid: string) {
```
dhaselhan, line 55: can remove the word User and just have getByUuid since this is the UserService

mhuseinov: I'm being consistent with the rest of the service. Do you want the rest to be renamed too?

alcs-frontend/src/app/shared/header/header.component.ts
mhuseinov: this is a way to avoid TS warnings about possible null. Another way would be to force it everywhere with '!' but it looks more correct to me to have this check.

alcs-frontend/src/app/shared/header/header.component.ts
mhuseinov: Sure, I'll add it on Board page, when I started this ticket I did not have titles on board at all 😅

alcs-frontend/src/app/shared/header/header.component.ts
dhaselhan: That makes sense

alcs-api/src/user/user.service.ts
dhaselhan: Ah oops, I just noticed update was called update. Should update be updateUser to be consistent then?

mhuseinov: I think we need to agree on the naming in general, and I can address all the API level code to be consistent across the app

mhuseinov: I like the idea that our service|entity|controller names include domain names in their naming and that way, we can avoid including the domain names into service method names to improve readability. If you agree, I can address the rest of API code to match it.

alcs-frontend/src/app/features/board/board.component.scss
```diff
@@ -1,5 +1,12 @@
 @import '../../../styles/colors';
 
-.gold {
-    color: $favorite-star
-  }
\ No newline at end of file
+.board-title:hover app-favorite-button {
+    ::ng-deep .favorite-btn {
+    visibility: visible ;
```
dhaselhan, line 5: Looks like it needs a format, odd spaces at the end before the semicolon

alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -20,10 +22,11 @@ export class BoardComponent implements OnInit {
   cards: CardData[] = [];
   columns: DragDropColumn[] = [];
   boardTitle = 'All Applications';
-  boardIsFavorite: boolean | undefined;
+  boardIsFavorite?: boolean = false;
```
dhaselhan, line 25: If you are initializing to false you can remove the ?

alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -20,10 +22,11 @@ export class BoardComponent implements OnInit {
...
+  currentUserProfile?: UserDto;
```
dhaselhan, line 26: I don't see this used in the HTML, why was it added?

alcs-frontend/src/app/shared/favorite-button/favorite-button.component.html
```diff
@@ -0,0 +1,3 @@
+<button mat-icon-button [ngClass]="{ gold: isFavorite, 'favorite-btn': true }" (click)="onFavoriteClicked($event)">
+  <mat-icon> star</mat-icon>
```
dhaselhan, line 2: intentional to have a space before star?

alcs-frontend/src/app/shared/favorite-button/favorite-button.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  templateUrl: './favorite-button.component.html',
+  styleUrls: ['./favorite-button.component.scss'],
+})
+export class FavoriteButtonComponent implements OnInit {
+  @Input()
+  currentUserProfile?: UserDto;
+  @Input()
+  dmCode: string | undefined;
+  @Input()
+  isFavorite?: boolean = false;
```
dhaselhan, line 17: remove ? since its initialized

alcs-api/src/user/user.service.ts
dhaselhan: As discussed this will be done in another MR with its own ticket to track.

dhaselhan (approved):
LGTM 👍

PR #65, opened 2022-08-19
Move business day calculations to SQL

dhaselhan:
* Add more PGTap tests


alcs-api/src/providers/typeorm/migrations/1660859890136-active_days_calculations.ts
```diff
@@ -0,0 +1,123 @@
...
+              holiday_entity
+          WHERE
+              holiday_entity. "day" BETWEEN start_date
+              AND end_date;
+          RETURN holiday_count;
+      END;
+      $$;
+    `);
+
+    await queryRunner.query(`DROP function calculate_paused_time;`);
```
mhuseinov, line 48: why do you need to drop function here?

mhuseinov (approved):
🎯

alcs-api/src/providers/typeorm/migrations/1660859890136-active_days_calculations.ts
dhaselhan: Because the return type changed, create or replace will error out and you must DROP it first

PR #67, opened 2022-08-23
Add Notifications

dhaselhan:
* Add job/queue/consumer to delete old notifications
* Create notifications for mentions and application assignment


alcs-api/src/application/application.controller.ts
```diff
@@ -153,7 +157,18 @@ export class ApplicationController {
       highPriority: application.highPriority,
     });
 
-    const mappedApps = await this.applicationService.mapToDtos([app]);
+    if (app.assigneeUuid !== existingApplication.assigneeUuid) {
```
mhuseinov, line 160: What if users mentioned themselves?

alcs-api/src/application/application.controller.ts
```diff
@@ -153,7 +157,18 @@ export class ApplicationController {
...
+      this.notificationService.createForApplication(
```
mhuseinov, line 161: Is await omitted on purpose?

alcs-api/src/comment/comment.controller.ts
```diff
@@ -49,12 +49,14 @@ export class CommentController {
     @Body() comment: CreateCommentDto,
     @Req() req,
   ): Promise<CommentDto> {
+    const mappedMentions = await this.mapMentions(comment);
     const newComment = await this.commentService.create(
       comment.fileNumber,
       comment.body,
       req.user.entity,
-      await this.mapMentions(comment),
+      mappedMentions,
```
mhuseinov, line 57: Is there any specific reason or just code-style here?

dhaselhan: I had originally changed this because I added some code here, but when I ended up re-working it and not needing the mappedMentions variable. I can revert this if you'd like.

alcs-api/src/application/application.controller.ts
dhaselhan: Yes, I figure if something fails with notifications we don't really need to alert the user or abort the return of the data below.

alcs-api/src/notification/notification.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+import { NotificationDto } from './notification.dto';
+import { NotificationService } from './notification.service';
+
+@ApiOAuth2(config.get<string[]>('KEYCLOAK.SCOPES'))
+@Controller('notification')
+@UseGuards(RoleGuard)
+export class NotificationController {
+  constructor(private notificationService: NotificationService) {}
+
+  @Get('')
```
mhuseinov, line 24: Maybe without quotes @Get()?

alcs-api/src/notification/notification.service.ts
```diff
@@ -0,0 +1,106 @@
...
+      where: {
+        receiver: {
+          uuid: userUuid,
+        },
+      },
+      order: {
+        createdAt: 'DESC',
+      },
+    });
+    return this.mapToDto(notifications);
```
mhuseinov, line 34: Usually, you do not map service results to DTO on the Service layer. DTO responsible more for presentation layer => controllers

dhaselhan: Done! Moved to controller instead

alcs-api/src/notification/notification.controller.ts
dhaselhan: Ah good catch, removed the quotes

alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -143,7 +143,7 @@ export class BoardComponent implements OnInit {
       status: application.status,
       title: `${application.fileNumber} (${application.applicant})`,
       assigneeInitials: application.assignee
-        ? `${application.assignee?.givenName.charAt(0)}${application.assignee?.familyName.charAt(0)}`
+        ? `${application.assignee?.givenName?.charAt(0)}${application.assignee?.familyName?.charAt(0)}`
```
mhuseinov, line 146: If I'm not mistaken assignee is of type UserDto, and UserDto has the initials set in auto mapper. I think we should be able to use that here

alcs-frontend/src/app/services/notification/notification.service.ts
```diff
@@ -0,0 +1,24 @@
...
+import { NotificationDto } from './notification.dto';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class NotificationService {
+  constructor(private http: HttpClient) {}
+
+  async fetchMyNotifications() {
+    return await firstValueFrom(this.http.get<NotificationDto[]>(`${environment.apiRoot}/notification`));
```
mhuseinov, line 14: I think you should be able to remove async and await from function and use simple return

alcs-api/src/application/application.controller.ts
dhaselhan: Good idea, added logic + unit test to NOT create a notification if user is assigning to themselves

alcs-frontend/src/app/shared/header/notifications/notifications.component.scss
```diff
@@ -0,0 +1,83 @@
+@import '../../../../styles/colors';
+
+.unread-notifications::before {
+  content: '';
+  position: absolute;
+  top: 13px;
+  background: $error-color;
+  width: 7px;
```
mhuseinov, line 8: I know one designer who will be furious to see 7px instead of 8px 😅

alcs-api/src/comment/comment.controller.ts
mhuseinov: Feel free to leave it as is, I was just curious about the reason

alcs-frontend/src/app/features/board/board.component.ts
dhaselhan: You are very correct! I added initials to the front end DTO and it works great.

alcs-frontend/src/app/shared/header/notifications/notifications.component.scss
dhaselhan: The good news is I already got approval via screenshot haha

alcs-frontend/src/app/services/notification/notification.service.ts
dhaselhan: Done

mhuseinov (approved):
💯

PR #73, opened 2022-08-25
Add Board Entity and associated features

dhaselhan:
* Add new board entity to replace decision maker
* Board entity knows which statuses and their order are in that board
* Add new service to get app by board and move apps between boards

### DIFF
--- a/alcs-api/src/application/application.controller.ts
+++ b/alcs-api/src/application/application.controller.ts
@@ -45,7 +45,7 @@ export class ApplicationController {
   @Get()
   @UserRoles(...ANY_AUTH_ROLE)
   async getAll(): Promise<ApplicationDto[]> {
-    const applications = await this.applicationService.getAll({});
+    const applications = await this.applicationService.getAll();
     return this.applicationService.mapToDtos(applications);
   }
 
--- a/alcs-api/src/application/application.controller.ts
+++ b/alcs-api/src/application/application.controller.ts
@@ -20,7 +20,6 @@ import { UserRoles } from '../common/authorization/roles.decorator';
 import { ServiceValidationException } from '../common/exceptions/base.exception';
 import { NotificationService } from '../notification/notification.service';
 import { ApplicationCodeService } from './application-code/application-code.service';
-import { ApplicationDecisionMaker } from './application-code/application-decision-maker/application-decision-maker.entity';
 import { ApplicationRegion } from './application-code/application-region/application-region.entity';
 import { ApplicationType } from './application-code/application-type/application-type.entity';
 import { ApplicationStatus } from './application-status/application-status.entity';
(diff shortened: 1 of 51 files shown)


alcs-api/src/application/application.controller.ts
```diff
@@ -45,14 +44,8 @@ export class ApplicationController {
...
-  async getAll(@Query('dm') dm?: string): Promise<ApplicationDto[]> {
-    let decisionMaker;
-    if (dm) {
-      decisionMaker = await this.codeService.fetchDecisionMaker(dm);
-    }
-    const applications = await this.applicationService.getAll({
-      decisionMaker,
-    });
+  async getAll(): Promise<ApplicationDto[]> {
+    const applications = await this.applicationService.getAll({});
```
mhuseinov, line 48: is it possible to mark filter as optional? So if you need to get all values it will looks like this: ....applicationService.getAll();

alcs-api/src/board/board.service.ts
```diff
@@ -0,0 +1,57 @@
...
+      board: {
+        code,
+      },
+    });
+  }
+
+  async changeBoard(fileNumber: string, code: string) {
+    const application = await this.applicationService.get(fileNumber);
+    if (!application) {
+      throw new ServiceNotFoundException('Failed to find application');
```
mhuseinov, line 39: I prefer to include search criteria into error message since it make debugging on deployed env easier

alcs-api/src/board/board.service.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const board = await this.boardRepository.findOne({
+      where: {
+        code,
+      },
+      relations: ['statuses'],
```
mhuseinov, line 46: I think it is worth moving 'statuses' into defaultRelations as we did in other services

alcs-api/src/application/application.controller.ts
dhaselhan: Absolutely, not sure why I didn't do that in the first place lol.

alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -100,8 +104,12 @@ export class BoardComponent implements OnInit {
...
-      dialogRef.afterClosed().subscribe(() => {
+      dialogRef.afterClosed().subscribe((isDirty) => {
         this.setUrl();
+
+        if (isDirty && this.selectedBoardCode) {
+          this.loadApplications(this.selectedBoardCode);
+        }
       });
     } catch (err) {
       console.log(err);
```
mhuseinov, line 115: probably we need to create a ticket to replace console.log with appropriate toast messages where applicable

alcs-api/src/board/board.service.ts
dhaselhan: Good call, done!

alcs-api/src/board/board.service.ts
dhaselhan: Added the filenumber here and the code to the error below

alcs-api/src/application/application.controller.ts
dhaselhan: Done!

alcs-frontend/src/app/features/board/board.component.ts
dhaselhan: added a toast message for this catch block

mhuseinov (approved):
💯

PR #74, opened 2022-08-26
Add Received Date to Apps

dhaselhan:
* Add Received date to apps and add date picker in create dialog
* Improve logging on board controller

### DIFF
--- a/alcs-api/src/application/application.entity.ts
+++ b/alcs-api/src/application/application.entity.ts
@@ -34,7 +34,6 @@ export class Application extends Base {
   @AutoMap()
   @Column({
     type: 'boolean',
-    nullable: false,
     default: false,
   })
   paused: boolean;
@@ -42,7 +41,6 @@ export class Application extends Base {
   @AutoMap()
   @Column({
     type: 'boolean',
-    nullable: false,
     default: false,
   })
   highPriority: boolean;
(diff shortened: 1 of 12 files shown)


alcs-api/src/application/application.entity.ts
```diff
@@ -54,6 +47,13 @@ export class Application extends Base {
   })
   highPriority: boolean;
 
+  @AutoMap()
+  @Column({
+    type: 'timestamptz',
+    nullable: false,
```
mhuseinov, line 53: if I'm not mistaken nullable is false by default

alcs-api/src/application/application.entity.ts
```diff
@@ -31,13 +31,6 @@ export class Application extends Base {
   @Column()
   applicant: string;
 
-  @AutoMap()
-  @Column({
-    type: 'uuid',
-    nullable: true,
-  })
-  assigneeUuid: string;
```
mhuseinov, line 39: why did you decide to remove assigneUuid? I can see references to these fields, is the feature still working properly?

dhaselhan: I moved it down to be next to assignee instead of being near the top, it is still there

alcs-api/src/application/application.entity.ts
dhaselhan: cleaned them up!

mhuseinov (approved):
🔥

PR #75, opened 2022-08-26
ALCS-75 display decision meeting date on board

mhuseinov:
* adjust board styles according to design
* added decision meeting date on card
Screen Shot 2022-08-26 at 3 21 07 PM

### DIFF
--- a/alcs-frontend/src/app/shared/card/card.component.ts
+++ b/alcs-frontend/src/app/shared/card/card.component.ts
@@ -1,5 +1,6 @@
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { ApplicationTypeDto } from '../../services/application/application-code.dto';
+import { ApplicationDecisionMeetingDto } from '../../services/application/application.dto';
 
 export interface CardData {
   id: string;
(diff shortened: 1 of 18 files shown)


alcs-frontend/src/app/shared/card/card.component.ts
```diff
@@ -28,5 +31,23 @@ export class CardComponent implements OnInit {
     this.cardSelected.emit(cardId);
   }
 
-  ngOnInit(): void {}
+  private getLatestDecisionDate() {
+    return new Date(
+      Math.max(
+        ...this.cardData.decisionMeetings.map((element) => {
+          return new Date(element.date).valueOf();
```
dhaselhan, line 38: This looks like a redundant conversion, looks like its going Number -> Date -> Number -> Date. I think you can remove the new Date() from inside the map.

mhuseinov: nope, without conversion it return Invalid Date

dhaselhan: My bad, element.date is a string

dhaselhan (approved):

PR #76, opened 2022-08-26
Add Intake Page

dhaselhan:
* Create inline edit for date picker
* Add new ApplicationDetailService to share Application across all pages

### DIFF
--- a/alcs-api/src/application/application.controller.ts
+++ b/alcs-api/src/application/application.controller.ts
@@ -89,7 +89,7 @@ export class ApplicationController {
   async update(
     @Body() application: ApplicationUpdateDto,
     @Req() req,
-  ): Promise<ApplicationDto> {
+  ): Promise<ApplicationDetailedDto> {
     const existingApplication = await this.applicationService.get(
       application.fileNumber,
     );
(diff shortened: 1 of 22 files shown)


alcs-api/src/application/application.controller.ts
```diff
@@ -132,27 +132,54 @@ export class ApplicationController {
...
   @Delete()
   @UserRoles(...ANY_AUTH_ROLE)
   async softDelete(@Body() applicationNumber: string): Promise<void> {
     await this.applicationService.delete(applicationNumber);
   }
+
+  private formatIncomingDate(date?: number) {
+    if (date) {
+      return new Date(date);
+    } else if (date === null) {
```
mhuseinov, line 179: why do you need to distinguish between null and undefined?

dhaselhan: I need to set null to clear the date in case of someone removing it

alcs-frontend/src/app/shared/inline-datepicker/inline-datepicker.component.html
```diff
@@ -0,0 +1,27 @@
+<div>
+  <span *ngIf="!isEditing">
+    <span style="color: #666; font-style: italic" *ngIf="!selectedValue"> Not set </span>
+    <span *ngIf="selectedValue">
+      {{ formattedDate }}
+    </span>
+    <button style="height: 24px; width: 24px" mat-icon-button (click)="toggleEdit()">
```
mhuseinov, line 7: is it possible to have styles in scss instead?

mhuseinov (approved):
👍

alcs-frontend/src/app/shared/inline-datepicker/inline-datepicker.component.html
dhaselhan: Done! Cleaned up here and also the intake.component.html

PR #77, opened 2022-08-26
Make navigation dynamic

dhaselhan:
* use same var for routes and for menu
* add highlighting for active and hover menu items


mhuseinov (approved):
🎯

PR #81, opened 2022-08-30
Change to Postgres 12

dhaselhan:
* Fix migration to set proper INT type
* Fix PG Tap tests for new date received column


mhuseinov (approved):
💯

PR #82, opened 2022-08-30
OpenShift: Kustomize templates and helm charts for deploying ALCS

pdittaro:
Components:
- API
- Frontend
- Patroni (postgres)
- Redis standalone


alcs-frontend/src/assets/env.js
```diff
@@ -0,0 +1,7 @@
+(function(window) {
```
mhuseinov, line 1: Do I get it right that we will be duplicating configuration to assets config files for deployment?

openshift/components/frontend/base/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+              port: 8081
+          livenessProbe:
+            timeoutSeconds: 3
+            initialDelaySeconds: 30
+            httpGet:
+              path: "/"
+              port: 8081
+          resources:
+            limits:
+              cpu: 100m
```
mhuseinov, line 35: Are these updated resource values or initial ones?

mhuseinov (approved):
🔥

PR #86, opened 2022-09-06
Add Decision Documents

dhaselhan:
* Create generic Document and Document service entity
* Create AppDocuments that use Documents and link to Applications
* Add local minio to the docker-compose


alcs-api/src/application/application-document/application-document.entity.ts
```diff
@@ -0,0 +1,36 @@
...
+      Object.assign(this, data);
+    }
+  }
+
+  @AutoMap()
+  @PrimaryGeneratedColumn('uuid')
+  uuid: string;
+
+  @Column()
+  type: 'decisionDocument';
```
mhuseinov, line 28: why not default value?

alcs-api/src/main.ts
```diff
@@ -72,6 +73,19 @@ function registerPipes(app: NestFastifyApplication) {
   );
 }
 
+async function registerMultiPart(app: NestFastifyApplication) {
+  await app.register(fastifyMultipart, {
+    limits: {
```
mhuseinov, line 78: any file type limits?

alcs-frontend/src/app/features/application/review/decision-document/decision-document.component.scss
```diff
@@ -0,0 +1,33 @@
...
+  width: 100%;
+}
+
+.mat-column-action {
+  width: 8%;
+  text-align: right;
+}
+
+a {
+  font-size: 14px !important;
```
mhuseinov, line 26: just to clarify, we are not following 8, 16, 24 .... sizes anymore?

dhaselhan: The typography file has fonts all over the place, 8,9,14,16, etc

alcs-frontend/src/app/services/application/application-document/application-document.service.ts
```diff
@@ -0,0 +1,52 @@
+import { HttpClient } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { firstValueFrom } from 'rxjs';
+import { environment } from '../../../../environments/environment';
+import { ToastService } from '../../toast/toast.service';
+import { ApplicationDocumentDto } from './application-document.dto';
+
+const MAX_FILE_SIZE = 104857600;
```
mhuseinov, line 8: I would move this to config. I think this is something that should be easily controlled without commit

alcs-api/src/main.ts
dhaselhan: I wasn't given any, what are you thinking?

alcs-api/src/application/application-document/application-document.entity.ts
dhaselhan: I want to avoid having to do a migration when we add a second type of document.

alcs-frontend/src/app/services/application/application-document/application-document.service.ts
dhaselhan: Will do, I was on the fence about it. I can also move the one on the front-end to the environment file.

alcs-frontend/src/app/services/application/application-document/application-document.service.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  constructor(private http: HttpClient, private toastService: ToastService) {}
+
+  async list(fileNumber: string) {
+    return firstValueFrom(this.http.get<ApplicationDocumentDto[]>(`${this.url}/application/${fileNumber}`));
+  }
+
+  async upload(fileNumber: string, file: File) {
+    if (file.size > MAX_FILE_SIZE) {
+      this.toastService.showWarningToast('Maximum file size is 100MB, please choose a smaller file');
```
mhuseinov, line 24: the size can be calculated based on the MAX_FILE_SIZE variable, so this stays in sync with limits

alcs-api/src/main.ts
mhuseinov: pdf, doc, docx, xls, xlsx, png, jpg, jpeg and something for geo data

alcs-frontend/src/app/features/application/review/decision-document/decision-document.component.scss
mhuseinov: then it will probably be later thing

mhuseinov (approved):
🎯

alcs-api/src/main.ts
dhaselhan: Just checked and there is no setting for Fastify Multipart to limit file types, this is something we will have to do on our side.

alcs-frontend/src/app/services/application/application-document/application-document.service.ts
dhaselhan: Done!

alcs-frontend/src/app/services/application/application-document/application-document.service.ts
dhaselhan: Done!

PR #87, opened 2022-09-07
Generalize Document Uploads

dhaselhan:
* Refactor existing decision document component to general application document
* Use general component and setup new decision page with uploads
* Add decisionDate field to applications and add picker for it
* Add a base application/fileNumber path for each upload


alcs-frontend/src/app/features/application/application.component.ts
```diff
@@ -24,15 +25,20 @@ export const childRoutes = [
     menuTitle: 'Processing',
     component: ProcessingComponent,
   },
+  {
+    path: 'site-visit-applicant',
```
pdittaro, line 29: This should probably be `site-visit-meeting`?

pdittaro (approved):
LGTM

alcs-frontend/src/app/features/application/application.component.ts
dhaselhan: Changed!

PR #89, opened 2022-09-08
Add new card UI

dhaselhan:
* Also add fix for duplicate columns on executive board


alcs-api/src/providers/typeorm/migrations/1662659560119-fix_duplicate_board_status.ts
```diff
@@ -0,0 +1,43 @@
...
+    await queryRunner.query(
+      `DELETE FROM "board_status" WHERE board_uuid = 'd8c18278-cb41-474e-a180-534a101243ab'`,
+    );
+
+    //Re-insert a single set
+    await queryRunner.query(`
+      INSERT INTO "public"."board_status"
+      ("uuid", "audit_deleted_date_at", "audit_created_at", "audit_updated_at", "audit_created_by", "audit_updated_by", "order", "board_uuid", "status_uuid") VALUES
+       --- Executive Committee
+      (DEFAULT, DEFAULT, DEFAULT, DEFAULT, 'alcs-api', DEFAULT, 0, 'd8c18278-cb41-474e-a180-534a101243ab', 'e6ddd1af-1cb9-4e45-962a-92e8d532b149'),
```
pdittaro, line 19: Are all the board UUIDs pregenerated via migrations?

alcs-frontend/src/app/features/board/card-detail-dialog/card-detail-dialog.component.html
```diff
@@ -1,78 +1,115 @@
...
+          [routerLink]="['application', application.fileNumber, 'site-visit-meeting']"
+          mat-menu-item
+        >
+          Add Site Visit / Meeting
+        </button>
+      </mat-menu>
+    </div>
+  </div>
+  <div>
+    <span class="region">{{ application.regionDetails?.label }} Region</span>
```
pdittaro, line 48: If this needs a safe traversal operator, is there a chance it doesn't exist? What should this display if so?

alcs-api/src/providers/typeorm/migrations/1662659560119-fix_duplicate_board_status.ts
dhaselhan: Yes! We pre-generated them all so its the same across all systems to keep things easy like this script.

alcs-frontend/src/app/features/board/card-detail-dialog/card-detail-dialog.component.html
dhaselhan: Region is only nullable temporarily, as fallout of doing the new card ui before doing the new card creation UI. Once we complete the new card creation UI, this will no longer be nullable and will always be displayed.

pdittaro: Let's remember to change it afterwards then

pdittaro (approved):

PR #91, opened 2022-09-09
Add Subtasks and first subtask type

dhaselhan:
* Add subtask and subtask type entites
* Add new service, controller to handle these
* Add new component to frontend for rendering
* Clean up code by adding util time formatter for api to DRY up code


alcs-api/src/application/application-subtask/application-subtask.controller.spec.ts
```diff
@@ -0,0 +1,107 @@
+import { classes } from '@automapper/classes';
+import { AutomapperModule } from '@automapper/nestjs';
+import { createMock, DeepMocked } from '@golevelup/nestjs-testing';
+import { Test, TestingModule } from '@nestjs/testing';
+import { ClsService } from 'nestjs-cls';
+import { ApplicationSubtaskProfile } from '../../common/automapper/application-subtask.automapper.profile';
+import { ApplicationProfile } from '../../common/automapper/application.automapper.profile';
+import { mockKeyCloakProviders } from '../../common/utils/test-helpers/mockTypes';
```
pdittaro, line 8: Little thing, but why isn't this upper camel case?

alcs-api/src/application/application-subtask/application-subtask.entity.ts
```diff
@@ -0,0 +1,33 @@
...
+    super();
+    if (data) {
+      Object.assign(this, data);
+    }
+  }
+
+  @CreateDateColumn({ type: 'timestamptz' })
+  createdAt: Date;
+
+  @Column({ nullable: true })
```
pdittaro, line 19: Do we not specify timestamptz for this attribute as well?

alcs-api/src/application/application-subtask/application-subtask.service.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      applicationSubtaskService.create('mock-file', 'fake-type'),
+    ).rejects.toMatchObject(
+      new ServiceNotFoundException(`Invalid subtask type fake-type`),
+    );
+
+    expect(applicationService.get).toHaveBeenCalled();
+    expect(mockSubtaskTypeRepo.findOne).toHaveBeenCalled();
+  });
+
+  it('should call save and set fields for update', async () => {
```
pdittaro, line 97: Is there any case where the createdAt needs to change (or should potentially be startedAt)?

alcs-api/src/application/application-subtask/application-subtask.service.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    if (!existingTask) {
+      throw new ServiceNotFoundException(`Failed to find task ${uuid}`);
+    }
+
+    if (updates.completedAt === null) {
+      existingTask.completedAt = null;
+    }
+    if (updates.completedAt) {
+      existingTask.completedAt = new Date(updates.completedAt);
```
pdittaro, line 75: `existingTask.completedAt = updates.completedAt ? new Date(updates.completedAt) : null;` ?

alcs-frontend/src/app/shared/utils/api-date-formatter.ts
```diff
@@ -0,0 +1,4 @@
+import * as dayjs from 'dayjs';
+
+export const formatDateForApi = (date: Date | number) =>
+  date ? dayjs(date).startOf('day').add(12, 'hours').valueOf() : date;
```
pdittaro, line 4: Black magic?

alcs-api/src/application/application-subtask/application-subtask.controller.spec.ts
dhaselhan: Its a variable, not a type or class.

alcs-api/src/application/application-subtask/application-subtask.entity.ts
dhaselhan: It should be, good catch

alcs-api/src/application/application-subtask/application-subtask.service.ts
dhaselhan: This would set the variable to null when updates.completedAt was undefined, clearing it from the DB unintentionally.

alcs-frontend/src/app/shared/utils/api-date-formatter.ts
dhaselhan: Yes, since we have to give everything a time (javascript has no concept of date without time) we move it to the middle of the day to keep things a little more standard than using the exact time from the user.

pdittaro: Lets through that in as a comment for posterity

dhaselhan: Good idea, this would be very confusing for anyone to randomly see. Added a comment explaining this is done to "standardize" the dates.

pdittaro (approved):

PR #96, opened 2022-09-14
Only show GIS users for assigning GIS tasks

dhaselhan:


alcs-frontend/src/app/features/home/gis/gis-subtasks.component.ts
```diff
@@ -33,9 +31,8 @@ export class GisSubtasksComponent implements OnInit {
       this.statuses = statuses;
     });
 
-    this.$users = this.userService.$users;
     this.userService.$users.subscribe((users) => {
-      this.users = users;
+      this.gisUsers = users.filter((user) => user.clientRoles.includes('GIS'));
```
pdittaro, line 35: Think this is already the case, but why are we not calling these userRoles or just roles? Client has a different connotation with OAuth2

dhaselhan: They were called clientRoles on the token, and currently we have 1:1 mapped the token fields to our fields.

pdittaro (approved):

PR #98, opened 2022-09-14
GIS Subtask Fixes

dhaselhan:
* Replace null with isNull to only fetch incomplete subtasks
* Change alignment of row to better align dropdowns
* Add min-width to assignee to prevent it from getting squished on smaller screens


pdittaro (approved):
LGTM

PR #99, opened 2022-09-15
CICD Pipeline

pdittaro:
This work captures the initial end-to-end CD pipeline into the development environment.

On push to main branch:
- Build API, Frontend, and Migrate Images and push them to the internal Github Container Registry
- Import the Image Stream in OpenShift to reference the latest version of these images in preparation for,
- Tagging the images in the registry with the specific 'Github Action Run Number'
- Then, create a workflow dispatch in the bcgov-c gitops repository which includes the image names to be updated
- A seperate github action updates the appropriate namespace patches to reference these new images
- Once this has been commited (by the action), ArgoCD will pick up the changes and apply the latest images, including running a database migration job before the Sync action


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,51 @@
+name: build
+
+on: 
+  workflow_call:
```
dhaselhan, line 4: What does this do?

dhaselhan (approved):
LGTM 👍

mhuseinov (commented):
Would be nice to get some readme file that could provide some overview to the deployment structure

mhuseinov (approved):

.github/workflows/deploy.yml
pdittaro: Only runs when something explictly calls it, see the cd.yaml file

pdittaro: If you mean the environment input, right now its specifying which dispatch event to trigger on the child repo

PR #100, opened 2022-09-15
Update Create Card UI and Standardize Dates

dhaselhan:
* Add date format and provider to shared module
* Remove dayjs and use moment everywhere
* Update UI for create card
* Add !important for small fix on board UI
* Make region required in UI and DB


mhuseinov (approved):
🔥

PR #101, opened 2022-09-15
Add Timeline to Overview

dhaselhan:
* Create generic timeline component
* Add to Overview and add mapping logic

### DIFF
--- a/alcs-frontend/src/app/features/application/overview/overview.component.ts
+++ b/alcs-frontend/src/app/features/application/overview/overview.component.ts
@@ -1,15 +1,106 @@
 import { Component, OnInit } from '@angular/core';
+import { combineLatestWith, merge, tap } from 'rxjs';
+import { ApplicationDetailService } from '../../../services/application/application-detail.service';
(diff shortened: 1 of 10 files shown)


alcs-frontend/src/app/features/application/overview/overview.component.ts
```diff
@@ -1,15 +1,106 @@
...
+
+    if (application.decisionDate) {
+      mappedEvents.push({
+        name: 'Decision Made',
+        startDate: new Date(application.decisionDate),
+        isFulfilled: true,
+      });
+    }
+
+    meetings.sort((a, b) => a.startDate - b.startDate);
```
mhuseinov, line 87: I can see that bellow you are using getTime for sorting is intentional for omitting it here?

mhuseinov (approved):
💯

alcs-frontend/src/app/features/application/overview/overview.component.ts
dhaselhan: These are still numbers from the API, I convert them to Dates while dolng the mapping.

PR #103, opened 2022-09-20
Feature/alcs 48 - Separate Card from Application

mhuseinov:
* new Card entity
* address application features based on the new Card entity
     * assignee change
     * status change
     * comments
     * subtasks

NOTE! following pieces will be updated in next MR:
* file naming change on BE
* separating card entities from application module
* updating code to use cardUuid where applicable
* addressing default value for cardType


alcs-api/src/comment/comment.service.ts
```diff
@@ -83,36 +84,41 @@ export class CommentService {
...
+      relations: {
+        ...this.DEFAULT_COMMENT_RELATIONS,
+      },
     });
 
     if (body.trim() === '') {
       throw new ServiceValidationException('Comment body must be filled.');
     }
 
-    if (comment.body.trim() === body.trim()) {
```
dhaselhan, line 93: Why is this check being removed?

alcs-frontend/package-lock.json
```diff
@@ -13984,13 +13984,15 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/@csstools/postcss-unset-value/-/postcss-unset-value-1.0.2.tgz",
       "integrity": "sha512-c8J4roPBILnelAsdLr4XOAR/GsTm0GJi4XpcfvoWk3U6KiTCqiFYc63KhRMQQX35jYMp4Ao8Ij9+IZRgMfJp1g==",
-      "dev": true
+      "dev": true,
+      "requires": {}
```
dhaselhan, line 13988: Very odd that every time I run npm i it removes the empty requires, and yours adds it back.

dhaselhan (approved):
LGTM 👍

alcs-api/src/comment/comment.service.ts
mhuseinov: initially, this check was added to avoid errors on entity updates if nothing changed. That was caused with date transformers that were addressed previously so this one is redundant

alcs-frontend/package-lock.json
mhuseinov: yes, I noticed that too, but I think we using the same npm version

PR #105, opened 2022-09-22
Add Meeting Overview to Commissioner home page

dhaselhan:
* Move meeting-overview to shared components

### DIFF
--- a/alcs-frontend/src/app/features/home/home.component.ts
+++ b/alcs-frontend/src/app/features/home/home.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { AuthenticationService, ICurrentUser } from '../../services/authentication/authentication.service';
+import { AuthenticationService, ICurrentUser, ROLES } from '../../services/authentication/authentication.service';
 
 @Component({
   selector: 'app-home',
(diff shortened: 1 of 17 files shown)


alcs-frontend/src/app/features/home/home.component.ts
```diff
@@ -9,17 +9,26 @@ import { AuthenticationService, ICurrentUser } from '../../services/authenticati
...
 
-    this.hasGIS = this.currentUser.client_roles ? this.currentUser.client_roles.includes('GIS') : false;
-
     this.authService.$currentUser.subscribe((currentUser) => {
       if (currentUser) {
         this.currentUser = currentUser;
+
+        this.hasGIS =
+          currentUser.client_roles && currentUser.client_roles.length === 1
+            ? currentUser.client_roles.includes(ROLES.GIS)
```
mhuseinov, line 25: does this mean that users with only 1 role = GIS will be able to get hasGIS?

dhaselhan: This is to check if the are ONLY GIS, since the current requirements you should only see one thing on your home page.

mhuseinov (approved):

PR #107, opened 2022-09-22
Bug Fixes

dhaselhan:
* Apply missing migrations
* Prevent end date from defaulting to 0
* Log level is warn not warning

### DIFF
--- a/alcs-api/config/default.json
+++ b/alcs-api/config/default.json
@@ -8,7 +8,7 @@
   "DATABASE_USER": "postgres",
   "DATABASE_PASSWORD": "postgres",
   "DATABASE_MIGRATION_TABLE": "migrations",
-  "LOG_LEVEL": "warning",
+  "LOG_LEVEL": "warn",
   "FRONTEND_ROOT": "http://localhost:4200",
   "KEYCLOAK": {
     "AUTH_SERVER": "dev.loginproxy.gov.bc.ca",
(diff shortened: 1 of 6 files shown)


alcs-api/src/providers/typeorm/migrations/1663876950264-fixIssues.ts
```diff
@@ -0,0 +1,47 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class fixIssues1663876950264 implements MigrationInterface {
+  name = 'fixIssues1663876950264';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "application_subtask" DROP COLUMN "in_progress"`,
```
mhuseinov, line 8: oh..., looks like I'll need to move this to new subtask entity

mhuseinov (approved):

alcs-api/config/default.json
```diff
@@ -8,7 +8,7 @@
   "DATABASE_USER": "postgres",
   "DATABASE_PASSWORD": "postgres",
   "DATABASE_MIGRATION_TABLE": "migrations",
-  "LOG_LEVEL": "warning",
+  "LOG_LEVEL": "warn",
```
pdittaro, line 11: Note that this will have to be updated on each environments config map in order to be applied in a deployed environment

PR #108, opened 2022-09-22
All the bug fixes

dhaselhan:
* Add paused message to dialog
* Style save and cancel buttons
* Unify imports to not have .scss
* Improve error message to include meeting type
* Remove line from timeline if only a single entry
* Add Other: to radio input
* Move info request to 3rd menu position
* Fix date format in prod env

### DIFF
--- a/alcs-frontend/src/environments/environment.prod.ts
+++ b/alcs-frontend/src/environments/environment.prod.ts
@@ -8,7 +8,7 @@ export const environment = {
   production: true,
   apiUrl: window.env?.['apiUrl'] || 'http://localhost:8080',
   homeUrl: window.env?.['homeUrl'] || '/home',
-  dateFormat: 'YYYY-MMM-dd',
+  dateFormat: 'YYYY-MMM-DD',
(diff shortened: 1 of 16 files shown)


alcs-frontend/src/environments/environment.prod.ts
```diff
@@ -8,7 +8,7 @@ export const environment = {
   production: true,
   apiUrl: window.env?.['apiUrl'] || 'http://localhost:8080',
   homeUrl: window.env?.['homeUrl'] || '/home',
-  dateFormat: 'YYYY-MMM-dd',
+  dateFormat: 'YYYY-MMM-DD',
```
pdittaro, line 11: 'Th'/'Fr' fix?

pdittaro (approved):
LGTM

alcs-frontend/src/environments/environment.prod.ts
dhaselhan: yep! Looks like I just forgot to update the prod file

PR #110, opened 2022-09-22
Make pause nullable, filter correct boards

dhaselhan:
* Should remove vett instead of exec
* Make application_pause nullable so we can delete meetings


mhuseinov (approved):
💯

PR #112, opened 2022-09-23
Feature/alcs 147

mhuseinov:
* move to card-module:
  * card
  * card-status
  * card-type 
  * card-history
  * subscriber
* allow all users to update their settings
* fix card-history end date


dhaselhan (approved):
LGMT 👍

PR #117, opened 2022-09-26
Add Search Bar to Meetings Component

dhaselhan:


alcs-frontend/src/styles.scss
```diff
@@ -1,7 +1,6 @@
 // Custom Theming for Angular Material
 // For more information: https://material.angular.io/guide/theming
 @use '@angular/material' as mat;
-@use '@matheo/datepicker/theming' as datepicker;
```
mhuseinov, line 4: Why do you need to remove this one? I remember from the documentation that theming is required, and it can be adjusted as any angular material

mhuseinov (approved):

alcs-frontend/src/styles.scss
dhaselhan: I was getting duplicate style errors, in theory the mtx / material extensions library says to remove this when you add that styling as it should include both, however I noticed without this the inline date picker has some issues. Reverting this as I'd rather have duplicate style warnings than a broken date picker.

PR #123, opened 2022-10-03
Card Bug Fixes

dhaselhan:
* Use matToolTip instead of title and add missing tooltips
* Change dialogs to be wider, add maxHeight so they overflow properly
* Remove overflow from comments as its now scrolling the whole card
* Add fullName to GIS subtask circles
* Change home page to show multiple components if user has multiple roles


mhuseinov (approved):
🔥

PR #124, opened 2022-10-03
Add Proposal Summary

dhaselhan:
* Add summary to application
* Update inline-edit component to newer styling
* Add to Overview page and adjust margins

### DIFF
--- a/alcs-api/src/application/application.dto.ts
+++ b/alcs-api/src/application/application.dto.ts
@@ -7,6 +7,7 @@ import {
   IsOptional,
   IsString,
   IsUUID,
+  MaxLength,
 } from 'class-validator';
 import { UserDto } from '../user/user.dto';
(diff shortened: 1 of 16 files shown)


alcs-api/src/application/application.dto.ts
```diff
@@ -193,4 +197,9 @@ export class ApplicationUpdateDto {
   @IsBoolean()
   @IsOptional()
   highPriority?: boolean;
+
+  @AutoMap()
+  @IsString()
+  @IsOptional()
+  summary?: string;
```
mhuseinov, line 204: is there an option to limit length?

dhaselhan: Added!

mhuseinov (approved):
🎯

PR #129, opened 2022-10-06
Add new fields to decisions + update recon create card styling

dhaselhan:


alcs-api/src/application/application-decision/application-decision.entity.ts
```diff
@@ -18,6 +18,18 @@ export class ApplicationDecision extends Base {
   @Column({ type: 'timestamptz' })
   date: Date;
 
+  @AutoMap()
+  @Column({ type: 'timestamptz', nullable: true })
+  auditDate: Date;
```
mhuseinov, line 23: This can be misleading, currently, all fields prefixed with 'audit' are for technical auditing

mhuseinov (approved):

alcs-api/src/application/application-decision/application-decision.entity.ts
dhaselhan: The label for it is Audit Date in the ui, any suggestions for what to call it?

PR #130, opened 2022-10-07
Fix Provision Page + Split Application Create/Update

dhaselhan:


alcs-api/src/application/application.service.ts
```diff
@@ -57,26 +66,82 @@ export class ApplicationService {
...
 
-    const updatedApp = Object.assign(existingApplication, application);
+    return this.update(existingApplication, updates);
+  }
+
+  async update(
+    existingApplication: Application,
+    updates: ApplicationUpdateServiceDto,
+  ): Promise<Application> {
+    //TODO: Actual Update Logic
```
mhuseinov, line 140: is this still relevant

mhuseinov (approved):

alcs-api/src/application/application.service.ts
dhaselhan: oh snap, maybe! I was thinking of moving away from object.assign, Ill go back to it.

dhaselhan: Reviewed how we are using this, since its using its own specific type that is'nt the same as the controller, I think using Object.assign is okay.

PR #131, opened 2022-10-07
Bug Fixes + Cleanups

dhaselhan:
* Change startOfDay to use actual pacific start of day time
* Remove duplicated assignee from application, use one on card
* Remove accidental highPriority from applicationdto and use one from card


mhuseinov (approved):
🎯

PR #134, opened 2022-10-07
Add notification sent date + add assignee name to avatar circle on board

dhaselhan:


alcs-api/src/providers/typeorm/migrations/1665175956310-rename_decision_to_review.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class renameDecisionToReview1665175956310 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    queryRunner.query(`
+        UPDATE public."card_status" SET "label" ='Ready for Review Sent / Going to Next Review Meeting' WHERE "uuid" = '64944bb8-f2f2-4709-9062-214f5c4d3187'
```
mhuseinov, line 6: This can cause issues with displaying column headers. Words may be split strangely in column header.

alcs-frontend/src/app/features/application/overview/overview.component.ts
```diff
@@ -89,6 +89,14 @@ export class OverviewComponent implements OnInit {
       });
     }
 
+    if (application.notificationSentDate) {
+      mappedEvents.push({
+        name: "'Ready for Review' Notification Sent to Applicant",
```
mhuseinov, line 94: Do we need to adjust the casing in name?

alcs-frontend/src/app/features/application/review/review.component.html
```diff
@@ -1,4 +1,8 @@
 <h3>Review</h3>
+<div class="notification-date">
+  <div class="subheading1">'Ready for Review' Notification Sent to Applicant</div>
```
mhuseinov, line 3: Same with casing

alcs-frontend/src/app/features/application/overview/overview.component.ts
dhaselhan: Everything else in the timeline is currently title cased, so this follows that pattern

alcs-frontend/src/app/features/application/review/review.component.html
dhaselhan: This is a heading, so I think title case is correct.

alcs-api/src/providers/typeorm/migrations/1665175956310-rename_decision_to_review.ts
dhaselhan: Good idea, updated with some nbsp; to break where we want.

mhuseinov: The only catch maybe with the workflow dropdown if that one is using "text" for rendering

mhuseinov (approved):
🔥

PR #145, opened 2022-10-14
Add new Decision fields

dhaselhan:
* Add Decision Maker, Ceo Criterion, and isTimeExtension fields with special logic
* Add no documents when no documents are uploaded
* Start refactor of code tables

### DIFF
--- a/alcs-api/src/application/application-decision/application-decision.controller.ts
+++ b/alcs-api/src/application/application-decision/application-decision.controller.ts
@@ -19,17 +19,19 @@ import { RoleGuard } from '../../common/authorization/role.guard';
 import { ANY_AUTH_ROLE } from '../../common/authorization/roles';
 import { UserRoles } from '../../common/authorization/roles.decorator';
 import { ApplicationService } from '../application.service';
-import { ApplicationDecisionOutcomeType } from './application-decision-outcome.entity';
+import { DecisionOutcomeCode } from './application-decision-outcome.entity';
 import {
   ApplicationDecisionDto,
   ApplicationDecisionOutcomeTypeDto,
   CreateApplicationDecisionDto,
-  DecisionDocumentDto,
   UpdateApplicationDecisionDto,
 } from './application-decision.dto';
 import { ApplicationDecision } from './application-decision.entity';
 import { ApplicationDecisionService } from './application-decision.service';
-import { DecisionDocument } from './decision-document.entity';
+import { CeoCriterionCodeDto } from './ceo-criterion/ceo-criterion.dto';
+import { CeoCriterionCode } from './ceo-criterion/ceo-criterion.entity';
+import { DecisionMakerCodeDto } from './decision-maker/decision-maker.dto';
+import { DecisionMakerCode } from './decision-maker/decision-maker.entity';
 
 @ApiOAuth2(config.get<string[]>('KEYCLOAK.SCOPES'))
 @Controller('application-decision')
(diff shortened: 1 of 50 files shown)


alcs-api/src/application/application-decision/application-decision.controller.ts
```diff
@@ -43,28 +45,39 @@ export class ApplicationDecisionController {
...
-    const mappedCodes = this.mapper.mapArray(
-      codes,
-      ApplicationDecisionOutcomeType,
-      ApplicationDecisionOutcomeTypeDto,
-    );
+  }
 
+  @Get('/codes')
+  @UserRoles(...ANY_AUTH_ROLE)
+  async getCodes() {
```
mhuseinov, line 63: With this code here we have a new pattern now. There is CodeService that aggregates all code types and here we have entity-specific types. So do you want to keep these separate? What if we need one of this type on board, should we duplicate it there?

alcs-api/src/application/application-decision/application-decision.controller.ts
```diff
@@ -43,28 +45,39 @@ export class ApplicationDecisionController {
...
+    const codes = await this.appDecisionService.fetchCodes();
     return {
-      decisions: mappedDecisions,
-      codes: mappedCodes,
+      outcomes: this.mapper.mapArray(
```
mhuseinov, line 66: not a required thing, but I prefer async mapping for arrays

alcs-api/src/application/application-decision/application-decision.dto.ts
```diff
@@ -1,71 +1,99 @@
...
+  @AutoMap()
+  chairReviewRequired: boolean;
 
   @AutoMap()
-  @IsNumber()
-  @IsOptional()
   auditDate?: number | null;
 
   @AutoMap()
-  @IsOptional()
```
mhuseinov, line 56: 👍

alcs-api/src/application/application-decision/application-decision.service.ts
```diff
@@ -95,30 +106,40 @@ export class ApplicationDecisionService {
       );
     }
```
mhuseinov, line 108: I wonder if it is possible to break this method into smaller parts

alcs-api/src/application/application-decision/decision-maker/decision-maker.dto.ts
```diff
@@ -0,0 +1,5 @@
+import { Entity } from 'typeorm';
+import { BaseCodeDto } from '../../../common/dtos/base.dto';
+
+@Entity()
+export class DecisionMakerDto extends BaseCodeDto {}
```
mhuseinov, line 5: DecisionMakerCodeDto ?

alcs-frontend/src/app/features/application/decision/decision.component.html
```diff
@@ -25,29 +25,49 @@ <h3>Decision</h3>
...
-      </span>
+      <div class="subheading2">Decision Maker</div>
+      {{ decision.decisionMaker?.label }}
     </div>
+    <div>
+      <div *ngIf="decision.ceoCriterion" class="subheading2">CEO Criterion</div>
+      {{ decision.ceoCriterion?.label }}
+      <ng-container *ngIf="decision.ceoCriterion?.code === 'MODI'">
+        {{ decision.isTimeExtension ? '- Time extension' : '- Other' }}
+      </ng-container>
```
mhuseinov, line 40: Is it because decision.isTimeExtension can be null/undefined ?

alcs-api/src/application/application-decision/application-decision.controller.ts
dhaselhan: As we move away from needing to map the code using the code service to get the label, we might even end up deleting the main CodeService. Now you will only need the full list of codes in places where you can edit them or need to display them all (boards). I made this independent since they are only needed in the create/edit dialog of decisions, you can display them on the board since it will eventually return the full dto without needing the full list.

alcs-frontend/src/app/features/application/decision/decision.component.html
dhaselhan: Ill be honest Im not 100% sure anymore, I could replace this with decision.isTimeExtension !== null for the *ngIf.

alcs-api/src/application/application-decision/application-decision.controller.ts
dhaselhan: Done!

alcs-api/src/application/application-decision/application-decision.service.ts
dhaselhan: I cleaned up the check at the top but couldnt find any clean way to split up the logic.

alcs-api/src/application/application-decision/decision-maker/decision-maker.dto.ts
dhaselhan: Done!

alcs-frontend/src/app/features/application/decision/decision.component.html
dhaselhan: Changed to a null check.

mhuseinov (approved):

PR #147, opened 2022-10-14
Improve Token Handling

dhaselhan:
* Auto refresh token every 4 minutes as long as refresh token isnt expired
* If refresh token becomes expired, redirect user to login


mhuseinov (approved):
🔥

PR #152, opened 2022-10-18
Add Go to Card Menu and Organize dialogs

dhaselhan:
* Structure per object type and rename to create more consistency
* Add cleanup to board component to prevent duplicate HTTP requests


mhuseinov (approved):
🔥

PR #155, opened 2022-10-18
ALCS-177: send email on register user if no roles

mhuseinov:
* send an email on register user if no roles
* added await in exchangeCodeForToken


dhaselhan (approved):
LGTM 👍

PR #157, opened 2022-10-19
Refactor Code Tables

dhaselhan:
* Remove ObsoleteBaseCodeEntity
* Change all tables to use code directly
* Update API and DTOs to match


alcs-api/src/application/application.controller.spec.ts
```diff
@@ -408,6 +370,13 @@ describe('ApplicationController', () => {
       card: updatedCard,
     } as Application);
 
+    applicationService.mapToDtos.mockResolvedValue([
+      {
+        ...mockApplicationEntity,
+        card: updatedCard as any,
```
mhuseinov, line 376: why do you need any everywhere?

alcs-api/src/code/code.service.spec.ts
```diff
@@ -144,7 +140,7 @@ describe('CodeService', () => {
     const mockStatus = {
       uuid: '2',
     };
-    mockStatusRepository.findOne.mockResolvedValue(mockStatus as CardStatus);
+    mockStatusRepository.findOne.mockResolvedValue(mockStatus as any);
```
mhuseinov, line 143: the same with any

alcs-api/src/application/application.controller.spec.ts
dhaselhan: Removed

alcs-api/src/code/code.service.spec.ts
dhaselhan: Changed back to as CardStatus and fixed.

mhuseinov (approved):
🔥

PR #160, opened 2022-10-20
Subtask Cleanup

dhaselhan:
* Introduce AssigneeDTO and map on backend rather than on front-end
* Pull the subtask type into its own DTO to stay consistent with other codes
* Renamed to card-subtask on frontend


alcs-frontend/src/app/features/home/gis/gis-subtasks.component.ts
```diff
@@ -15,22 +14,17 @@ import { UserService } from '../../../services/user/user.service';
...
-  private statuses: CardStatusDto[] = [];
   public gisUsers: UserDto[] = [];
 
   constructor(
     private homeService: HomeService,
     private applicationService: ApplicationService,
     private userService: UserService,
     private router: Router,
-    private applicationSubtaskService: ApplicationSubtaskService
+    private applicationSubtaskService: CardSubtaskService
```
mhuseinov, line 24: since the type changed maybe the name also needs to change: cardSubtaskService?

mhuseinov (approved):

alcs-frontend/src/app/features/home/gis/gis-subtasks.component.ts
dhaselhan: Good catch!

PR #161, opened 2022-10-20
Feature/alcs 121 part 3

mhuseinov:
* removed Active days if the application has reconsiderations
* added application type to reconsideration card
* add reconsiderations to timeline
* added reconsiderations to home page in assigned to me


alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -347,11 +348,13 @@ export class BoardComponent implements OnInit, OnDestroy {
   }
 
   private mapReconsiderationDtoToCard(recon: ApplicationReconsiderationDto): CardData {
+    const mappedType = this.applicationTypes.find((type) => type.code === recon.application.type.code);
```
dhaselhan, line 351: If you already have the full type under recon.application.type, why do you need to find it?

PR #162, opened 2022-10-20
Refactor Subtasks for Homepage + Add Planning Review

dhaselhan:


alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -105,6 +106,59 @@ export class BoardComponent implements OnInit, OnDestroy {
     this.destroy.complete();
   }
 
+  async onSelected(card: CardSelectedEvent) {
+    switch (card.cardType) {
```
dhaselhan, line 110: I should move setUrl to here and then I can delete it from each function below.

alcs-api/src/application/application.service.ts
```diff
@@ -199,7 +199,7 @@ export class ApplicationService {
     return application;
   }
 
-  async getAllApplicationsWithIncompleteSubtasks(subtaskType: string) {
+  async getBySubtaskType(subtaskType: string) {
```
mhuseinov, line 202: It still uses the incomplete filter, I think it should be reflected somehow

alcs-api/src/planning-review/planning-review.service.ts
```diff
@@ -108,4 +108,27 @@ export class PlanningReviewService {
       relations: this.DEFAULT_RELATIONS,
     });
   }
+
+  async getBySubtaskType(subtaskType: string) {
```
mhuseinov, line 112: the same about incomplete filter

alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -105,6 +106,60 @@ export class BoardComponent implements OnInit, OnDestroy {
...
+        await this.openAppCardDetailDialog(card.uuid);
+        break;
+      case CardType.RECON:
+        await this.openReconCardDetailDialog(card.uuid);
+        break;
+      case CardType.PLAN:
+        await this.openPlanningCardDialog(card.uuid);
+        break;
+      default:
+        console.error('Card type is not configured for a dialog');
```
mhuseinov, line 122: Should we show an error toast to user "Failed to open card". I know that this should not happen but if we already providing a error message here

alcs-api/src/application/application.service.ts
dhaselhan: Renamed

alcs-api/src/planning-review/planning-review.service.ts
dhaselhan: Renamed

alcs-frontend/src/app/features/board/board.component.ts
dhaselhan: Done!

PR #171, opened 2022-10-21
Feature/alcs 197

mhuseinov:
* added no data to assigned to me
* added no data to GIS subtasks

### DIFF
--- a/alcs-frontend/src/app/features/home/assigned/assigned.component.scss
+++ b/alcs-frontend/src/app/features/home/assigned/assigned.component.scss
@@ -26,8 +26,17 @@
   }
 
   tr:hover {
-    box-shadow: 0 0 0 1px rgba($primary-color, .9);
+    box-shadow: 0 0 0 1px rgba($primary-color, 0.9);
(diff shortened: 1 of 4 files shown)


alcs-frontend/src/app/features/home/assigned/assigned.component.scss
```diff
@@ -26,9 +26,19 @@
   }
 
   tr:hover {
-    box-shadow: 0 0 0 1px rgba($primary-color, .9);
+    box-shadow: 0 0 0 1px rgba($primary-color, 0.9);
```
dhaselhan, line 29: Was this auto-format or your change? I've been writing them as .2 just wondering if I should always include the 0.

dhaselhan (approved):
LGTM 👍

alcs-frontend/src/app/features/home/assigned/assigned.component.scss
mhuseinov: I do not think that is a hard requirement, but yes, it is automatic formatting

PR #172, opened 2022-10-24
UX Updates

dhaselhan:
* Change layout overflow so that board and sticky headers scroll bar nicely
* Add border color to labels so we have more visual options
* Add loading indicator to decisions for attach/delete file/delete decision
* Change comments white-space so it preserves line breaks
* Add disable-able tooltips to Favourite button


mhuseinov (approved):
💯

PR #177, opened 2022-10-26
User Cleanup

dhaselhan:
* Change all lookups to use either our uuid or bceidguid/idirguid
* Add constraint to ensure bceidguid and idirguid are unique
* Rename RoleGuard -> RolesGuard to ensure we are using the correct one, as there is a RoleGuard from Keycloak
* Update UserDTO to only return currently used fields
* Add new update dto to make it clear only settings can be updated
* Change user update to be consistent with other controllers, taking in uuid and dto separately

### DIFF
--- a/alcs-api/src/common/automapper/user.automapper.profile.ts
+++ b/alcs-api/src/common/automapper/user.automapper.profile.ts
@@ -3,7 +3,8 @@ import { AutomapperProfile, InjectMapper } from '@automapper/nestjs';
 import { Injectable } from '@nestjs/common';
 import {
   AssigneeDto,
-  CreateOrUpdateUserDto,
+  CreateUserDto,
+  UpdateUserDto,
   UserDto,
   UserSettingsDto,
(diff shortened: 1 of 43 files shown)


alcs-api/src/common/automapper/user.automapper.profile.ts
```diff
@@ -17,36 +18,50 @@ export class UserProfile extends AutomapperProfile {
...
+          mapFrom((u) => {
+            if (u.givenName && u.familyName) {
+              return (
+                u.givenName?.charAt(0).toUpperCase() +
+                u.givenName?.slice(1) +
+                u.familyName?.charAt(0).toUpperCase() +
+                u.familyName?.slice(1)
+              );
+            } else {
+              //TODO: how do mentions work for bceid users?
```
mhuseinov, line 52: I think mentions not relevant for bceid users since only staff members can work on board

mhuseinov (approved):
👍

PR #179, opened 2022-10-27
Add Amendments and Amendment Cards

dhaselhan:
* Fix pills on all card dialogs


alcs-api/src/application-amendment/application-amendment.controller.ts
```diff
@@ -0,0 +1,88 @@
...
+    const amendment = await this.amendmentService.create(createDto, board);
+
+    return this.amendmentService.mapToDtos([amendment]);
+  }
+
+  @Delete('/:uuid')
+  @UserRoles(...ROLES_ALLOWED_APPLICATIONS)
+  async delete(@Param('uuid') uuid: string) {
+    await this.amendmentService.delete(uuid);
+    return { deleted: true };
```
mhuseinov, line 60: Do you really need this? I do not think we return any values on delete on other endpoints

alcs-api/src/application-amendment/application-amendment.service.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+      relations: DEFAULT_RELATIONS,
+    };
+
+    await service.getByBoardCode(fakeBoardCode);
+
+    expect(amendmentRepoMock.find).toBeCalledWith(findOptions);
+  });
+
+  it('should successfully create application and amendment card if app does not exist', async () => {
+    const mockamendmentCreateDto = {
```
mhuseinov, line 103: You can move this into separate function and reuse in other places

alcs-api/src/application-amendment/application-amendment.service.ts
```diff
@@ -0,0 +1,185 @@
...
+    const amendment = await this.getByUuidOrFail(uuid);
+    return this.amendmentRepository.softRemove([amendment]);
+  }
+
+  private async getByUuidOrFail(uuid: string) {
+    const amendment = await this.amendmentRepository.findOneByOrFail({
+      uuid,
+    });
+
+    if (!amendment) {
```
mhuseinov, line 134: I think this will not execute. The line above should throw an exception if there is no amendment

alcs-api/src/home/home.controller.ts
```diff
@@ -167,4 +183,23 @@ export class HomeController {
     }
     return result;
   }
+
+  private mapAmendmentsToDTOs(amendments: ApplicationAmendment[]) {
```
mhuseinov, line 187: a nitpick 'mapAmendmentsToDtos' so it will match other paces

alcs-frontend/src/app/features/home/gis/gis-subtasks.component.ts
```diff
@@ -39,14 +39,17 @@ export class GisSubtasksComponent implements OnInit {
     const applications = nonOrderedSubtasks.filter((s) => s.card.type === CardType.APP);
     const reconsiderations = nonOrderedSubtasks.filter((s) => s.card.type === CardType.RECON);
     const planningReviews = nonOrderedSubtasks.filter((s) => s.card.type === CardType.PLAN);
+    const amendments = nonOrderedSubtasks.filter((s) => s.card.type === CardType.AMEND);
 
     this.subtasks.push(
       // high priority
       ...applications.filter((a) => a.card.highPriority).sort((a, b) => b.activeDays! - a.activeDays!),
+      ...amendments.filter((r) => r.card.highPriority).sort((a, b) => a.createdAt! - b.createdAt!),
```
mhuseinov, line 47: so amendments need to show up before recons?

alcs-api/src/application-amendment/application-amendment.controller.ts
dhaselhan: This was copied from the reconsideration controller, but its done because Angular expects a JSON response by so this was the easiest thing to return that made sense.

alcs-frontend/src/app/features/home/gis/gis-subtasks.component.ts
dhaselhan: correct, as per the ticket amendments should be below applications and above recons.

alcs-api/src/application-amendment/application-amendment.service.spec.ts
dhaselhan: Good catch, done! Moved to beforeEach and re-used.

alcs-api/src/application-amendment/application-amendment.service.ts
dhaselhan: Ah good point, changed to findOneBy and throw this exception so we can control the exception type and error.

alcs-api/src/home/home.controller.ts
dhaselhan: fixed

mhuseinov (approved):
💯

PR #180, opened 2022-10-27
Commissioner Fixes

dhaselhan:
* Change meetings page to go directly to review page if not a comissioner
* Use setters so that when the input changes the logic is re-ran


mhuseinov (approved):
👍

PR #183, opened 2022-11-01
Add import service and new command

dhaselhan:
* Add virtual fs libraries to dev dependencies to help with file system mocking


alcs-api/src/import/import.service.ts
```diff
@@ -0,0 +1,383 @@
...
+          );
+          if (!localGovernment) {
+            this.logger.error(
+              `${mappedRow.fileNumber}: Failed to load government ${sheetType.localGovernment}`,
+            );
+            return;
+          }
+        }
+
+        const application = await this.applicationService.create({
```
mhuseinov, line 163: So this is going to create an application in DB for every single record from csv?

mhuseinov (approved):

PR #185, opened 2022-11-02
Feature/alcs 211

mhuseinov:
* Oracle DB container
* Oracle schema and data migration to Postgres

### DIFF
--- a/alcs-api/OracleToPg/README.MD
+++ b/alcs-api/OracleToPg/README.MD
@@ -1,0 +1,107 @@
+# Oracle in Docker
+
+## Prerequisites
+
+- Docker
+- Oracle account
+- SQLDeveloper tool
+- At least 4GB of memory allocated to docker
+
+## Oracle 12.2 in Docker container
+
+- login to Oracle Docker registry with your oracle account
+
+```sh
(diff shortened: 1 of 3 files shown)


alcs-api/OracleToPg/README.MD
```diff
@@ -0,0 +1,266 @@
...
+   -v <path on the host to config folder with ora2pg.conf>:/config \
+   -v <path on the host to data folder where you will see the output of migration>:/data \
+   georgmoser/ora2pg
+
+   # example of command above with populated values
+   docker run \
+   --name ora2pg \
+   --network alpine-net \
+   -it \
+   -v /Users/mekhti/Documents/Bit3/sources/alcs/alcs-api/OracleToPg/ora2pg/config:/config \
```
dhaselhan, line 171: Might want to genericize this or put in a placeholder rather than your username

dhaselhan (approved):
LGTM 👍

alcs-api/OracleToPg/README.MD
mhuseinov: I've added the template/explanation above, but sure I can update it so this one will be more generic. I'll just use 'user' instead

PR #205, opened 2022-11-09
Move Decision Meetings to Decisions Module

dhaselhan:
* Load Decision Meetings for both Applications and Recons
* Rename AppService to getMany to be consistent with other services
* Turn on strict null checks


alcs-api/src/application/application-meeting/application-meeting.controller.spec.ts
```diff
@@ -149,26 +148,4 @@ describe('ApplicationMeetingController', () => {
       reportEndDate: meetingToUpdate.reportEndDate,
     });
   });
-
-  it('should fail create meeting if meeting type does not exist', async () => {
```
pdittaro, line 153: Why is this test no longer relevant?

dhaselhan: Before my change, this was using .find() meaning we had our own check if it was null and throw our own error. Now that its using findOrFail, it will automatically throw an exception if it does not exist. So since the logic is now part of typeorm (and would need to be mocked in), we don't need a unit test for it.

alcs-api/src/application/application.service.ts
```diff
@@ -142,27 +142,17 @@ export class ApplicationService {
...
-    const updatedApp = Object.assign(existingApplication, updates);
-    await this.applicationRepository.save(updatedApp);
-
-    return this.applicationRepository.findOne({
-      where: {
-        uuid: updatedApp.uuid,
-      },
-      relations: this.DEFAULT_RELATIONS,
-    });
+    await this.applicationRepository.update(existingApplication.uuid, updates);
```
mhuseinov, line 145: so this is the spot where save fails because of soft deleted card, correct?

alcs-api/src/decision/application-amendment/application-amendment.entity.ts
```diff
@@ -28,16 +28,16 @@ export class ApplicationAmendment extends Base {
...
+  isReviewApproved: boolean | null;
 
   @AutoMap()
   @Column()
   isTimeExtension: boolean;
 
   @AutoMap()
   @Column({ type: 'timestamptz', nullable: true })
-  reviewDate: Date;
+  reviewDate: Date | null;
```
mhuseinov, line 40: Is it always supplied? Do we need to include undefined so it the property will be left unchanged if it is not provided?

alcs-api/src/decision/application-decision/application-decision.entity.ts
```diff
@@ -71,23 +71,23 @@ export class ApplicationDecision extends Base {
...
+  decisionMakerCode: string | null;
 
   @AutoMap()
   @ManyToOne(() => CeoCriterionCode, { nullable: true })
   ceoCriterion?: CeoCriterionCode;
 
   @AutoMap()
   @Column({ type: 'text', nullable: true })
-  ceoCriterionCode?: string;
+  ceoCriterionCode: string | null;
```
mhuseinov, line 82: the same thing about undefined, so it is intentional that this one should always be supplied or null?

mhuseinov (approved):

PR #206, opened 2022-11-10
Set Import Dates to Vancouver Time

dhaselhan:
* Set imported times to beginning of Vancouver day
* Fix ApplicationDto mapping to use numbers

### DIFF
--- a/alcs-api/src/application/application.service.ts
+++ b/alcs-api/src/application/application.service.ts
@@ -267,7 +267,6 @@ export class ApplicationService {
       activeDays: appTimeMap.get(app.uuid).activeDays || 0,
       pausedDays: appTimeMap.get(app.uuid).pausedDays || 0,
       paused: appPausedMap.get(app.uuid) || false,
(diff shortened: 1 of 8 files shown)


alcs-api/src/application/application.service.ts
```diff
@@ -267,7 +267,6 @@ export class ApplicationService {
       activeDays: appTimeMap.get(app.uuid).activeDays || 0,
       pausedDays: appTimeMap.get(app.uuid).pausedDays || 0,
       paused: appPausedMap.get(app.uuid) || false,
-      hasReconsideration: app.reconsiderations?.length > 0 || false,
```
mhuseinov, line 270: I remember there was a UI code that was using this. Is it also gone?

mhuseinov (approved):

alcs-api/src/application/application.service.ts
dhaselhan: yep! This wasn't in the DTO on the front end so I don't think it was used and this is the only reference to hasReconsideration

PR #209, opened 2022-11-10
Show if a decision was recond/amended by another decision

dhaselhan:
* Add UI to indicate if a decisions has been modified or amended


mhuseinov (approved):
👍

PR #210, opened 2022-11-14
Rename Amendment -> Modification

dhaselhan:


alcs-api/src/board/board.controller.ts
```diff
@@ -60,9 +60,9 @@ export class BoardController {
       planningReviews = await this.planningReviewService.getCards();
     }
 
-    let amendments: ApplicationAmendment[] = [];
+    let modifications: ApplicationModification[] = [];
     if (boardCode === 'ceo') {
```
mhuseinov, line 64: We need gradually get rid of hardcoded values and switch to enums.

alcs-api/src/common/automapper/reconsideration.automapper.profile.ts
```diff
@@ -88,6 +88,10 @@ export class ReconsiderationProfile extends AutomapperProfile {
           (a) => a.card,
           mapFrom((rd) => this.mapper.map(rd.card, Card, CardDto)),
         ),
+        forMember(
```
mhuseinov, line 91: Do you really need this one? Docs mentions that automapper is able to understand boolean | null

alcs-frontend/src/app/features/application/decision/decision-dialog/decision-dialog.component.ts
```diff
@@ -246,7 +246,7 @@ export class DecisionDialogComponent implements OnInit {
   }
 
   onSelectPostDecision(postDecision: { type: string }) {
-    if (postDecision.type === 'amendment') {
+    if (postDecision.type === 'modification') {
```
mhuseinov, line 249: maybe an enum for this?

alcs-frontend/src/app/features/application/decision/decision.component.scss
```diff
@@ -46,7 +46,7 @@ section {
     background: $accent-color;
   }
 
-  &.amendment {
+  &.modification {
```
mhuseinov, line 49: 👍

mhuseinov (approved):

alcs-api/src/board/board.controller.ts
dhaselhan: Added enum for board codes

alcs-frontend/src/app/features/application/decision/decision-dialog/decision-dialog.component.ts
dhaselhan: Good call, cleaned up

alcs-api/src/common/automapper/reconsideration.automapper.profile.ts
dhaselhan: Yes, without this it is no longer mapping and doesnt return the property at all.

PR #224, opened 2022-11-22
update meeting sorting, fix inline edit on decision

mhuseinov:
- fixed inline edit on decision meeting
- changed scheduled discussion sorting

### DIFF
--- a/alcs-api/src/decision/application-decision/application-decision.service.ts
+++ b/alcs-api/src/decision/application-decision/application-decision.service.ts
@@ -154,8 +154,10 @@ export class ApplicationDecisionService {
     await this.validateDecisionChanges(updateDto);
 
     if (
-      existingDecision.resolutionNumber !== updateDto.resolutionNumber ||
-      existingDecision.resolutionYear !== updateDto.resolutionYear
+      updateDto.resolutionNumber &&
+      updateDto.resolutionYear &&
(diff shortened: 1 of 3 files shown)


alcs-api/src/decision/application-decision/application-decision.service.ts
```diff
@@ -154,8 +154,10 @@ export class ApplicationDecisionService {
     await this.validateDecisionChanges(updateDto);
 
     if (
-      existingDecision.resolutionNumber !== updateDto.resolutionNumber ||
-      existingDecision.resolutionYear !== updateDto.resolutionYear
+      updateDto.resolutionNumber &&
+      updateDto.resolutionYear &&
+      (existingDecision.resolutionNumber !== updateDto.resolutionNumber ||
+        existingDecision.resolutionYear !== updateDto.resolutionYear)
```
dhaselhan, line 160: Does this additional check do anything? If the updateDto.resolutionNumber or resolutionYear was undefined it would not equal the numbers in resolutionNumber/resolutionYear.

dhaselhan (approved):
LGTM 👍

alcs-api/src/decision/application-decision/application-decision.service.ts
mhuseinov: yep, the validate function does not care about the undefined values. Since the typeorm ignores the undefined value on update we do not need to run validation against it

PR #227, opened 2022-11-23
Add Portal Frontend

dhaselhan:
* Add Portal frontend and hook up to Auth
* Create header, footer, home page, login page


portal-frontend/src/app/shared/footer/footer.component.html
```diff
@@ -0,0 +1,11 @@
+<footer>
+  <div>
+    Provincial Agricultural Land Commission © 2022
```
mhuseinov, line 3: Lets make year populate dynamically

portal-frontend/src/app/shared/footer/footer.component.scss
```diff
@@ -0,0 +1,24 @@
+@import '../../../styles/colors.scss';
+@import '../../../styles/typography.scss';
+
```
mhuseinov, line 3: I guess we will adjust these once we get WCAG requirements

portal-frontend/src/app/shared/utils/date-format.ts
```diff
@@ -0,0 +1,16 @@
...
+    dateInput: 'DD.MM.YYYY',
+  },
+  display: {
+    dateInput: environment.dateFormat,
+    monthYearLabel: 'YYYY',
+    dateA11yLabel: 'LL',
+    monthYearA11yLabel: 'YYYY',
+    monthLabel: 'MMM',
+    monthDayLabel: 'MMM-DD',
+    popupHeaderDateLabel: 'YYYY-MMM-DD',
```
mhuseinov, line 14: now I wonder if we need to avoid using the '-' in date due to JS specification

portal-frontend/src/styles/typography.scss
```diff
@@ -0,0 +1,65 @@
+@import '@bcgov/bc-sans/css/BCSans.css';
```
mhuseinov, line 1: Could you add comment somewhere that we need to use relative values in typography

portal-frontend/src/app/shared/footer/footer.component.html
dhaselhan: Done!

portal-frontend/src/styles/typography.scss
dhaselhan: Done! Added a comment above the first h1

mhuseinov (approved):

PR #234, opened 2022-11-29
update resolution number type

mhuseinov:


dhaselhan (approved):
LGTM 👍 
As long as the migration is not already on main.

mhuseinov:
create-portal is not merged yet, so I do not think so

PR #239, opened 2022-11-30
Load Local Governments from Redis for Portal

dhaselhan:
* Add Redis Service to Portal
* Load Local Governments into Redis on ACLS Local Government Service Init
* Fetch Local Governments from Redis on Portal Side
* Remove form validation so empty application can be updated
* Split Update and Attach Document into two separate endpoints
* Add Toasts to Portal for Success/Failure


portal-frontend/src/app/features/edit-application/edit-application.component.ts
```diff
@@ -71,26 +70,38 @@ export class EditApplicationComponent implements OnInit {
...
+          this.localGovernment.setErrors({
+            invalid: true,
+          });
+          return;
+        }
+      }
+
+      await this.applicationService.updatePending(this.fileId, {
+        applicant: this.applicantName.getRawValue(),
+        localGovernmentUuid: localGovernment ? localGovernment.uuid : undefined,
```
mhuseinov, line 89: Maybe localGovernment?.uuid ?

portal-frontend/src/app/features/edit-application/edit-application.component.ts
```diff
@@ -71,26 +70,38 @@ export class EditApplicationComponent implements OnInit {
...
   async onSubmit() {
-    //RUN FORM VALIDATION
+    this.applicantName.setValidators([Validators.required]);
+    this.applicantName.updateValueAndValidity();
+
+    this.localGovernment.setValidators([Validators.required]);
+    this.localGovernment.updateValueAndValidity();
+
+    if (this.firstForm.valid) {
+      alert('App Submitted');
```
mhuseinov, line 103: is it for debugging?

portal-frontend/src/app/services/application/application.service.ts
```diff
@@ -10,37 +11,45 @@ import { ApplicationDto, UpdateApplicationDto } from './application.dto';
...
 
-  constructor(private httpClient: HttpClient) {}
+  constructor(private httpClient: HttpClient, private toastService: ToastService) {}
 
   async create() {
     try {
       return firstValueFrom(this.httpClient.post<{ fileId: string }>(`${this.serviceUrl}`, {}));
     } catch (e) {
-      //Do something
+      this.toastService.showErrorToast('Failed to create Application, please try again later');
```
mhuseinov, line 20: Looks like 'throw e;' is missing in the service.  If an error happens it will not stop the follow-up logic execute, is it intentional?

portal-frontend/src/app/services/toast/toast.service.ts
```diff
@@ -0,0 +1,32 @@
+import { Injectable } from '@angular/core';
+import { MatSnackBar } from '@angular/material/snack-bar';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ToastService {
+  constructor(private snackBar: MatSnackBar) {}
+
+  private TOAST_DURATION = 3000;
```
mhuseinov, line 10: Not in the scope of this ticket. I do not remember for sure, but I think, based on WCAG these durations must be adjustable by a user.

services/apps/portal/src/application/application.controller.ts
```diff
@@ -49,14 +52,37 @@ export class ApplicationController {
...
+    @Req() req,
+  ) {
+    //Verify User has Access
+    const existingApplication = this.applicationService.getByFileId(
+      fileId,
+      req.user.entity,
+    );
+
+    if (!existingApplication) {
+      throw new Error('Failed to find application with given File ID and User');
```
mhuseinov, line 67: Maybe ServiceNotFoundException instead of Error?

services/apps/portal/src/application/application.entity.ts
```diff
@@ -37,20 +37,21 @@ export class Application extends BaseEntity {
   @UpdateDateColumn({ type: 'timestamptz' })
   updatedAt: Date;
 
-  @AutoMap()
+  @AutoMap(() => String)
```
mhuseinov, line 40: 👍

portal-frontend/src/app/features/edit-application/edit-application.component.ts
dhaselhan: Kind of, this will eventually do something else but I wanted something to show when it passed and alert was the easiest thing.

portal-frontend/src/app/services/application/application.service.ts
dhaselhan: Any callers to this function will get the type Application | undefined and will need to handle the undefined case. Since this is showing an error toast it made more sense not to re-throw. We could alternatively have it not have any try catch and rely on anyone calling it to try/catch and show the error toast.

portal-frontend/src/app/features/edit-application/edit-application.component.ts
dhaselhan: Good idea, done!

services/apps/portal/src/application/application.controller.ts
dhaselhan: Done!

portal-frontend/src/app/services/application/application.service.ts
mhuseinov: I think it is fine for now, we probably will revisit this once we get the actual use case. As for me, it is more like what happens, for example with the get list endpoint that returns [] on an error and [] if there are no data to return so the result is not clear in that case. 
As for me, ideally, service methods can show generic errors related to their actions, and client that calls them can override those based on the requirements.

mhuseinov (approved):

portal-frontend/src/app/services/application/application.service.ts
dhaselhan: Yeah its kinda a stuck in between situation, in some cases its nice to have the service catch and show the toast so that we dont need to duplicate it in everyone calling it, but sometimes we need to override it and show a warning or have other behaviour. I think in this case we would need a safe version that returns Application | undefined and an unsafe version that throws the exception and relies on the callee to catch it. Unfortunately I don't think typescript lets you specify a function that will throw an error and you should catch it.

PR #240, opened 2022-11-30
UI Fixes and Titles

dhaselhan:
* Fix search on schedule page not scrolling / highlighting
* Fix search bar not showing warning when search fails
* Add dynamic titles per page

### DIFF
--- a/alcs-frontend/src/app/shared/meeting-overview/meeting-overview.component.ts
+++ b/alcs-frontend/src/app/shared/meeting-overview/meeting-overview.component.ts
@@ -12,7 +12,7 @@ import { UserService } from '../../services/user/user.service';
 
 type MeetingWithApplications = {
   meetingDate: number;
-  applications: (UpcomingMeetingDto & { isExpanded: boolean; isHighlighted: boolean })[];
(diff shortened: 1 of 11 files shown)


alcs-frontend/src/app/shared/meeting-overview/meeting-overview.component.ts
```diff
@@ -12,7 +12,7 @@ import { UserService } from '../../services/user/user.service';
 
 type MeetingWithApplications = {
   meetingDate: number;
-  applications: (UpcomingMeetingDto & { isExpanded: boolean; isHighlighted: boolean })[];
```
mhuseinov, line 15: just to double check, is it still expands and scrolls to meeting using the search feature

mhuseinov (approved):

alcs-frontend/src/app/shared/meeting-overview/meeting-overview.component.ts
dhaselhan: yep! It's just that applications themselves can't be expanded anymore so there is no need to keep a isExpanded boolean for them.

PR #241, opened 2022-11-30
Automapper Cleanup

dhaselhan:
* Add () => Boolean to remove need to manually map, same with string


services/apps/alcs/src/common/automapper/application-subtask.automapper.profile.ts
```diff
@@ -47,9 +45,7 @@ export class ApplicationSubtaskProfile extends AutomapperProfile {
         HomepageSubtaskDTO,
         forMember(
           (a) => a.completedAt,
-          mapFrom((ad) =>
-            ad.completedAt ? ad.completedAt.getTime() : undefined,
-          ),
+          mapFrom((ad) => ad.completedAt?.getTime()),
```
mhuseinov, line 48: Not sure, but I think the previous version would prevent deleting data from db if completedAt is sent as null, where the current will allow that.

mhuseinov (approved):

PR #242, opened 2022-12-01
Change ALCS Frontend to use Jest

dhaselhan:
* Migrate tests to use createMock
* Update expects as needed
* Add NO_ERRORS_SCHEMA to prevent warnings about child components
* With esModuleInterop: true, fix moment imports by removing * as


mhuseinov (approved):
in comparison this looks so much better

PR #245, opened 2022-12-05
Add file upload

dhaselhan:
* Set meeting scroll to center element

### DIFF
--- a/services/apps/alcs/src/document/document.service.ts
+++ b/services/apps/alcs/src/document/document.service.ts
@@ -1,5 +1,4 @@
 import {
-  DeleteObjectCommand,
   GetObjectCommand,
   PutObjectCommand,
   S3Client,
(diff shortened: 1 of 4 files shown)


services/apps/alcs/src/document/document.service.ts
```diff
@@ -63,6 +62,19 @@ export class DocumentService {
...
+  getUploadUrl(filePath: string) {
+    const fileKey = `${filePath}/${v4()}`;
+    const command = new PutObjectCommand({
+      Bucket: this.bucket,
+      Key: fileKey,
+      ACL: 'bucket-owner-full-control',
+    });
+
+    return getSignedUrl(this.dataStore, command, {
+      expiresIn: 60,
```
mhuseinov, line 74: I think expiration should go to config

mhuseinov (approved):

services/apps/alcs/src/document/document.service.ts
dhaselhan: Done!

PR #247, opened 2022-12-05
Hide Decision Released Cards

dhaselhan:
* Do not show in meetings view


services/apps/alcs/src/decision/application-decision-meeting/application-decision-meeting.service.ts
```diff
@@ -72,6 +72,7 @@ export class ApplicationDecisionMeetingService {
       .innerJoin('meeting.application', 'application')
       .innerJoin('application.reconsiderations', 'reconsideration')
       .innerJoin('reconsideration.card', 'card')
+      .where("card.status_code != 'RELE'")
```
mhuseinov, line 75: Is it worth moving 'card.status_code != 'RELE' ' into constant and use at both places?

mhuseinov (approved):

PR #249, opened 2022-12-07
Feature/acls 90

mhuseinov:
* initial setup for grpc server on alcs
* initial setup for grpc client on portal
* local certificates


services/package.json
```diff
@@ -62,9 +65,11 @@
     "config": "^3.3.7",
     "csv-parser": "^3.0.0",
     "dayjs": "^1.11.4",
+    "fs": "^0.0.1-security",
```
dhaselhan, line 68: This isn't a real package, https://www.npmjs.com/package/fs, FS is built natively into node now and I dont think this is needed.

dhaselhan (approved):
LGTM 👍

PR #251, opened 2022-12-07
Add Application Types to Portal

dhaselhan:
* Update ALCS to load into Redis
* Portal loads from redis and serves to UI
* Update UI to show types


mhuseinov (approved):
👍

PR #258, opened 2022-12-09
ALCS Frontend Code Coverage Pt 2

dhaselhan:


alcs-frontend/src/app/services/application/application-decision-meeting/application-decision-meeting.service.ts
```diff
@@ -56,12 +56,7 @@ export class ApplicationDecisionMeetingService {
   }
 
   fetchOne(uuid: string) {
-    try {
-      return firstValueFrom(this.http.get<ApplicationDecisionMeetingDto>(`${this.url}/meeting/${uuid}`));
-    } catch (err) {
-      this.toastService.showErrorToast('Failed to fetch scheduled discussion');
```
mhuseinov, line 62: Why this is removed? Is it a requirement change?

dhaselhan: It would never work, since it doesn't await the firstValueFrom the catch block will never actually fire. Both places this is used have their own try catch.

mhuseinov (approved):
👍

PR #259, opened 2022-12-09
Move animation to only hover effect for cards

dhaselhan:


mhuseinov (approved):

alcs-frontend/src/app/shared/card/card.component.scss
```diff
@@ -10,12 +10,12 @@ $date-ht: 48px;
   width: 280px;
   font-size: 16px;
   border: 2px solid transparent;
-  transition: all 100ms ease-in-out;
```
sandratoh, line 13: Can also specify the transition property instead so the border colour still gradually gets changed when hover is off. We can update in future UX ticket instead

```css
transition: border-color 100ms ease-in-out;
```

PR #260, opened 2022-12-09
Feature/alcs 90 part 2

mhuseinov:
* GRPC application service: submit application from portal to alcs
* GRPC document service: upload document from portal -> dellEcs, portal -> alcs
* Applicant info page in alcs
* Mutual TLS
* unit tests
* adjust deployment and unit tests for GRPC


portal-frontend/src/app/services/document/document.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class DocumentService {
+  private serviceUrl = `${environment.apiUrl}/document`;
+  private httpClientNoAuth: HttpClient | null = null;
+
+  constructor(private httpClient: HttpClient, handler: HttpBackend) {
+    this.httpClientNoAuth = new HttpClient(handler);
```
dhaselhan, line 15: very clean way or working around the http interceptor!

services/apps/alcs/src/application/application-code/application-local-government/application-local-government.service.ts
```diff
@@ -51,4 +51,15 @@ export class ApplicationLocalGovernmentService {
       },
     });
   }
+
+  async getByUuid(uuid: string) {
```
dhaselhan, line 55: Can you add a test for this new function?

services/apps/alcs/src/application/application-document/application-document.service.ts
```diff
@@ -100,4 +102,25 @@ export class ApplicationDocumentService {
...
     return this.documentService.getDownloadUrl(document.document);
   }
+
+  async attachExternalDocuments(
+    fileNumber: string,
+    data: ApplicationDocumentCreateDto[],
+  ) {
+    const application = await this.applicationService.getOrFail(fileNumber);
+
+    const documentsToSave: ApplicationDocument[] = [];
```
dhaselhan, line 112: nitpick: I think it would be cleaner using a .map instead of creating a new array and using an of loop to push

services/apps/alcs/src/document/document.service.ts
```diff
@@ -12,9 +12,10 @@ import { Repository } from 'typeorm';
 import { v4 } from 'uuid';
 import { CONFIG_TOKEN, IConfig } from '../common/config/config.module';
 import { User } from '../user/user.entity';
+import { CreateDocumentDto } from './document.dto';
 import { Document } from './document.entity';
 
-const DEFAULT_DB_TAGS = ['ORCS Classification: 85100-20'];
+export const DEFAULT_DB_TAGS = ['ORCS Classification: 85100-20'];
```
dhaselhan, line 18: Why did you add export here? I couldnt find it being imported anywhere.

services/apps/alcs/src/providers/typeorm/migrations/1670636483523-update_document_table.ts
```diff
@@ -0,0 +1,32 @@
...
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."document" DROP CONSTRAINT "FK_9b9254f412e8a5a07063209a08c"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."document" ALTER COLUMN "uploaded_by_uuid" DROP NOT NULL`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
```
dhaselhan, line 19: Why is the down function dropping the same constraint that up also drops?

services/apps/portal/src/alcs/document-grpc/alcs-document.message.interface.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface DocumentUploadResponseGrpc {
+  uploadUrl: string;
+  fileKey: string;
+}
+
+export class CreateDocumentRequestGrpc {
+  uploadedByUuid?: string;
+
+  mimeType: string;
```
dhaselhan, line 16: Does this also need @IsString()? Does that work here?

services/apps/portal/src/application/application-document/application-document.controller.ts
```diff
@@ -110,4 +117,30 @@ export class ApplicationDocumentController {
     await this.applicationDocumentService.delete(document);
     return {};
   }
+
+  @Post('/application/:uuid/attachExternal')
+  async attachExternalDocument(
```
dhaselhan, line 122: Needs a unit test

services/apps/portal/src/application/application.service.ts
```diff
@@ -69,6 +79,47 @@ export class ApplicationService {
...
+      );
+
+      //TODO set failed status here?
+
+      throw new BaseServiceException(
+        `Failed to submit application: ${fileNumber}`,
+      );
+    }
+
+    //TODO set submitted status here
```
dhaselhan, line 118: Are you going to do another MR after this for submitted status?

mhuseinov: nope, this is more placeholders of when we decide to add those statuses. At this point there is not clear requirement on what should happen and how it should look like after application is submitted

services/apps/alcs/src/providers/typeorm/migrations/1670636483523-update_document_table.ts
mhuseinov: That is a good typeorm question 🙂 I'll check if I can regenerate it and if not I'll drop it

services/apps/portal/src/alcs/document-grpc/alcs-document.message.interface.ts
mhuseinov: We do not need it on ALCS side since it is only used for GRPC. In portal controller dto and grpc_dto usually identical so it is possible to use one model with validator tags

services/apps/portal/src/application/application.service.ts
dhaselhan: Good point!

services/apps/portal/src/alcs/document-grpc/alcs-document.message.interface.ts
dhaselhan: Okay thanks for explaining

dhaselhan (approved):
LGTM 👍

PR #263, opened 2022-12-12
ALCS Front end tests Pt3

dhaselhan:


alcs-frontend/src/app/services/authentication/authentication.service.ts
```diff
@@ -118,12 +118,14 @@ export class AuthenticationService {
       );
       return true;
     } catch (e) {
+      console.error(e);
```
mhuseinov, line 121: I assume this one is intentional here, correct?

mhuseinov (approved):

alcs-frontend/src/app/services/authentication/authentication.service.ts
dhaselhan: It is not, deleted!

PR #264, opened 2022-12-12
Refactor Pills + Add Types to Subtasks

dhaselhan:
* Create ApplicaionTypePillComponent
* Replace existing usages with new component
* Add constants file to contain pill types
* Use parent type to display types on the subtasks components

### DIFF
--- a/services/apps/alcs/src/home/home.controller.ts
+++ b/services/apps/alcs/src/home/home.controller.ts
@@ -3,6 +3,7 @@ import { InjectMapper } from '@automapper/nestjs';
 import { Controller, Get, Param, Req, UseGuards } from '@nestjs/common';
 import { ApiOAuth2 } from '@nestjs/swagger';
 import * as config from 'config';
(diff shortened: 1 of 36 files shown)


services/apps/alcs/src/home/home.controller.ts
```diff
@@ -160,6 +161,8 @@ export class HomeController {
           completedAt: subtask.completedAt?.getTime(),
           paused: false,
           title: `${recon.application.fileNumber} (${recon.application.applicant})`,
+          appType: recon.application.type,
+          parentType: 'reconsideration',
```
mhuseinov, line 165: Is parentType typed? If yes then perfect otherwise I think it is better to have a type: 'reconsideration'|'application'...

dhaselhan: Good suggestion, added the union type as suggested.

mhuseinov (approved):
👍

PR #269, opened 2022-12-13
Re-enable Schemaspy

dhaselhan:
- Add flag to scan alcs and portal schema
- Update URLs to new Locations

### DIFF
--- a/.github/workflows/schemaspy.yml
+++ b/.github/workflows/schemaspy.yml
@@ -25,17 +25,24 @@ jobs:
       - uses: actions/setup-node@master
         with:
           node-version: 16
+      - run: |
+          sudo apt-get install --yes --no-install-recommends postgresql-client
+      - run: |
+          export PGPASSWORD='postgres';
+          psql -h 127.0.0.1 -U postgres -c 'CREATE SCHEMA IF NOT EXISTS portal;'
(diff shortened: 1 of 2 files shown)


.github/workflows/schemaspy.yml
```diff
@@ -25,17 +25,23 @@ jobs:
       - uses: actions/setup-node@master
         with:
           node-version: 16
+      - run: |
+          sudo apt-get install --yes --no-install-recommends postgresql-client
+      - run: |
+          export PGPASSWORD='postgres';
+          psql -h 127.0.0.1 -U postgres -c 'CREATE SCHEMA IF NOT EXISTS portal;'
```
mhuseinov, line 32: Why do we need to create portal schema here?

dhaselhan: So that it exists. The default postgres image has public created, and currently our migration scripts creates ALCS, but once we flatten the migration scripts to use only ALCS this will need to be updated to also create ALCS. Without this line the portal migrations will fail with something like, schema portal does not exist.

mhuseinov: I thought we already doing that in scope of postgres-schema.sql in docker folder

mhuseinov (approved):

PR #274, opened 2022-12-14
Flatten and clean up Migrations

dhaselhan:
* Fix sorting to not rely on DB order for ceo fields


services/apps/alcs/src/providers/typeorm/migrations/1570975355875-initial_migration_v2.ts
```diff
@@ -0,0 +1,549 @@
...
+      `ALTER TABLE "alcs"."reconsidered_decisions" ADD CONSTRAINT "FK_d7e09c1b2ce86005b7d0465215f" FOREIGN KEY ("application_reconsideration_uuid") REFERENCES "alcs"."application_reconsideration"("uuid") ON DELETE CASCADE ON UPDATE CASCADE`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."reconsidered_decisions" ADD CONSTRAINT "FK_09a58410db3f332490181279771" FOREIGN KEY ("application_decision_uuid") REFERENCES "alcs"."application_decision"("uuid") ON DELETE NO ACTION ON UPDATE NO ACTION`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."reconsidered_decisions" DROP CONSTRAINT "FK_09a58410db3f332490181279771"`,
```
pdittaro, line 314: Not that we would 'down' the initial migration, but is there going to be an issue with this uniquely named constraints not matching?

pdittaro (commented):
Looks good, thank you for splitting the DDL and the DML. Just had the one comment about the FK IDs on downgrade.

mhuseinov (approved):

services/apps/alcs/src/providers/typeorm/migrations/1570975355875-initial_migration_v2.ts
dhaselhan: The constraint names are generated by hashing the options of the relationship, this is how TypeORM knows if you change something on an existing relationship as the key names won't match. The good news is that theese IDs match what already exist in production as I haven't changed any relationship settings.

PR #280, opened 2022-12-15
Add Zap in Manual Dispatch Mode

dhaselhan:
* Add two actions, one for api one for backend
* Set to manual dispatch
* Prepare initial rules file


mhuseinov (approved):
👍

pdittaro (approved):

PR #284, opened 2022-12-16
Feature/alcs 91

mhuseinov:
* generate fileNumber using sequence in ALCS
* get fileNumber from ALCS in Portal during application creation
* unit tests


dhaselhan (approved):
Nice work! LGTM 👍

PR #287, opened 2022-12-20
Feature/codeclimate2

mhuseinov:
updated code climate to use rule of third for duplication and similarities

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,4 +1,4 @@
-version: "2"         # required to adjust maintainability checks
+version: "2" # required to adjust maintainability checks
 checks:
   argument-count:
     config:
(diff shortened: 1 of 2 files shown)


.codeclimate.yml
```diff
@@ -24,12 +24,17 @@ checks:
...
-      threshold: 100
+# https://docs.codeclimate.com/docs/duplication
+plugins:
+  duplication:
+    enabled: true
+    config:
+      count_threshold: 3
+      languages:
+        typescript:
+          mass_threshold: 80
```
pdittaro, line 35: Do we want to adjust this threshold up? Are all the duplications identified valid?

mhuseinov: At first glance it looks like they are valid enough. However, we may want to add specific rules to exclude some language constructions, for example, class/function initialization, filters etc. The overall amount of smells dropped from 190 to 129

pdittaro: Would be nice if we can drop it even more in a subsequent pr

pdittaro (approved):

dhaselhan (approved):
LGTM 👍

PR #292, opened 2022-12-23
Feature/alcs 338

mhuseinov:
* new styles
  * typography
  * header
  * footer
* flip secondary and accent colours
* landing page

### DIFF
--- a/portal-frontend/src/styles/functions.scss
+++ b/portal-frontend/src/styles/functions.scss
@@ -1,0 +1,5 @@
+@use 'sass:math';
+
+@function rem($pixels, $context: 16) {
+  @return math.div($pixels, $context) * 1rem;
+}
--- a/portal-frontend/src/styles/typography.scss
+++ b/portal-frontend/src/styles/typography.scss
@@ -59,9 +59,10 @@ p,
 }
 
 @mixin button-text {
-  @include typography-style(rem(14), 400);
+  @include typography-style(rem(14), 700);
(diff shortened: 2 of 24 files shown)


portal-frontend/src/styles/functions.scss
```diff
@@ -0,0 +1,5 @@
+@use 'sass:math';
+
+@function rem($pixels, $context: 16) {
```
dhaselhan, line 3: Do I want to ask why? lol

mhuseinov: All the sizing in designs is in pixels. Also, the base size was specified as 18px (which does not make sense since the browser default is 16px). Anyway, I thought this way it should be easier for us to adjust sizes in the long run and at the same time the values will match pixels.

portal-frontend/src/styles/typography.scss
```diff
@@ -1,8 +1,14 @@
 @import '@bcgov/bc-sans/css/BCSans.css';
 @import './colors';
+@import './functions.scss';
 
 $font-family: 'BCSans', 'Noto Sans', Verdana, Arial, sans-serif;
 
+$tabletBreakpoint: rem(481);
+$desktopBreakpoint: rem(769);
+$midBreakpoint: rem(669);
```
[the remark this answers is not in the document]
mhuseinov: got that as experimenting using the zoom 500%, agree that name could be more descriptive

PR #302, opened 2023-01-06
Add loading of submitted applications to Governments

dhaselhan:
* Add new flags to control which actions can be taken on the inbox page to applications
* Add loading of submitted to LG applications to attached accounts
* Add new Submitted to L/FNG Status

### DIFF
--- a/services/apps/portal/src/application/application-document/application-document.controller.ts
+++ b/services/apps/portal/src/application/application-document/application-document.controller.ts
@@ -46,8 +46,7 @@ export class ApplicationDocumentController {
     @Param('documentType') documentType: string,
     @Req() req,
   ): Promise<ApplicationDocumentDto> {
-    //Verify Access
-    await this.applicationService.getIfCreator(fileNumber, req.user.entity);
+    await this.applicationService.verifyAccess(fileNumber, req.user.entity);
 
     if (!req.isMultipart()) {
       throw new BadRequestException('Request is not multipart');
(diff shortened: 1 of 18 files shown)


services/apps/portal/src/providers/typeorm/migrations/1673042070216-update_app_statuses.ts
```diff
@@ -0,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class updateAppStatuses1673042070216 implements MigrationInterface {
+  name = 'updateAppStatuses1673042070216';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `UPDATE "portal"."application_status" SET "label" = 'Submitted to ALC' WHERE "code" = 'SUBM'`,
```
mhuseinov, line 8: I wonder if this executed successfully without ';'

services/apps/portal/src/application/application-document/application-document.controller.ts
```diff
@@ -43,6 +46,9 @@ export class ApplicationDocumentController {
     @Param('documentType') documentType: string,
     @Req() req,
   ): Promise<ApplicationDocumentDto> {
+    //Verify Access
+    await this.applicationService.getIfCreator(fileNumber, req.user.entity);
```
mhuseinov, line 50: You could create a wrapper method 'verifyAccess' for 'getIfCreator' that does not return any value and simple calls the 'getIfCreator'. This way you do not need to add a comment on every call

services/apps/portal/src/providers/typeorm/migrations/1673042070216-update_app_statuses.ts
dhaselhan: I believe so, only need ; if you are using multiple statements, if its just a single statement Postgres doesn't seem to care

services/apps/portal/src/application/application-document/application-document.controller.ts
dhaselhan: Good idea, done!

mhuseinov (approved):

PR #308, opened 2023-01-10
Add Application View Page

dhaselhan:
* Add View page accessible by LGs and Applicants
* Move edit/cancel/review buttons to new page
* Update getApplication to allow LGs to fetch submitted applications

### DIFF
--- a/portal-frontend/src/app/features/view-application/view-application.component.ts
+++ b/portal-frontend/src/app/features/view-application/view-application.component.ts
@@ -1,0 +1,55 @@
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
+import { Subject, takeUntil } from 'rxjs';
+import { ApplicationDto } from '../../services/application/application.dto';
+import { ApplicationService } from '../../services/application/application.service';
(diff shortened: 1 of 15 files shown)


portal-frontend/src/app/features/view-application/view-application.component.ts
```diff
@@ -0,0 +1,55 @@
...
+  $destroy = new Subject<void>();
+
+  constructor(
+    private applicationService: ApplicationService,
+    private route: ActivatedRoute,
+    private confirmationDialogService: ConfirmationDialogService
+  ) {}
+
+  ngOnInit(): void {
+    this.route.paramMap.pipe(takeUntil(this.$destroy)).subscribe((routeParams) => {
```
mhuseinov, line 25: I wish destroy would be a default option, so we do not need to wright it every single time

mhuseinov (approved):
💯

portal-frontend/src/app/features/view-application/view-application.component.ts
dhaselhan: 100%, not sure why there isnt an easier way

PR #311, opened 2023-01-11
fix read more

mhuseinov:
* fix read more inconsistent behaviour
* do not remove selected type value on navigating between steps
* refactor read more in create dialog


dhaselhan (approved):
LGTM 👍

PR #313, opened 2023-01-13
Feature/alcs 336 csp connect src

mhuseinov:
dynamic csp content-src for portal and alcs


pdittaro (approved):
LGTM, but we'll have to see how it behaves when deployed

PR #316, opened 2023-01-16
Add files and add submission of Application Reviews

dhaselhan:
* Move front end files to their own module

### DIFF
--- a/services/apps/portal/src/application-review/application-review.service.ts
+++ b/services/apps/portal/src/application-review/application-review.service.ts
@@ -3,6 +3,7 @@ import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
 import { LocalGovernment } from '../alcs/local-government/local-government.service';
+import { DOCUMENT_TYPES } from '../application/application-document/application-document.entity';
 import { Application } from '../application/application.entity';
 import { UpdateApplicationReviewDto } from './application-review.dto';
 import { ApplicationReview } from './application-review.entity';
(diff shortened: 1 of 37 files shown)


services/apps/portal/src/application-review/application-review.service.ts
```diff
@@ -103,4 +124,57 @@ export class ApplicationReviewService {
...
+      ) {
+        throw new BaseServiceException('Zoning information not complete');
+      }
+    }
+
+    if (applicationReview.isAuthorized === null) {
+      throw new BaseServiceException('Review authorization needs to be set');
+    }
+
+    //TODO: Where do we verify the documents are uploaded?
```
mhuseinov, line 176: I thought that verification of documents would check that at least one document of each required type is attached. However, have no idea what is required other than certificate of title

services/apps/portal/src/application/application-document/application-document.entity.ts
```diff
@@ -11,8 +11,17 @@ import {
 import { Document } from '../../document/document.entity';
 import { Application } from '../application.entity';
 
-export const DOCUMENT_TYPES = ['certificateOfTitle'] as const;
-export type DOCUMENT_TYPE = 'certificateOfTitle';
+export const DOCUMENT_TYPES = [
```
mhuseinov, line 14: probably need a ticket to move this types to code tables on ALCS

mhuseinov (approved):

services/apps/portal/src/application-review/application-review.service.ts
dhaselhan: Added it here, I was on the edge of not passing the application to the application review service, but I think its easier to have all the validation checks done in a single place.

services/apps/portal/src/application/application-document/application-document.entity.ts
dhaselhan: Probably a good idea, its not the nicest method currently

PR #320, opened 2023-01-18
Add expansion of panels that have errors

dhaselhan:
* Allow isAuthorized null if no ocp or zoning


portal-frontend/src/app/features/review-application/review-submit/review-submit.component.ts
```diff
@@ -35,7 +43,14 @@ export class ReviewSubmitComponent implements OnInit, OnDestroy {
     private applicationService: ApplicationService
   ) {}
 
+  @HostListener('window:resize', ['$event'])
+  onWindowResize() {
+    this.isMobile = window.innerWidth < 480;
```
mhuseinov, line 48: Is it also working on a zoomed window?

dhaselhan: Yes, it kicks in around 300% zoom level based on window size

mhuseinov (approved):

PR #323, opened 2023-01-18
Adjust Board Statuses

dhaselhan:
* Rename and add new CEO statuses
* Add new status for Soil/fill


services/apps/alcs/src/providers/typeorm/migrations/1674082002832-adjust_board_statuses.ts
```diff
@@ -0,0 +1,105 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
mhuseinov, line 1: And I assume you've checked that there are no cards on ceo board in prod, correct?

mhuseinov (approved):

services/apps/alcs/src/providers/typeorm/migrations/1674082002832-adjust_board_statuses.ts
dhaselhan: There are 11, there is code here to reset them to the first status on the board which is fine with Ansley

PR #331, opened 2023-01-20
Upgrade Front ends to Angular 15

dhaselhan:


alcs-frontend/src/app/features/application/post-decision/post-decision.component.html
```diff
@@ -131,7 +131,7 @@ <h5 id="modifications" class="post-decision-header">
             (save)="onSaveModificationOutcome(modification.uuid, $event)"
           ></app-inline-review-outcome>
         </span>
-        <span *ngIf="modification.reviewOutcome.code !== 'PEN'">{{ modification.reviewOutcome?.label }}</span>
+        <span *ngIf="modification.reviewOutcome.code !== 'PEN'">{{ modification.reviewOutcome.label }}</span>
```
mhuseinov, line 134: So now it properly handles undefined?

mhuseinov (approved):

alcs-frontend/src/app/features/application/post-decision/post-decision.component.html
dhaselhan: I think its more that now the linter suggests places where you used ? and dont actually need it

PR #339, opened 2023-01-25
Bugfix/alcs 395

mhuseinov:
* align dropdown text
* uppercase and font-weight for buttons
* card padding
* detailed card padding
* place the mentions list in front of the text input area
* search field styling


dhaselhan (approved):
LGTM 👍

PR #349, opened 2023-01-27
fix scroll when 500 zoom, fix radio buttons in modals

mhuseinov:
* fix scroll when 500% zoom,
* fix radio buttons in modals


dhaselhan (approved):
LGTM 👍

PR #350, opened 2023-01-27
Accept notification title as param instead of loading application

dhaselhan:
* Comments are not always on applications
* Add it in to the DTO so that the backend doesn't need to understand each entity type

### DIFF
--- a/alcs-frontend/src/app/shared/commenting/comment.component.ts
+++ b/alcs-frontend/src/app/shared/commenting/comment.component.ts
@@ -49,7 +49,6 @@ export class CommentComponent implements OnInit {
   }
 
   onSave(comment: UpdateCommentDto) {
-    debugger;
     this.isEditing = false;
     this.edit.emit({
       uuid: comment.uuid,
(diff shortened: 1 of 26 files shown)


alcs-frontend/src/app/shared/commenting/comment.component.ts
```diff
@@ -48,11 +49,13 @@ export class CommentComponent implements OnInit {
   }
 
   onSave(comment: UpdateCommentDto) {
+    debugger;
```
mhuseinov, line 52: Looks like a debug leftover

alcs-frontend/src/app/shared/commenting/comments.component.ts
```diff
@@ -61,6 +62,7 @@ export class CommentsComponent implements OnInit {
   }
 
   async onSave(comment: CreateCommentDto) {
+    debugger;
```
mhuseinov, line 65: the same with debug

mhuseinov (approved):

PR #365, opened 2023-02-02
Make owners only save on parcel when parcel saves

dhaselhan:
* Change code to no longer re-load the application but just owners
* Only link owners to parcels when pressing save

### DIFF
--- a/services/apps/portal/src/application/application.service.ts
+++ b/services/apps/portal/src/application/application.service.ts
@@ -296,7 +296,9 @@ export class ApplicationService {
           document: true,
         },
         owners: {
-          parcels: true,
+          type: true,
+          corporateSummary: true,
+            parcels: true,
         },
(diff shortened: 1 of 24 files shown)


services/apps/portal/src/application/application.service.ts
```diff
@@ -296,7 +296,9 @@ export class ApplicationService {
           document: true,
         },
         owners: {
-          parcels: true,
+          type: true,
+          corporateSummary: true,
+            parcels: true,
```
mhuseinov, line 301: something weird with spacing

mhuseinov (approved):

services/apps/portal/src/application/application.service.ts
dhaselhan: Fixed!

PR #380, opened 2023-02-10
Feature/alcs 290

mhuseinov:
This is a work in progress of step 8


portal-frontend/src/app/shared/no-data/no-data.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="no-data">
-  <div class="text">No Data</div>
+  <div class="text">{{ text || 'No Data' }}</div>
```
dhaselhan, line 2: Since the logic for showing No Data is often more complicated than just a piece of text existing, such as number of flles or other conditions, I dont think moving the text inside the component is a good idea. It will create multiple patterns of using this component instead of a single way.

portal-frontend/src/app/shared/validation-error/validation-error.component.html
```diff
@@ -0,0 +1,4 @@
+<div class="subtext error">
+  <mat-icon>warning</mat-icon>
+  {{ text }}
```
dhaselhan, line 3: Might be a good case to use children instead of an input

services/apps/portal/src/application/application-parcel/application-parcel.dto.ts
```diff
@@ -38,6 +41,9 @@ export class ApplicationParcelDto {
   @AutoMap(() => String)
   ownershipTypeCode?: string | null;
 
+  @AutoMap(() => String)
```
dhaselhan, line 44: I think this should not be String since its an object, AutoMap(() => [ApplicationParcelOwnershipTypeDto])

mhuseinov: Good catch! I'll remove it. It is not going to work anyways since I have forMember in automapper

portal-frontend/src/app/shared/no-data/no-data.component.html
mhuseinov: I think I already refactored this and not using it anywhere. I'll revert changes

dhaselhan (approved):
LGTM 👍

PR #382, opened 2023-02-10
New health endpoint for prod and portal api

pdittaro:


mhuseinov (approved):
Looks good! Do I get right that you need url to be host:port/health for some deployment/healthcheck tools

pdittaro:
> Looks good! Do I get right that you need url to be host:port/health for some deployment/healthcheck tools

Not necessary for the infrastructure, this is just exposing the healthcheck publicly for the production APIs - so that our uptime monitor can rech them.

PR #383, opened 2023-02-10
lookup improvements

mhuseinov:
* set max date for purchasedDate
* lookup improvements:
   * search if dashes provided in search field
   * add leading 0 if needed for pid
   * based on lookup result populate pid
   * reset only lookup related fields


dhaselhan (approved):
LGTM 👍

PR #403, opened 2023-02-21
Add file error message + bug fixes

dhaselhan:
* Do not show agent owners after editing in all owners dialog
* Prevent duplicate saving, was happening on both button click and onDestroy
* Properly set ownerUuid so component doesn't keep creating new owners


portal-frontend/src/app/features/edit-application/primary-contact/primary-contact.component.ts
```diff
@@ -83,7 +83,7 @@ export class PrimaryContactComponent implements OnInit, OnDestroy {
   }
 
   async onSaveExit() {
-    await this.save();
+    //await this.save();
```
mhuseinov, line 86: Do we need this?

mhuseinov (approved):

PR #405, opened 2023-02-22
Add L/FNG Page to ALCS

dhaselhan:
* Add column to store application review
* Add to application dto
* Add page to load / show the data

### DIFF
--- a/services/apps/portal/src/application/application-owner/application-owner.service.ts
+++ b/services/apps/portal/src/application/application-owner/application-owner.service.ts
@@ -4,6 +4,7 @@ import { InjectRepository } from '@nestjs/typeorm';
 import { Any, Repository } from 'typeorm';
 import { DocumentService } from '../../document/document.service';
 import { User } from '../../user/user.entity';
+import { PARCEL_TYPE } from '../application-parcel/application-parcel.dto';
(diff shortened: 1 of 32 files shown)


services/apps/portal/src/application/application-owner/application-owner.service.ts
```diff
@@ -217,9 +217,9 @@ export class ApplicationOwnerService {
     const parcels =
       await this.applicationParcelService.fetchByApplicationFileId(fileId);
 
-    const firstParcel = parcels.reduce((a, b) =>
-      a.auditCreatedAt > b.auditCreatedAt ? a : b,
-    );
+    const firstParcel = parcels
+      .filter((parcel) => parcel.parcelType === 'application')
```
mhuseinov, line 221: there is a PARCEL_TYPE in application-parcel.dto.ts

mhuseinov (approved):

services/apps/portal/src/application/application-owner/application-owner.service.ts
dhaselhan: It was only on the front end dto. update MR to add to backend DTO and use here.

PR #406, opened 2023-02-22
Feature/alcs 419

mhuseinov:
* rework other parcels
   * list all owners from step 1
   * CRUD 'OTHER' parcels
   * disabled state for owner and parcel components
   * unit tests
* remove save from next and back buttons click on stepper
* fix load application issues on edit application


dhaselhan (approved):
LGTM 👍

PR #407, opened 2023-02-22
Increase timeout and add log statement for debug

dhaselhan:
* Add log statement to confirm the issue
* Increase timeout as a suspect


portal-frontend/src/app/features/edit-application/select-government/select-government.component.ts
```diff
@@ -63,16 +63,18 @@ export class SelectGovernmentComponent implements OnInit, OnDestroy {
   }
 
   onBlur() {
-    //Blur will fire below optionSelected
+    //Blur will fire before onChange above, so use setTimeout to delay it
     setTimeout(() => {
+      debugger;
```
mhuseinov, line 68: Is this also a part of investigate?

mhuseinov (approved):

PR #415, opened 2023-02-24
Feature/alcs 447

mhuseinov:
* custom beforeStepChange event 
* fix Parcel and Other Parcels header
* can deactivate guard that allows custom component implementation
* fix owners in other parcels


dhaselhan (approved):
LGTM 👍

PR #417, opened 2023-02-27
Feature/alcs 447

mhuseinov:
* url navigation between steps
* navigate and expand parcels from Step 8
* validation parameter that allows validation based on which step user navigated from
* small loader improvement


dhaselhan (approved):
LGTM 👍

PR #420, opened 2023-02-28
Bugfix/alcs 524

mhuseinov:
* fix owners table scroll
* fix parcel lookup reset
* fix other parcel responsive layout


dhaselhan (approved):
LGTM 👍

PR #421, opened 2023-02-28
fix local government create dialog

mhuseinov:
* add preferred region dropdown
* fix issue with required bceid in dto


dhaselhan (approved):
LGTM 👍

PR #423, opened 2023-03-01
OATS Application ETL pt. 1

pdittaro:


bin/migrate-oats-data/applications.py
```diff
@@ -0,0 +1,47 @@
...
+    final_count = cursor.execute(
+        "select count(*) from alcs.application a where a.audit_created_by = 'oats_etl'"
+    )
+    print("- Actual inserted: ", final_count)
+    cursor.close()
+
+
+@inject_conn_pool
+def clean_applications(conn=None):
+    cursor = conn.cursor()
```
mhuseinov, line 42: Should this also drop temporary tables?

mhuseinov: nevermind, did not notice that it was specified with TEMPORARY

pdittaro: Temporary tables should be dropped when the connection closes, from my understanding

mhuseinov (approved):

PR #424, opened 2023-03-01
Add Application backend validation + UI Improvements

dhaselhan:
* UI Improvements to Frontend
* Add new validator service and unit tests
* Add invalid format errors to front end

### DIFF
--- a/services/apps/portal/src/alcs/code/code.controller.spec.ts
+++ b/services/apps/portal/src/alcs/code/code.controller.spec.ts
@@ -44,7 +44,6 @@ describe('CodeController', () => {
         uuid: 'fake-uuid',
         name: 'fake-name',
         isFirstNation: false,
-        isActive: true,
       },
     ]);
     mockAppTypeService.list.mockResolvedValue([]);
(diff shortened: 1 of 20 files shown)


services/apps/portal/src/alcs/code/code.controller.spec.ts
```diff
@@ -44,7 +44,6 @@ describe('CodeController', () => {
         uuid: 'fake-uuid',
         name: 'fake-name',
         isFirstNation: false,
-        isActive: true,
```
mhuseinov, line 47: Is it not relevant anymore?

services/apps/portal/src/application-review/application-review.controller.ts
```diff
@@ -156,9 +158,21 @@ export class ApplicationReviewController {
       userLocalGovernment.isFirstNation,
     );
 
+    const validationResult =
+      await this.applicationValidatorService.validateApplication(application);
+
+    if (!validationResult.application) {
+      throw new BaseServiceException(
+        `Invalid application found during LG Submission ${application.fileNumber}`,
```
mhuseinov, line 166: Do I get it right that later this message will be dynamic based on the application type?

mhuseinov (approved):

services/apps/portal/src/alcs/code/code.controller.spec.ts
dhaselhan: This was a mistake by me, it was never actually sent over from ALCS, so it was always undefined.

services/apps/portal/src/application-review/application-review.controller.ts
dhaselhan: This is more a safeguard, this error should never happen as the application was validated when the Applicant submitted it. So unless something went wrong then the application should still be valid after the LG review is added.

PR #432, opened 2023-03-03
Update PID/PIN Search

dhaselhan:
* Split into PID only or PIN only search
* Update UI to match figma
* Enable  / disable based on selections and options


mhuseinov (commented):
Is there a new validation requirement for the pid pin on step 8?

mhuseinov (approved):

dhaselhan:
There is another ticket to track changes to Step 8 for Crown Lands

PR #439, opened 2023-03-04
Make pin optional and update Step 8 for Crown Land

dhaselhan:
* also reset search by when switching from Crown Land -> Simple

### DIFF
--- a/portal-frontend/src/app/features/edit-application/parcel-details/parcel-entry/parcel-entry.component.ts
+++ b/portal-frontend/src/app/features/edit-application/parcel-details/parcel-entry/parcel-entry.component.ts
@@ -63,6 +63,7 @@ export class ParcelEntryComponent implements OnInit {
 
   searchBy = new FormControl<string | null>(null);
   isCrownLand = false;
+  isCertificateOfTitleRequired = true;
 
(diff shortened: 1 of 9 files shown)


portal-frontend/src/app/features/edit-application/parcel-details/parcel-entry/parcel-entry.component.ts
```diff
@@ -312,6 +312,13 @@ export class ParcelEntryComponent implements OnInit {
         });
       }
 
+      const pidValue = this.pid.getRawValue();
+      if (this.isCrownLand) {
```
mhuseinov, line 316: do we need to reflect this on API level too?

mhuseinov (approved):

portal-frontend/src/app/features/edit-application/parcel-details/parcel-entry/parcel-entry.component.ts
dhaselhan: MR was just updated 30 seconds after you commented to include API validation and tests

mhuseinov (approved):

PR #442, opened 2023-03-07
Add archive/unarchive cards

dhaselhan:
* Add ability to archive / unarchive cards


alcs-frontend/src/app/features/board/dialogs/planning-review/planning-review-dialog.component.ts
```diff
@@ -67,6 +70,13 @@ export class PlanningReviewDialogComponent implements OnInit, OnDestroy {
       }
     });
 
+    this.authService.$currentUser.pipe(takeUntil(this.$destroy)).subscribe((currentUser) => {
```
mhuseinov, line 73: I almost want to move this check to some kind CardDialogBaseComponent

services/apps/alcs/src/admin/unarchive-card/unarchive-card.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  ) {}
+
+  async fetchByFileId(fileId: string) {
+    const result: {
+      cardUuid: string;
+      type: string;
+      status: string;
+      createdAt: number;
+    }[] = [];
+    const application = await this.applicationService.getDeletedCard(fileId);
```
mhuseinov, line 25: This looks like a good candidate to split into multiple functions. It should be easier later to keep track of all the different card types that get attached to the result list

mhuseinov (approved):

alcs-frontend/src/app/features/board/dialogs/planning-review/planning-review-dialog.component.ts
dhaselhan: Done!

services/apps/alcs/src/admin/unarchive-card/unarchive-card.service.ts
dhaselhan: Done!

mhuseinov (approved):
🔥

PR #446, opened 2023-03-08
Bugfix/alcs 557

mhuseinov:
* fix step buttons not working on step 7
* do not show confirmation on step 2 if no other parcels


dhaselhan (approved):
LGTM 👍

PR #450, opened 2023-03-09
Bugfix/alcs 553

mhuseinov:
* allow users form the same LG to update primary contact
* update owner endpoints to support LG flow


dhaselhan (approved):
LGTM 👍

PR #451, opened 2023-03-09
Submit Refused to Forward Application to ALCS

dhaselhan:
* If app is not authorized disable menu items and show tooltip

### DIFF
--- a/services/apps/portal/src/application/application.service.ts
+++ b/services/apps/portal/src/application/application.service.ts
@@ -238,8 +238,6 @@ export class ApplicationService {
           },
         }),
       );
-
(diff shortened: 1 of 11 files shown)


mhuseinov (approved):

services/apps/portal/src/application/application.service.ts
```diff
@@ -238,8 +238,6 @@ export class ApplicationService {
           },
         }),
       );
-
-      await this.updateStatus(application, APPLICATION_STATUS.SUBMITTED_TO_ALC);
```
mhuseinov, line 242: should this method be renamed then since it is not submitting to ALCS

PR #455, opened 2023-03-13
Add TUR Fields and Logic to Portal

dhaselhan:
* Control components based on app type
* Update validation, proposal step, and view Step

### DIFF
--- a/portal-frontend/src/app/features/application-details/application-details.component.html
+++ b/portal-frontend/src/app/features/application-details/application-details.component.html
@@ -170,7 +170,18 @@
 </section>
 <section>
   <h3>6. Proposal</h3>
-  <app-nfu-details [showErrors]="showErrors" [showEdit]="showEdit" [application]="application"></app-nfu-details>
+  <app-nfu-details
+    *ngIf="application && application.typeCode === 'NFUP'"
+    [showErrors]="showErrors"
+    [showEdit]="showEdit"
+    [application]="application"
+  ></app-nfu-details>
+  <app-tur-details
+    *ngIf="application && application.typeCode === 'TURP'"
+    [showErrors]="showErrors"
+    [showEdit]="showEdit"
+    [application]="application"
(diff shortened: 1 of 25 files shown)


portal-frontend/src/app/features/application-details/application-details.component.html
```diff
@@ -170,7 +170,18 @@ <h4>Land Use of Adjacent Parcels</h4>
 </section>
 <section>
   <h3>6. Proposal</h3>
-  <app-nfu-details [showErrors]="showErrors" [showEdit]="showEdit" [application]="application"></app-nfu-details>
+  <app-nfu-details
+    *ngIf="application && application.typeCode === 'NFUP'"
```
mhuseinov, line 174: not in the scope of this MR, but it would be nice to have an enum of application types considering we will have 5+ different types

portal-frontend/src/app/features/edit-application/edit-application.component.html
```diff
@@ -69,12 +69,19 @@ <h6 *ngIf="application">Application ID: {{ application.fileNumber }} | {{ applic
   </cdk-step>
   <cdk-step label="Proposal">
     <ng-container *ngTemplateOutlet="header"></ng-container>
-    <div *ngIf="cdkStepper.selectedIndex === editAppSteps.Proposal">
+    <div *ngIf="application && cdkStepper.selectedIndex === editAppSteps.Proposal">
```
mhuseinov, line 72: Is an application check required only because of async data retrieval or is it possible not to have an app at all?

mhuseinov (approved):

portal-frontend/src/app/features/edit-application/edit-application.component.html
dhaselhan: Just due to async data retrieval, I figured it was cleaner for the components to require a defined application and to use ngIfs higher up.

PR #458, opened 2023-03-14
Feature/alcs 530

mhuseinov:
* email template generation using mjml and handlebars
* example of template using the actual design from figma


services/libs/common/src/email-template-service/alcs-logo.base64.ts
dhaselhan, line 1: Is this use anymore?

dhaselhan (approved):
LGTM 👍

services/libs/common/src/email-template-service/alcs-logo.base64.ts
mhuseinov: I do not think we even use the common.module anywhere. I'll remove EmailTemplateService for now

PR #467, opened 2023-03-15
Feature/alcs 582

mhuseinov:
* rename portal.application -> portal.application_proposal
* rename portal.application_review -> portal.application_proposal_review
* rename all realated API controllers, services, dtos, entities including file names

NOTE: UI part will be in separate MR


dhaselhan (approved):
LGTM 👍

PR #485, opened 2023-03-22
Portal Fixes

dhaselhan:
Fix file uploading / reloading on several pages

### DIFF
--- a/services/apps/alcs/src/portal/application-document/application-document.controller.ts
+++ b/services/apps/alcs/src/portal/application-document/application-document.controller.ts
@@ -135,12 +135,11 @@ export class ApplicationDocumentController {
   async delete(@Param('uuid') fileUuid: string, @Req() req) {
     const document = await this.applicationDocumentService.get(fileUuid);
 
-    await this.applicationSubmissionService.verifyAccess(
-      document.applicationUuid,
-      req.user.entity,
-    );
-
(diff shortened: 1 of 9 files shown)


services/apps/alcs/src/portal/application-document/application-document.controller.ts
```diff
@@ -135,12 +135,11 @@ export class ApplicationDocumentController {
   async delete(@Param('uuid') fileUuid: string, @Req() req) {
     const document = await this.applicationDocumentService.get(fileUuid);
 
-    await this.applicationSubmissionService.verifyAccess(
-      document.applicationUuid,
-      req.user.entity,
-    );
-
-    //TODO: How do we control who can delete which document types?
+    //TODO: How do we know which documents applicant can delete?
```
mhuseinov, line 138: no verifyAccess anymore? Is it because LUP should be able to manage these?

mhuseinov (approved):

services/apps/alcs/src/portal/application-document/application-document.controller.ts
dhaselhan: We can't call verify access as we don't have the file number, just the documents uuid. Might need to think about how we handle this and add fileNumber to the endpoints just so we have it.

PR #488, opened 2023-03-23
Feature/remove grpc app submit

mhuseinov:
* use application submission instead of jsonb column
* remap ui to use correct values for applicant info
* fix documents for application submission on alcs side (temp solution for parcel till we get finalized document version)
* unit tests


alcs-frontend/src/app/services/application/application.dto.ts
```diff
@@ -66,11 +77,13 @@ export interface SubmittedApplicationParcelDto {
   parcelType: string;
   documentUuids: string[];
   owners: SubmittedApplicationOwnerDto[];
+  documents: ApplicationParcelDocumentDto[];
 }
 
 export interface SubmittedApplicationDto {
```
dhaselhan, line 83: Could probably rename to ApplicationSubmissionDto now

dhaselhan (approved):
LGTM 👍

alcs-frontend/src/app/services/application/application.dto.ts
mhuseinov: done

PR #493, opened 2023-03-24
Feature/alcs 628

mhuseinov:
* review ReviewJson and use the actual entity instead
* relink review to application
* unit tests
* cleanups


dhaselhan (approved):
Great cleanup! LGTM 👍

PR #495, opened 2023-03-24
Backport Hotfix

dhaselhan:


.github/workflows/build.yml
```diff
@@ -30,24 +30,9 @@ jobs:
       dockerfile: Dockerfile.migrate
       build-args: NEST_APP=alcs
     secrets: inherit
-  build-portal-api:
```
mhuseinov, line 33: this was not in develop initially?

mhuseinov (approved):

.github/workflows/build.yml
dhaselhan: I guess not, not sure what happened

PR #498, opened 2023-03-27
Add new intake page fields

dhaselhan:
* Add inline boolean and number inputs
* Add new fields to app entity for new data

### DIFF
--- a/services/apps/alcs/src/alcs/application/application.dto.ts
+++ b/services/apps/alcs/src/alcs/application/application.dto.ts
@@ -263,39 +263,45 @@ export class UpdateApplicationDto {
   @IsNumber()
   dateSubmittedToAlc?: number;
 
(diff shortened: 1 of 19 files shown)


services/apps/alcs/src/alcs/application/application.dto.ts
```diff
@@ -338,7 +342,17 @@ export class ApplicationDto {
...
+
+  feePaidDate?: number;
+
+  @AutoMap(() => Boolean)
+  feeWaived?: boolean;
+
+  @AutoMap(() => Boolean)
+  feeSplitWithLg?: boolean;
+
+  @AutoMap(() => String)
```
mhuseinov, line 354: just a nitpick I do not think you need a String in properties.

mhuseinov (approved):
👍

PR #499, opened 2023-03-28
Alcs/playwright demo

mhuseinov:
* TUR application submission creation in portal
* Partial NFU submission creation in portal


portal-frontend/test/tests-examples/demo-todo-app.spec.ts
dhaselhan, line 1: Do we need this file?

portal-frontend/test/tests/submissions/TUR_submission_creation.spec.ts
```diff
@@ -0,0 +1,174 @@
+import { test } from '@playwright/test';
+
+const baseUrl = 'http://localhost:4201/login';
```
dhaselhan, line 3: Since these are copied between files, could we use a central config file?

portal-frontend/test/tests-examples/demo-todo-app.spec.ts
mhuseinov: I left it as an example of an actual test, but happy to remove it if it is not needed

portal-frontend/test/tests/submissions/TUR_submission_creation.spec.ts
dhaselhan: I wasn't thinking of config library, could you just make a config.ts in the same folder that these import from?

portal-frontend/test/tests-examples/demo-todo-app.spec.ts
dhaselhan: I think you can remove it

mhuseinov: done

portal-frontend/test/tests/submissions/TUR_submission_creation.spec.ts
mhuseinov: done

dhaselhan (approved):
LGTM 👍

PR #506, opened 2023-03-31
Update Evidentiary Record to use VIsibility Flags

dhaselhan:
Evidentiary record now shows records with C visibility flag


services/apps/alcs/src/alcs/application/application-document/application-document.entity.ts
```diff
@@ -12,6 +12,13 @@ import { Document } from '../../../document/document.entity';
 import { Application } from '../application.entity';
 import { ApplicationDocumentCode } from './application-document-code.entity';
 
+export enum VISIBILITY_FLAG {
```
mhuseinov, line 15: 👍

PR #518, opened 2023-04-05
Feature/alcs 375

mhuseinov:
* submission document generation
   * nfu
   * tur


portal-frontend/src/app/services/application-submission/application-submisison-document-generation/application-submission-document-generation.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+interface Report {
+  fileName: string;
+  data: Blob;
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ApplicationSubmissionDocumentGenerationService {
```
dhaselhan, line 18: Do you think its better to have a service for each type or maybe a general DocumentGenerationService that has generateSubmissionPdf, generateReviewPdf, etc?

dhaselhan (approved):
LGTM 👍

portal-frontend/src/app/services/application-submission/application-submisison-document-generation/application-submission-document-generation.service.ts
mhuseinov: I prefer service per document type since that way it will be easier to maintain them. For example, we will have +-8 submission types and potentially each one of them with a custom payload.

PR #528, opened 2023-04-11
Load Application Review to enable/disable side menu items

dhaselhan:


services/apps/alcs/src/alcs/application/application-submission-review/application-submission-review.service.ts
```diff
@@ -15,7 +15,7 @@ export class ApplicationSubmissionReviewService {
   ) {}
 
   async get(fileNumber: string) {
-    return await this.applicationSubmissionReviewRepository.findOneOrFail({
+    return await this.applicationSubmissionReviewRepository.findOne({
```
mhuseinov, line 18: Are you sure that there is no code that is relying on getOneOrFail to throw an error?

mhuseinov (approved):

services/apps/alcs/src/alcs/application/application-submission-review/application-submission-review.service.ts
dhaselhan: Yes, this service is ALCS only and the only usage of get is the controller.

PR #556, opened 2023-04-20
adjust fonts, set no data for Proof of Homesite Severance Qualification

mhuseinov:
* adjust fonts to BC Sans on Subd
* show no data in Proof of Homesite Severance Qualification if no files


dhaselhan (approved):
LGTM 👍

PR #562, opened 2023-04-21
Publish Submissions with Auditing

dhaselhan:
* Add Publish button to draft submit page
* Delete old submission and set draft to not draft
* Generate new PDF and clear visibility flags on old one
* Rename service/module to better match front end
* Add auditing of submissions via Postgres triggers


services/apps/alcs/src/portal/application-submission-draft/application-submission-draft.service.ts
```diff
@@ -159,6 +159,50 @@ export class ApplicationEditService {
...
+      );
+    }
+
+    //Delete Original Submission
+    for (const owner of original.owners) {
+      await this.applicationOwnerService.delete(owner);
+    }
+    const parcelUuids = original.parcels.map((parcel) => parcel.uuid);
+    await this.applicationParcelService.deleteMany(parcelUuids);
+    await this.applicationSubmissionRepository.remove(original);
```
mhuseinov, line 193: I assume that the log table persists the removed submission well?

mhuseinov (approved):
👍

services/apps/alcs/src/portal/application-submission-draft/application-submission-draft.service.ts
dhaselhan: yes! It will track the delete statement with the deleted row :)

PR #563, opened 2023-04-21
Update create new app modal UX and UI

sandratoh:
- Fix modal width
- Fix modal height to app type modal
  - Did not fix modal height in submission type modal because modal has less fluctuation compared to app type modal, and the increased gap felt more confusing instead
- Route successful application creation to form directly
- Apply responsive styling to modal (tested with 400% zoom and mobile)
- Use angular material `mat-dialog-action` attribute to fix action button to the bottom of the modal

### DIFF
--- a/portal-frontend/src/app/features/create-application-dialog/create-application-dialog.component.scss
+++ b/portal-frontend/src/app/features/create-application-dialog/create-application-dialog.component.scss
@@ -52,6 +52,22 @@
   display: none;
 }
 
+.step-description {
+  margin-bottom: 0.5rem;
+}
+
+.dialog-container {
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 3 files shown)


portal-frontend/src/app/features/create-application-dialog/create-application-dialog.component.scss
```diff
@@ -52,6 +52,22 @@
   display: none;
 }
 
+.step-description {
+  margin-bottom: 0.5rem;
```
dhaselhan, line 56: For css sizes in the Portal we are using the rem() function instead of rems directly, so this for example should be rem(8).

dhaselhan (approved):
LGTM 👍

PR #566, opened 2023-04-25
Feature/alcs 495 -part-1

mhuseinov:
- separated api decisions into v1 and v2
- separate ui decisions into v1 and v2 
- added isDraft flag
- isDraft on creation


dhaselhan (approved):
LGTM :

PR #585, opened 2023-05-02
Bug Fixes

dhaselhan:
* Allow Proposal Summary to be Nullable
* Only load active governments


alcs-frontend/src/app/features/application/overview/overview.component.ts
```diff
@@ -281,8 +281,9 @@ export class OverviewComponent implements OnInit, OnDestroy {
 
   onSaveSummary(updatedSummary: string) {
     if (this.application) {
+      debugger;
```
mhuseinov, line 284: Look like a debug leftover

dhaselhan: removed!

PR #586, opened 2023-05-02
Removal of Soil Step 8 and Viewing in ALCS

dhaselhan:


alcs-frontend/src/app/services/application/application.dto.ts
```diff
@@ -140,6 +140,25 @@ export interface ApplicationSubmissionDto {
   subdAgricultureSupport?: string;
   subdIsHomeSiteSeverance?: boolean;
   subdProposedLots: ProposedLot[];
+
+  //Soil Fields
+  soilIsNOIFollowUp?: boolean | null;
```
mhuseinov, line 145: I wonder if it worth starting to split interface into smaller ones and use inheritance instead of comment separation?

portal-frontend/src/app/features/edit-submission/proposal/roso-proposal/roso-proposal.component.ts
```diff
@@ -186,7 +197,7 @@ export class RosoProposalComponent extends StepComponent implements OnInit, OnDe
   onChangeNOI(selectedValue: string) {
     if (selectedValue === 'true') {
```
mhuseinov, line 198: Not sure if it will look better, but you could use parseStringToBoolean and work with selectedValue as boolean type

mhuseinov (approved):

portal-frontend/src/app/features/edit-submission/proposal/roso-proposal/roso-proposal.component.ts
dhaselhan: Good suggestion, done!

alcs-frontend/src/app/services/application/application.dto.ts
dhaselhan: It might be, but since we have 3-4 copies of this DTO I will leave it to its own discussion / ticket.

PR #590, opened 2023-05-03
Improve disabled fields UI

sandratoh:
- Bugfix on disabled dropdown placeholder text to match the rest of input fields
- Improve disabled fields to make them more disabled: cursor behaviour and background color


dhaselhan (approved):
LGTM 👍

PR #592, opened 2023-05-04
Display minimum file size to user

sandratoh:
Use native JS Math functions to calculate and compare file size instead of number to string conversion for improved performance


dhaselhan (approved):
LGTM 👍

PR #611, opened 2023-05-16
Portal interaction and scroll fixes

sandratoh:
- Fix toast overlay from blocking other elements
- Add scroll behaviour to Create New + Change App Type modals
- Changes to parcel owner dropdown if owner is added already
- Collapse panel if no other parcels in application

### DIFF
--- a/portal-frontend/src/app/features/create-application-dialog/create-application-dialog.component.ts
+++ b/portal-frontend/src/app/features/create-application-dialog/create-application-dialog.component.ts
@@ -5,6 +5,7 @@ import { Router } from '@angular/router';
 import { ApplicationSubmissionService } from '../../services/application-submission/application-submission.service';
 import { ApplicationTypeDto, SubmissionTypeDto } from '../../services/code/code.dto';
 import { CodeService } from '../../services/code/code.service';
+import { scrollToElement } from '../../shared/utils/scroll-helper';
 
 export enum ApplicationCreateDialogStepsEnum {
   submissionType = 0,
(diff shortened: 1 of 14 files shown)


portal-frontend/src/app/features/create-application-dialog/create-application-dialog.component.ts
```diff
@@ -79,11 +79,35 @@ export class CreateApplicationDialogComponent implements OnInit, AfterViewChecke
 
   onReadMoreClicked() {
     this.readMoreClicked = !this.readMoreClicked;
```
dhaselhan, line 81: These 4 all look the same to me, can you DRY it up?

portal-frontend/src/app/features/edit-submission/other-parcels/other-parcels.component.ts
```diff
@@ -254,7 +254,6 @@ export class OtherParcelsComponent extends StepComponent implements OnInit, OnDe
           } else {
             this.hasOtherParcelsInCommunity.patchValue('true');
             this.formDisabled = false;
-            await this.setHasOtherParcelsInCommunity(true);
```
dhaselhan, line 257: Can you explain why we don't want to do this?

sandratoh: This was resetting and clearing the form when user wanted to cancel removing parcel.

portal-frontend/src/app/features/create-application-dialog/create-application-dialog.component.ts
sandratoh: Updated!

dhaselhan (approved):
LGTM 👍

PR #617, opened 2023-05-17
Send notifications for all card types

dhaselhan:


services/apps/alcs/src/alcs/notification/notification.service.ts
```diff
@@ -45,9 +45,7 @@ export class NotificationService {
     });
   }
 
-  async createNotificationForApplication(
-    notificationDto: CreateNotificationServiceDto,
-  ) {
+  async saveNotification(notificationDto: CreateNotificationServiceDto) {
```
mhuseinov, line 48: Just a nitpick of nitpicks, but I would prefer something like add/create notification rather than save

mhuseinov (approved):
🎯

services/apps/alcs/src/alcs/notification/notification.service.ts
dhaselhan: Done!

PR #619, opened 2023-05-18
Properly save return comment for applications

dhaselhan:


services/apps/alcs/src/portal/application-submission-review/application-submission-review.controller.spec.ts
```diff
@@ -418,13 +418,17 @@ describe('ApplicationSubmissionReviewController', () => {
...
         statusCode: APPLICATION_STATUS.IN_REVIEW,
       }),
     );
     mockAppSubmissionService.updateStatus.mockResolvedValue({} as any);
     mockAppReviewService.getByFileNumber.mockResolvedValue(applicationReview);
     mockAppReviewService.delete.mockResolvedValue();
     mockAppDocService.delete.mockResolvedValue({} as any);
+    mockAppSubmissionService.update.mockResolvedValue(
+      new ApplicationSubmission(),
+    );
```
sandratoh, line 431: Code looks good! Just wanted to learn more about what's happening here

sandratoh (approved):
LGTM!

services/apps/alcs/src/portal/application-submission-review/application-submission-review.controller.spec.ts
dhaselhan: I added a new update service call in the code, since I need to call update to save the returned comment. So in this case I need to provide a mock for the service's update method. Update returns an instance of the entity being updated, and since I don't care about the fields on the mocked entity, I create an empty one.

PR #621, opened 2023-05-18
Added OATS and OATS_P flag to migrated OATS data in ALCS system

lstod:
Documents created by PROXY_OATS_APPLICANT and PROXY_OATS_LOCGOV are considered to have come from the portal and given the system flag of 'OATS_P'. All other documents are assumed to have come from the oats system and given the flag of 'OATS'.


dhaselhan (approved):
LGTM 👍

PR #622, opened 2023-05-19
Update Change App Type modal to match Create New modal

sandratoh:
In addition to UI changes, this also includes an UX update which disables user from updating app type to the same one as the current submission's app type


dhaselhan (approved):
LGTM 👍

PR #624, opened 2023-05-19
Add Revert to Draft and Re-release

dhaselhan:
* Add DB flag to track if a decision was previously released
* Add new dialog for reverting to draft
* Change release dialog if a decision was previously released


sandratoh (approved):
LGTM 👍

PR #638, opened 2023-05-26
Feature/alcs 559

lstod:
Partial MR for ALCS 559, batched import of applications

### DIFF
--- a/bin/migrate-oats-data/batch_applications.py
+++ b/bin/migrate-oats-data/batch_applications.py
@@ -85,7 +85,7 @@ def process_applications(conn=None, batch_size=10000):
                 cursor.execute(
                     f"{application_sql} WHERE ae.application_id > {last_application_id} ORDER by ae.application_id"
                 )
-                print("batch size ", batch_size)
+                
                 rows = cursor.fetchmany(batch_size)
                 if not rows:
                     break
@@ -118,11 +118,14 @@ def process_applications(conn=None, batch_size=10000):
 
 @inject_conn_pool
 def clean_applications(conn=None):
+    print("Start applications cleaning")
     with conn.cursor() as cursor:
         cursor.execute(
             "DELETE FROM alcs.application a WHERE a.audit_created_by = 'oats_etl'"
         )
+        print(f"Deleted application items count = {cursor.rowcount}")
         cursor.execute("DELETE FROM alcs.card c WHERE c.audit_created_by = 'oats_etl'")
+        print(f"Deleted cards count = {cursor.rowcount}")
 
 
         
--- a/bin/migrate-oats-data/batch_applications.py
+++ b/bin/migrate-oats-data/batch_applications.py
@@ -146,8 +146,8 @@ def process_applications(conn=None, batch_size=10000):
                     insert_query = compile_application_insert_query(
                         applications_to_be_inserted_count
                     )
-                    print(insert_query)
-                    print(rows) 
+                    # print(insert_query)
+                    # print(rows) 
                     cursor.execute(insert_query, rows)
                     conn.commit()
  
--- a/bin/migrate-oats-data/batch_applications.py
+++ b/bin/migrate-oats-data/batch_applications.py
@@ -32,19 +32,7 @@ def application_etl_insert():
             oats.oats_alr_applications AS oa
             LEFT JOIN alcs.application AS a ON oa.alr_application_id :: text = a.file_number
     """
-# def associated_card():
-#     return f"""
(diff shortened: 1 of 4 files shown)


bin/migrate-oats-data/batch_applications.py
```diff
@@ -0,0 +1,143 @@
...
+            DISTINCT oa.alr_application_id AS application_id,
+            CASE
+                WHEN a.uuid IS NOT NULL THEN TRUE
+                ELSE false
+            END AS duplicated
+        FROM
+            oats.oats_alr_applications AS oa
+            LEFT JOIN alcs.application AS a ON oa.alr_application_id :: text = a.file_number
+    """
+# def associated_card():
```
mhuseinov, line 35: Do you need this commented code?

bin/migrate-oats-data/batch_applications.py
```diff
@@ -0,0 +1,143 @@
...
+                                      applicant, region_code, local_government_uuid, audit_created_by)
+
+        VALUES{applications_to_insert}
+        ON CONFLICT (file_number) DO UPDATE SET
+            file_number = EXCLUDED.file_number,
+            
+            type_code = EXCLUDED.type_code,
+            applicant = EXCLUDED.applicant,
+            region_code = EXCLUDED.region_code,
+            local_government_uuid = EXCLUDED.local_government_uuid,
```
mhuseinov, line 66: Have you checked if all local governments from OATS(prod) exist in ALCS (all environments)? If not could you check that? The insert will fail on the FK constraint in ALCS if Local Government does not exist in ALCS.

bin/migrate-oats-data/batch_applications.py
```diff
@@ -0,0 +1,143 @@
...
+    function takes the number of rows to insert and generates an SQL insert statement with upserts using the ON CONFLICT clause
+    """
+
+    applications_to_insert = ",".join(["%s"] * number_of_rows_to_insert)
+    return f"""
+        INSERT INTO alcs.application (file_number, type_code, 
+                                      applicant, region_code, local_government_uuid, audit_created_by)
+
+        VALUES{applications_to_insert}
+        ON CONFLICT (file_number) DO UPDATE SET
```
mhuseinov, line 60: If you have not confirmed the "on conflict" condition with Ansley, let's do that. This condition needs to be approved by the business. Otherwise, we may be overriding the application with outdated info

bin/migrate-oats-data/batch_applications.py
```diff
@@ -0,0 +1,143 @@
...
+                    )
+
+                    print(
+                        f"retrieved/inserted items count: {applications_to_be_inserted_count}; total successfully inserted/updated applications so far {successful_inserts_count}; last inserted applidation_id: {last_application_id}"
+
+                    )
+                except Exception as e:
+                    conn.rollback()
+                    print("Error", e)
+                    failed_inserts += len(rows)
```
mhuseinov, line 123: We discussed that the failed_inserts would show the wrong value at this point since the +=len(rows) is not equal to the number of rows after the shift in last_application_id +1. Let's replace it with the count_total-successful_inserts_count after the last batch insert.

bin/migrate-oats-data/migrate.py
```diff
@@ -105,6 +121,9 @@ def setup_menu_args_parser(document_import_batch_size):
                     if args and args.batch_size:
                         document_import_batch_size = args.batch_size
 
+                    console.log("Batching applications:")
+                    process_applications(batch_size=application_import_batch_size)
```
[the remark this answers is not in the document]
lstod: updated to single variable

bin/migrate-oats-data/batch_applications.py
lstod: logic updated

bin/migrate-oats-data/batch_applications.py
lstod: Discussed loc gov to be separate ticket

bin/migrate-oats-data/batch_applications.py
```diff
@@ -0,0 +1,196 @@
...
+                rows = cursor.fetchmany(batch_size)
+                if not rows:
+                    break
+                try: 
+                    applications_to_be_inserted_count = len(rows)
+
+                    insert_query = compile_application_insert_query(
+                        applications_to_be_inserted_count
+                    )
+                    # print(insert_query)
```
mhuseinov, line 149: lets clean this up

mhuseinov (approved):

PR #639, opened 2023-05-29
Add NOI Detail Page

dhaselhan:
* Add Overview and Intake
* Add more fields to NOIs
* Change Staff Journal to work for all object types


alcs-frontend/src/app/features/notice-of-intent/overview/overview.component.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  constructor(private noticeOfIntentDetailService: NoticeOfIntentDetailService) {}
+
+  ngOnInit(): void {
+    this.noticeOfIntentDetailService.$noticeOfIntent
+      .pipe(takeUntil(this.$destroy))
+      .pipe(
+        tap((noi) => {
+          if (noi) {
+            //Load other stuff here
```
mhuseinov, line 42: I guess this is a todo for the next ticket?

alcs-frontend/src/app/shared/details-header/details-header.component.ts
```diff
@@ -1,27 +1,35 @@
...
 })
-export class ApplicationHeaderComponent {
+export class DetailsHeaderComponent {
   destroy = new Subject<void>();
 
   linkedCards: (CardDto & { displayName: string })[] = [];
 
-  _application: ApplicationDto | CommissionerApplicationDto | undefined;
-  @Input() set application(application: ApplicationDto | CommissionerApplicationDto | undefined) {
+  @Input() heading = 'Title Here';
```
mhuseinov, line 23: Is it also a TODO?

services/apps/alcs/src/providers/typeorm/migrations/1685382185264-unlink_journal.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."application_staff_journal" DROP CONSTRAINT "FK_a074329b9f848c30b4165a79889"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."application_staff_journal" RENAME TO "staff_journal";`,
+    );
+
+    await queryRunner.query(
```
mhuseinov, line 14: The only concern that I have is probably the fact that there is no FK on db for staff journal and no way to easily tell based on the record that it belongs to application or NOI

mhuseinov (approved):

alcs-frontend/src/app/shared/details-header/details-header.component.ts
dhaselhan: Nope, this is an obvious placeholder so that you don't forget to set the input.

alcs-frontend/src/app/features/notice-of-intent/overview/overview.component.ts
dhaselhan: Yes, this will be needed eventually for loading other things that will go in the timeline

services/apps/alcs/src/providers/typeorm/migrations/1685382185264-unlink_journal.ts
dhaselhan: Yeah I was thinking the same thing, should I add a staff journal column to Application and NOI to keep an FK? The good news is that we never actually delete NOIs or Applications.

mhuseinov (approved):
🔥

PR #656, opened 2023-06-02
Fix disabled upload area styling in non-local env

sandratoh:
Found out why styling wasn't applied when deployed:
* `<div>` doesn't support `disabled` attribute
* Ng-Reflect attributes only exist as dev helpers, and they won't show up in non-debug env (prod)


dhaselhan (approved):
LGTM 👍

PR #672, opened 2023-06-08
Delete app pause directly when report start and end date are null

dhaselhan:
* Setting to null was not actually deleting the pause entity


services/apps/alcs/src/alcs/application/application-meeting/application-meeting.service.spec.ts
```diff
@@ -161,4 +168,24 @@ describe('ApplicationMeetingService', () => {
       ),
     );
   });
+
+  it('should delete the meeting pause if one exists and there is no stard and end date', async () => {
```
sandratoh, line 172: nitpick: typo on start

dhaselhan: Fixed!

mhuseinov (approved):

PR #674, opened 2023-06-09
Log user out of Siteminder at same time as Keycloak

dhaselhan:
* Add Log Out button to Provision Page

### DIFF
--- a/services/config/default.json
+++ b/services/config/default.json
@@ -29,6 +29,9 @@
     "SCOPES": ["openid"],
     "REALM": "standard"
   },
+  "SITEMINDER": {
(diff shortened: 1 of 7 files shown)


services/config/default.json
```diff
@@ -29,6 +29,9 @@
     "SCOPES": ["openid"],
     "REALM": "standard"
   },
+  "SITEMINDER": {
+    "LOGOUT_URL": "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl="
```
mhuseinov, line 33: do not forget to add this to openshift 🙂

mhuseinov (approved):

PR #675, opened 2023-06-09
Feature/alcs 559

lstod:
Implemented case when for type codes.

Added local gov base functionality, still needs tweaking to handle mismatches.


bin/migrate-oats-data/sql/insert-batch-application.sql
```diff
@@ -0,0 +1,116 @@
...
+        WHEN atl.code = 'NAR' THEN 'NARU'
+        ELSE 'NARU'
+    END AS type_code,
+    CASE
+        WHEN applicant_lookup.orgs IS NOT NULL THEN applicant_lookup.orgs
+        WHEN applicant_lookup.persons IS NOT NULL THEN applicant_lookup.persons
+        ELSE 'Unknown'
+    END AS applicant,
+	ar.code as region_code,
+    --TODO: local government lookup
```
mhuseinov, line 99: do we need this?

lstod: updated to more descriptive comment

bin/migrate-oats-data/sql/insert-batch-application.sql
```diff
@@ -0,0 +1,115 @@
...
+    CASE
+        WHEN atl.code = 'TUR' THEN 'TURP'
+        WHEN atl.code = 'INC' THEN 'INCL'
+        WHEN atl.code = 'EXC' THEN 'EXCL'
+        WHEN atl.code = 'SDV' THEN 'SUBD'
+        WHEN atl.code = 'NFU' THEN 'NFUP'
+        WHEN atl.code = 'SCH' THEN 'PFRS'
+        WHEN atl.code = 'EXT' THEN 'ROSO'
+        WHEN atl.code = 'FILL' THEN 'POFO'
+        -- WHEN atl.code = 'SRW' THEN 'NARU'
```
mhuseinov, line 87: Could you remind me what was the decision with SRW and CSC. Do we need to import those?

bin/migrate-oats-data/sql/insert-batch-application.sql
```diff
@@ -38,44 +62,58 @@ panel_lookup AS (
...
-    -- TODO: panel region lookup
-    -- 'INTR',
 	ar.code as region_code,
-    --Peace river TODO: local government lookup
-    '001cfdad-bc6e-4d25-9294-1550603da980',
+    --IN PROGRESS: local government lookup name mismatches
+    --Currently using peace river as default uuid
+    CASE
+        WHEN alcs_gov.gov_uuid IS NOT NULL THEN alcs_gov.gov_uuid
+        ELSE '001cfdad-bc6e-4d25-9294-1550603da980'
```
mhuseinov, line 106: Is it because LG is a required field on application?

bin/migrate-oats-data/sql/insert-batch-application.sql
mhuseinov: Agreed that this mapping will be done in a separate ticket since we do not have equivalents in ALCS at this point. @lstod will create a ticket

bin/migrate-oats-data/sql/insert-batch-application.sql
lstod: It is required to be a uuid, you had requested that it was not left to be NULL

bin/migrate-oats-data/batch_applications.py
```diff
@@ -188,6 +132,15 @@ def clean_applications(conn=None):
         )
         print(f"Deleted items count = {cursor.rowcount}")
 
+        cursor.execute(
```
mhuseinov, line 135: I do not think you need to drop these tables after the import. They could work as a good reference to why some of the applications were excluded.

mhuseinov (approved):

bin/migrate-oats-data/sql/insert-batch-application.sql
mhuseinov: I meant is it a nullable field in DB? If there is no way to determine what LG it belongs to and the field is not nullable then you need to set that value. Otherwise, it should be null.

PR #681, opened 2023-06-12
Add Soil Proposal Component

dhaselhan:
* Shared among all three soil types
* Add ha to ALR Area label


services/apps/alcs/src/alcs/application/application.entity.ts
```diff
@@ -235,7 +235,7 @@ export class Application extends Base {
     comment: 'The date at which the non-farm use ends',
```
mhuseinov, line 235: I guess the comments also needs to be updated

alcs-frontend/src/app/features/application/proposal/soil/soil.component.html
```diff
@@ -0,0 +1,49 @@
...
+        <td *ngIf="submission.soilToRemoveArea !== null">{{ submission.soilToRemoveArea }} ha</td>
+      </tr>
+      <tr>
+        <td class="soil-table-label">Maximum Depth</td>
+        <td *ngIf="submission.soilToPlaceMaximumDepth !== null">{{ submission.soilToPlaceMaximumDepth }} m</td>
+        <td *ngIf="submission.soilToRemoveMaximumDepth !== null">{{ submission.soilToRemoveMaximumDepth }} m</td>
+      </tr>
+      <tr>
+        <td class="soil-table-label">Average Depth</td>
+        <td *ngIf="submission.soilToPlaceAverageDepth !== null">{{ submission.soilToPlaceAverageDepth }} m</td>
```
mhuseinov, line 44: why only !==null, I thinks it is possible to have undefined at least based on dto?

mhuseinov (approved):

alcs-frontend/src/app/features/application/proposal/soil/soil.component.html
dhaselhan: I think they will always be null as the database defaults to null, looks like the DTO is quite wrong since it has ? for a lot of fields.

dhaselhan: Updated the DTO to have null.

PR #683, opened 2023-06-13
Bugfix/alcs 634

mhuseinov:
- once an application is cancelled redirect to portal inbox
- retrieve application data using the roles/owner safe method


dhaselhan (approved):
LGTM 👍

PR #685, opened 2023-06-13
Add NOI Import Service

dhaselhan:
* Import NOIs!

### DIFF
--- a/services/apps/alcs/src/alcs/import/noi-import.service.ts
+++ b/services/apps/alcs/src/alcs/import/noi-import.service.ts
@@ -1,0 +1,380 @@
+import { Injectable, Logger } from '@nestjs/common';
+import * as csv from 'csv-parser';
+import * as dayjs from 'dayjs';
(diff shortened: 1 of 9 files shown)


services/apps/alcs/src/alcs/import/noi-import.service.ts
```diff
@@ -0,0 +1,380 @@
...
+  }
+
+  private mapSubTypes(mappedRow: NOIRow) {
+    const subtypes: string[] = [];
+
+    ['subtype1', 'subtype2'].forEach((fieldKey) => {
+      let subtype = mappedRow[fieldKey];
+
+      //Quick data fix for single type with typo
+      if (subtype === 'Farm Building -  Intensive Livestock') {
```
mhuseinov, line 214: 👍

mhuseinov (approved):
Do we need to update readme to include instructions for import jobs?

mhuseinov:
it is nice to have. So this code will make sense to whoever is going to work on it later.

PR #689, opened 2023-06-13
NOI Modifications Bug Fixes

dhaselhan:
* Show Modification outcome and Received all items on timeline
* Remove Create Covenant and Create Recon from NOI Board
* Show noi modifications cards in modification header
* Hook up NOI_MODI Card type to changes


alcs-frontend/src/app/features/notice-of-intent/notice-of-intent.component.ts
```diff
@@ -90,6 +92,10 @@ export class NoticeOfIntentComponent implements OnInit, OnDestroy {
         this.noticeOfIntent = noticeOfIntent;
       }
     });
+
+    this.noticeOfIntentModificationService.$modifications.pipe(takeUntil(this.destroy)).subscribe((value) => {
+      this.modifications = value;
```
mhuseinov, line 97: Do you need to clear this.modifications in OnDestroy?

dhaselhan: Ohh good catch, definitely should

PR #702, opened 2023-06-16
Feature/alcs 738

mhuseinov:
- Decision components:
   - POFO
   - ROSO
   - PFRS
- TUR updates
- API validation for decision components

### DIFF
--- a/services/apps/alcs/src/alcs/application-decision/application-decision-v2/application-decision/application-decision-v2.service.spec.ts
+++ b/services/apps/alcs/src/alcs/application-decision/application-decision-v2/application-decision/application-decision-v2.service.spec.ts
@@ -566,34 +566,4 @@ describe('ApplicationDecisionV2Service', () => {
       );
     });
   });
-
-  // it('should throw an exception on release decision if no components', async () => {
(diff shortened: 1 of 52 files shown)


services/apps/alcs/src/alcs/application-decision/application-decision-v2/application-decision/application-decision-v2.service.spec.ts
```diff
@@ -553,4 +566,34 @@ describe('ApplicationDecisionV2Service', () => {
       );
     });
   });
+
+  // it('should throw an exception on release decision if no components', async () => {
```
dhaselhan, line 570: Commented out test?

dhaselhan (approved):
LGTM 👍

services/apps/alcs/src/alcs/application-decision/application-decision-v2/application-decision/application-decision-v2.service.spec.ts
mhuseinov: thanks! it is not relevant to this service and I forgot to delete it, I'll do it now

PR #703, opened 2023-06-19
Feature/alcs 868

lstod:
Local Gov lookup, currently handling exact matches & Islands Trust


bin/migrate-oats-data/sql/insert-batch-application.sql
```diff
@@ -45,11 +45,19 @@ alcs_gov as(
...
     FROM oats_gov
-        JOIN alcs.application_local_government alg ON oats_gov.oats_gov_name = alg."name"
+        join alcs.application_local_government alg on
+   (case
+   	when oats_gov.oats_gov_name LIKE 'Islands Trust%' then 'Islands Trust'
+   	else oats_gov.oats_gov_name
+   end) 
+   =
+   (case
+   	when alg."name" LIKE 'Islands Trust%' then 'Islands Trust'
```
mhuseinov, line 55: I thought that alcs.application_local_government has only 'Islands Trust' and no variations. If that is the case then the right 'case when' is redundant

lstod: It definitely is

mhuseinov (approved):
🎯

PR #704, opened 2023-06-19
Fix commissioner placeholder to correct label

sandratoh:


alcs-frontend/src/app/shared/details-header/details-header.component.ts
```diff
@@ -25,7 +25,7 @@ export class DetailsHeaderComponent {
 
   linkedCards: (CardDto & { displayName: string })[] = [];
 
-  @Input() heading = 'Title Here';
+  @Input() heading = 'Application';
```
dhaselhan, line 28: This was intentional set here as an obvious placeholder. A more correct fix would be to set this input property in the HTML file where it's used.

dhaselhan (approved):
LGMT 👍

PR #720, opened 2023-06-22
Bug Fixes

dhaselhan:
* Use long date format for staff journal dates
* Hide the whole Resulting Outcome Div so there is no gap


sandratoh (approved):
LGTM 👍

PR #743, opened 2023-06-28
Show primary contact field validations in review if selected contact

sandratoh:
Step 3 is unchanged
Screenshot 2023-06-28 at 1 56 04 PM

Step 8 is updated to show the same fields as app detail read only page
Screenshot 2023-06-28 at 1 56 20 PM


dhaselhan (approved):
LGTM 👍

PR #746, opened 2023-06-28
Add Decision Condition Types to Admin

dhaselhan:


services/apps/alcs/src/alcs/admin/application-decision-condition-types/application-decision-condition-types.controller.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    controller = module.get<ApplicationDecisionConditionTypesController>(
+      ApplicationDecisionConditionTypesController,
+    );
+  });
+
+  it('should be defined', () => {
+    expect(controller).toBeDefined();
+  });
+
+  it('should call out to service when fetching ceo criterion', async () => {
```
mhuseinov, line 42: looks like a typo, should be 'condition type'. This is in all tests for this file

mhuseinov (approved):

services/apps/alcs/src/alcs/admin/application-decision-condition-types/application-decision-condition-types.controller.spec.ts
dhaselhan: Fixed

PR #748, opened 2023-06-29
Fix empty agents creation/deletion and auth letter validation

sandratoh:
- Delete empty agents before creating new one
- Update backend unit test
- Fix step 3 authorization letter validation bug if user switches from 3rd party agent to owner post step 8.

### DIFF
--- a/portal-frontend/src/app/features/edit-submission/primary-contact/primary-contact.component.ts
+++ b/portal-frontend/src/app/features/edit-submission/primary-contact/primary-contact.component.ts
@@ -135,7 +135,11 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
         this.isCrownOwner = selectedOwner.type.code === APPLICATION_OWNER.CROWN;
       }
     }
-    this.needsAuthorizationLetter = !selectedOwner;
+    this.needsAuthorizationLetter = !(
+      this.owners[0].type.code === APPLICATION_OWNER.INDIVIDUAL &&
+      (this.owners.length === 1 ||
+        (this.owners.length === 2 && this.owners[1].type.code === APPLICATION_OWNER.AGENT && !hasSelectedAgent))
+    );
     this.files = this.files.map((file) => ({
       ...file,
       errorMessage: this.needsAuthorizationLetter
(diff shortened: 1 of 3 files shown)


portal-frontend/src/app/features/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -135,10 +135,7 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
         this.isCrownOwner = selectedOwner.type.code === APPLICATION_OWNER.CROWN;
       }
     }
-
-    this.needsAuthorizationLetter = !(
```
dhaselhan, line 139: What does this change do?

dhaselhan: I think this might not be enough, as now you are showing an error if they don't have a selected Owner, but what if they have a selected owner and its individual?

sandratoh: Oh it's definitely not, sorry I'm still fixing it right now, I couldn't find the draft mode

sandratoh: I noticed it created different bugs after committing the fix. I'll post on chat once it's ready for review!

portal-frontend/src/app/features/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -135,7 +135,11 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
         this.isCrownOwner = selectedOwner.type.code === APPLICATION_OWNER.CROWN;
       }
     }
-    this.needsAuthorizationLetter = !selectedOwner;
+    this.needsAuthorizationLetter = !(
+      this.owners[0].type.code === APPLICATION_OWNER.INDIVIDUAL &&
+      (this.owners.length === 1 ||
+        (this.owners.length === 2 && this.owners[1].type.code === APPLICATION_OWNER.AGENT && !hasSelectedAgent))
```
dhaselhan, line 141: What does this second check cover for? I feel like the original rule of owner.length == 1 and owner is individual is the ONLY case where we don't require an auth letter.

sandratoh:
The second check is for when there are two owners (one individual and one
agent), and the individual owner is selected as the primary contact. It is
possible to have two owners when the applicant originally select a third
party agent as the primary contact, saves progress, but then goes back to
this step and change the primary contact to individual owner instead.
Before they save progress, the old agent will not be deleted, and Auth
Letter becomes required on this step when it shouldn't. You can see this
error more prominently with all validations turned on.

dhaselhan (approved):
LGTM 👍

PR #750, opened 2023-06-30
Add Parcel Prep Table to App Prep

dhaselhan:
* Add editable ALR Area
* Add new column for it
* Add new component with table for display

### DIFF
--- a/alcs-frontend/src/app/features/application/applicant-info/application-details/parcel/parcel.component.ts
+++ b/alcs-frontend/src/app/features/application/applicant-info/application-details/parcel/parcel.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core';
+import { AfterContentChecked, Component, Input, OnChanges, OnDestroy, OnInit, SimpleChanges } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { Subject, takeUntil } from 'rxjs';
 import { ApplicationDocumentDto } from '../../../../../services/application/application-document/application-document.dto';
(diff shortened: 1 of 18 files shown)


alcs-frontend/src/app/features/application/applicant-info/application-details/parcel/parcel.component.ts
```diff
@@ -24,14 +28,30 @@ export class ParcelComponent implements OnInit, OnChanges {
...
+    private route: ActivatedRoute
   ) {}
 
   ngOnInit(): void {
     if (this.parcelType === 'other') {
       this.pageTitle = 'Other Parcels in the Community';
       this.showCertificateOfTitle = false;
     }
+
+    this.route.fragment.pipe(takeUntil(this.$destroy)).subscribe((fragment) => {
```
mhuseinov, line 41: Maybe you could try retrieving the id from this.route.snapshot.paramMap.get... and using afterIniti instead of onInit with setTimeout 200?

alcs-frontend/src/app/services/application/application-parcel/application-parcel.service.ts
```diff
@@ -21,4 +21,20 @@ export class ApplicationParcelService {
...
   }
+
+  async setParcelArea(uuid: string, alrArea: number | null): Promise<ApplicationParcelDto[]> {
+    try {
+      const res = await firstValueFrom(
+        this.http.post<ApplicationParcelDto[]>(`${this.baseUrl}/${uuid}`, {
+          alrArea,
+        })
+      );
+      this.toastService.showSuccessToast('Application updated');
```
mhuseinov, line 32: nitpick, "Application parcel updated" so will match the error toast?

mhuseinov (approved):

alcs-frontend/src/app/services/application/application-parcel/application-parcel.service.ts
dhaselhan: yeah this was a bit odd, but the requirements specifically say to say Application updated when the ALR area is saved.

alcs-frontend/src/app/features/application/applicant-info/application-details/parcel/parcel.component.ts
dhaselhan: Changed! Its no longer using a set timeout and executing in ngAfterContentChecked.

PR #751, opened 2023-06-30
fixing errors in migrations

lstod:
Migration to fix issues with previous local gov migration


services/apps/alcs/src/providers/typeorm/migrations/1688144882418-local_gov_corrections.ts
```diff
@@ -0,0 +1,25 @@
...
+        UPDATE "alcs"."application_local_government" SET "name" = 'Islands Trust Capital (Historical)' WHERE "name" = 'Islands Trust-Capital (Historical)';
+        UPDATE "alcs"."application_local_government" SET "name" = 'Islands Trust Powell River (Historical)' WHERE "name" = 'Islands Trust-Powell River (Historical)';
+        UPDATE "alcs"."application_local_government" SET "name" = 'Islands Trust Sunshine Coast (Historical)' WHERE "name" = 'Islands Trust-Sunshine Coast (Historical)';
+        UPDATE "alcs"."application_local_government" SET "name" = 'Islands Trust Cowichan Valley (Historical)' WHERE "name" = 'Islands Trust - Cowichan Valley (Historical)';
+        UPDATE "alcs"."application_local_government" SET "name" = 'Northern Rockies (Historical)' WHERE "uuid" = '55f665bc-c91b-4bbb-85c2-39691088b297';
+        UPDATE "alcs"."application_local_government" SET "preferred_region_code" = 'NORR' WHERE "uuid" = '55f665bc-c91b-4bbb-85c2-39691088b297';
+        DELETE FROM alcs.application_local_government WHERE uuid = '0136a261-6237-4332-ae6a-1c7d7f49f0e8';
```
mhuseinov, line 15: will this cause issues like previously if something already using the LG?

lstod: it shouldn't but It would be good practice to do the more foolproof way

mhuseinov (approved):

PR #762, opened 2023-07-10
ALCS View of under/pending LFNG Review and returned applications

sandratoh:
- Retrieve app status using `ApplicationSubmissionService`
- Display L/FNG banner based on status
- Add app status enum to ALCS (same enum exist in `portal-frontend` and `services` directories)

### DIFF
--- a/alcs-frontend/src/app/features/application/application.component.ts
+++ b/alcs-frontend/src/app/features/application/application.component.ts
@@ -44,7 +44,7 @@ export const unsubmittedRoutes = [
   },
 ];
 
-export const awaitingLfngRoutes = [
+export const submittedLfngRoutes = [
   {
     path: '',
     menuTitle: 'Overview',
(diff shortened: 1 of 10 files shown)


alcs-frontend/src/app/features/application/application.component.ts
```diff
@@ -128,17 +165,21 @@ export class ApplicationComponent implements OnInit, OnDestroy {
...
   fileNumber?: string;
   application: ApplicationDto | undefined;
   reconsiderations: ApplicationReconsiderationDto[] = [];
   modifications: ApplicationModificationDto[] = [];
+  submission?: ApplicationSubmissionDto;
 
   isApplicantSubmission = false;
-  isSubmitted = false;
+  isSubmittedAlc = false;
+  isSubmittedLfng = false;
```
dhaselhan, line 178: I recommend renaming this to isPendingLfngReview or something similar, as currently it would imply its true if it was submitted to the LFNG but its actually true when its in a status where the lfng needs to do something.

dhaselhan (approved):
LGTM 👍

alcs-frontend/src/app/features/application/application.component.ts
sandratoh: I initially named it something like that, but then this view is also true when Lfng returns the application and the applicant needs to do something. Any other recommendation for a name?

PR #764, opened 2023-07-10
Bugfix/alcs 502

mhuseinov:
- updated colours for condition labels
- new logic for incomplete
- fix condition to component link on reset


dhaselhan (approved):
LGTM 👍

PR #767, opened 2023-07-11
Update decision components field layout

sandratoh:


alcs-frontend/src/app/features/application/decision/decision-v2/decision-v2.component.html
```diff
@@ -204,6 +204,14 @@ <h5>Decision Components and Conditions</h5>
                     <div class="subheading2">Ag Cap Consultant</div>
                     {{ component.agCapConsultant }}
                   </div>
+                  <div *ngIf="['POFO', 'ROSO', 'PFRS'].includes(component.applicationDecisionComponentTypeCode)">
```
mhuseinov, line 207: Use End Date and Expiry Date were specifically used in custom components to avoid ngif in generic components

PR #768, opened 2023-07-11
Add Boards to Admin Pt 1

dhaselhan:
* Load Create Card Types from Backend
* Add Boards to Admin

### DIFF
--- a/alcs-frontend/src/app/features/admin/board-management/board-management.component.ts
+++ b/alcs-frontend/src/app/features/admin/board-management/board-management.component.ts
@@ -69,19 +69,6 @@ export class BoardManagementComponent implements OnInit, OnDestroy {
     });
   }
 
-  async onDelete(boardDto: BoardDto) {
-    this.confirmationDialogService
-      .openDialog({
-        body: `Are you sure you want to delete ${boardDto.title}?`,
-      })
-      .subscribe(async (answer) => {
-        if (answer) {
-          //TODO: Delete
-          await this.boardService.reloadBoards();
-        }
-      });
-  }
-
   ngOnDestroy(): void {
     this.$destroy.next();
     this.$destroy.complete();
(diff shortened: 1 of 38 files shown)


alcs-frontend/src/app/features/admin/board-management/board-management.component.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  async onDelete(boardDto: BoardDto) {
+    this.confirmationDialogService
+      .openDialog({
+        body: `Are you sure you want to delete ${boardDto.title}?`,
+      })
+      .subscribe(async (answer) => {
+        if (answer) {
+          //TODO: Delete
```
mhuseinov, line 79: I guess this is will be in part 2?

alcs-frontend/src/app/services/board/board.service.ts
```diff
@@ -27,9 +27,13 @@ export class BoardService {
     });
   }
 
-  private async publishBoards() {
+  async reloadBoards() {
+    await this.publishBoards(true);
```
mhuseinov, line 31: Do you need to clean boards on reload?

services/apps/alcs/src/alcs/board/board.service.ts
```diff
@@ -87,4 +93,61 @@ export class BoardService {
...
+    board.allowedCardTypes = updateDto.allowedCardTypes.map(
+      (code) => cardTypes.find((cardType) => cardType.code === code)!,
+    );
+    board.createCardTypes = updateDto.createCardTypes.map(
+      (code) => cardTypes.find((cardType) => cardType.code === code)!,
+    );
+
+    await this.boardRepository.save(board);
+
+    await this.boardStatusRepository.delete({
```
mhuseinov, line 136: a nitpick, it would be better to move this block into separate function, something like "setBoardStatuses". It will make it easier to read/maintain the code

mhuseinov (approved):

alcs-frontend/src/app/features/admin/board-management/board-management.component.ts
dhaselhan: ah good catch, function deleted since its all done in the dialog.

alcs-frontend/src/app/services/board/board.service.ts
dhaselhan: added!

PR #772, opened 2023-07-12
Update decision component field layouts

sandratoh:
- Move Ag Cap Map and Consultant fields to separate component
- Add to all app type components and style accordingly


dhaselhan (approved):
LGTM 👍

PR #780, opened 2023-07-14
Feature/main alcs 910

lstod:
Adding OATS to prod SchemaSpy


ora2pg/data/ddl_18_2_prod.sql
```diff
@@ -0,0 +1,4162 @@
+-- Generated by Ora2Pg, the Oracle database Schema converter, version 18.2
```
dhaselhan, line 1: Did you mean to add this file? Seems unrelated to Schemaspy.

.github/workflows/schemaspy.yml
```diff
@@ -34,14 +34,18 @@ jobs:
           npm install
           npm run alcs:migration:run
         working-directory: services
+      - name: Import OATS
```
dhaselhan, line 37: Looks like this will run every time we generate schemaspy, is that necessary?

ora2pg/data/ddl_18_2_prod.sql
dhaselhan, line 1: Why is this called ddl_18_2_prod? Should this have a more relatable name like oats_schemadump_ddl.sql or something?

.github/workflows/schemaspy.yml
lstod: removed!

dhaselhan (approved):
LGTM 👍

PR #784, opened 2023-07-17
Bugfix/alcs 908 lg in progress app

mhuseinov:
- fix issue with LG not able to open in progress app that they created


dhaselhan (approved):
LGTM 👍

PR #785, opened 2023-07-17
Feature/alcs 908 part 2

mhuseinov:
- moved statuses styling to DB
- updated portal statuses
- color schema for alcs statuses


portal-frontend/src/app/features/home/application-list/application-list.component.ts
```diff
@@ -24,18 +21,6 @@ export class ApplicationListComponent implements OnInit {
     'actions',
   ];
 
-  statusToCssMap = new Map<string, string>([
```
dhaselhan, line 27: 🔥 Burn it down! Good clean up

PR #791, opened 2023-07-18
Fix portal form re-renders other attachment to and from review

sandratoh:


mhuseinov (approved):

portal-frontend/src/app/features/application-details/application-details.component.ts
```diff
@@ -49,9 +49,9 @@ export class ApplicationDetailsComponent implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit(): void {
-    this.loadGovernments();
     this.$application.pipe(takeUntil(this.$destroy)).subscribe((app) => {
       this.applicationSubmission = app;
+      this.loadGovernments();
```
mhuseinov, line 54: this will reload LG every time application is reloaded, is it expected?

portal-frontend/src/app/features/edit-submission/other-attachments/other-attachments.component.ts
```diff
@@ -52,9 +52,8 @@ export class OtherAttachmentsComponent extends FilesStepComponent implements OnI
       }
     });
 
-    this.loadDocumentCodes();
-
     this.$applicationDocuments.pipe(takeUntil(this.$destroy)).subscribe((documents) => {
+      this.loadDocumentCodes();
```
mhuseinov, line 56: the same with codes, since they are static, do you need to reload them every time you reload documents?

portal-frontend/src/app/features/application-details/application-details.component.ts
sandratoh: You're right let me rethink this approach. Thanks Mekhti!

PR #796, opened 2023-07-18
Don't re-populate conditions on component changes

dhaselhan:
* Only need to update validity, not the actual conditions


sandratoh (approved):
👍

PR #809, opened 2023-07-21
Feature/alcs 880 decision v2 errors

mhuseinov:
- errors on decision v2
- highlight buttons
- scroll to error
- update SUBM status colour


dhaselhan (approved):
LGTM 👍

PR #814, opened 2023-07-24
Reduce Network Calls / Loads for Edit Submission

dhaselhan:
* Only save/load the application when components are dirty
* Remove logic to auto-save on every step navigation
* Add logging to help debug stepper issues

### DIFF
--- a/portal-frontend/src/app/features/review-submission/review-submission.component.ts
+++ b/portal-frontend/src/app/features/review-submission/review-submission.component.ts
@@ -97,7 +97,7 @@ export class ReviewSubmissionComponent implements OnInit, OnDestroy {
               // setTimeout is required for stepper to be initialized
               setTimeout(() => {
                 const stepInt = parseInt(stepInd);
-                this.customStepper.navigateToStep(stepInt, true);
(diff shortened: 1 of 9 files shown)


mhuseinov (approved):

portal-frontend/src/app/features/review-submission/review-submission.component.ts
```diff
@@ -97,7 +97,7 @@ export class ReviewSubmissionComponent implements OnInit, OnDestroy {
               // setTimeout is required for stepper to be initialized
               setTimeout(() => {
                 const stepInt = parseInt(stepInd);
-                this.customStepper.navigateToStep(stepInt, true);
```
mhuseinov, line 100: This will ignore if the step is provided in URL. Is it for testing purposes?

dhaselhan: Interestingly enough this only seems to break navigating to a specific parcel. Everything else seems fine.

dhaselhan: I found it odd that the review page had its own call to change steps but looked like it was the only place.

PR #819, opened 2023-07-25
Add Exclusion Step 8 Review & View section

sandratoh:
- Add exclusion proposal review section in step 8 with validations and update banner
- Fix backend bug that prevented empty proposal from being saved


dhaselhan (approved):
LGTM 👍

PR #824, opened 2023-07-26
Add inclusion fields when user is LFNG

dhaselhan:
* Add question to indicate parcel ownership and store in DB
* Add more file uploads if the answer is no


sandratoh (approved):
LGTM 👍

PR #826, opened 2023-07-27
nfu subtype with inline ng select

mhuseinov:
- inline editor using ng-select in App Prep page for nfu subtype


dhaselhan (approved):
LGTM 👍

PR #829, opened 2023-07-27
Add Exclusion applications in ALCS and fix LFNG info

sandratoh:
- Add exclusion detail in Applicant Info
- Add exclusion and inclusion proposal in ALCS edit submission
- Fix outdated L/FNG Info on ALCS to match Portal then fix some more bugs on both
- Create inline-applicant-type selector following other inline components
- Backend changes to add inclExclApplicantType to applications
- Migration to update application column with new applicant type field


alcs-frontend/src/app/features/application/lfng-info/lfng-info.component.html
```diff
@@ -108,35 +140,75 @@ <h4>Plans & Bylaws: Zoning</h4>
...
+      >
+        By indicating that the parcel(s) is not subject to Local Government OCP or Zoning,
+        <strong>no authorizing resolution is required</strong> as per S. 25 (3) or S. 29 (4) of the <em>ALC Act</em>.
+        The only option available is to forward this application on to the ALC.
+      </div>
+
+      <div
+        class="full-row"
+        *ngIf="
+          isFirstNationGovernment ||
```
dhaselhan, line 169: This is getting pretty complex, can we move this to the component and use a cleaner boolean here?

sandratoh: Updated!

dhaselhan (approved):
LGTM 👍

PR #830, opened 2023-07-27
Load the full type and set it instead of just code

dhaselhan:
* Setting just the code will not trump the actual value so saving a changed code doe not work


sandratoh (approved):
LGTM 👍

PR #836, opened 2023-07-31
Various Fixes

dhaselhan:
* load creating government to check rather than users as the logged in user may change
* add FIXME to Exclusion Link
* Fix back button going to prescribed body for non-exclusion apps
* Fix INCL going to Prescribed body and not just EXCL


mhuseinov (approved):

portal-frontend/src/app/features/edit-submission/proposal/incl-proposal/incl-proposal.component.ts
```diff
@@ -132,12 +139,14 @@ export class InclProposalComponent extends FilesStepComponent implements OnInit,
         inclGovernmentOwnsAllParcels: parseStringToBoolean(inclGovernmentOwnsAllParcels),
       };
 
+      debugger;
```
mhuseinov, line 142: looks like a debug leftover

mhuseinov (approved):

PR #848, opened 2023-08-03
Feature/alcs 799

lstod:
NOI document linking

Currently new table is called noi_document. Could be changed to notice_of_intent_document if needed


bin/migrate-oats-data/sql/documents/documents.sql
```diff
@@ -30,4 +30,4 @@
 		END AS "system"
     FROM 
       oats_documents_to_insert oti
-      JOIN alcs.application a ON a.file_number = oti.alr_application_id::text
+      JOIN alcs.application a ON a.file_number = oti.alr_application_id::text
```
mhuseinov, line 33: This "join" will filter down the documents only to application documents is it expected?

lstod: documents_noi handles insertion of noi related documents
that joins on alcs.notice_of_intent

mhuseinov (approved):

PR #851, opened 2023-08-03
Remove expiry date from naru and tur proposals

sandratoh:
- Delete blank tur proposal component after removing only field
- Update modules from non-existent tur component


dhaselhan (approved):
LGTM 👍

PR #855, opened 2023-08-04
Sort columns and implement column delete

dhaselhan:
* Sort columns while mapping them
* Hook up delete button and add to controller


sandratoh (approved):
LGTM 👍

PR #861, opened 2023-08-09
Feature/alcs 1011

lstod:
update app fee fields


bin/migrate-oats-data/applications/sql/application-populate/application_pop_fee.sql
```diff
@@ -0,0 +1,16 @@
+WITH
```
mhuseinov, line 1: Please provide some comments to SQL, for example, why only records with fee_amount equal to NUll are selected?

PR #864, opened 2023-08-11
Add more subtext

dhaselhan:
* Add subtext to all form fields so they line up
* Change some fields back to non-full width
* Remove a font size in typography so it can be changed


sandratoh (approved):
LGTM 👍

PR #876, opened 2023-08-15
Send application status emails

sandratoh:


services/apps/alcs/src/alcs/application-decision/application-decision-v2/application-decision/application-decision-v2.controller.ts
```diff
@@ -198,6 +204,55 @@ export class ApplicationDecisionV2Controller {
       modifies,
       reconsiders,
     );
+
+    // Send email after releasing draft
+    if (updateDto.isDraft === false) {
```
dhaselhan, line 209: I think this is only supposed to happen when the first decision is released, and reverting to draft then releasing again shouldn't do that.

dhaselhan: Put this block into its own function

services/apps/alcs/src/alcs/application/application.controller.ts
```diff
@@ -122,6 +128,31 @@ export class ApplicationController {
   @Post('/:fileNumber/cancel')
   @UserRoles(...ROLES_ALLOWED_APPLICATIONS)
   async cancel(@Param('fileNumber') fileNumber): Promise<void> {
+    const applicationSubmission = await this.applicationSubmissionService.get(
```
dhaselhan, line 131: The block of loading the submission finding the primary contact, then loading the submission government seems pretty common. Should there be a SubmissionStatusEmailService that DRYS this up?

sandratoh: Updated!

dhaselhan (approved):
LGTM 👍

PR #886, opened 2023-08-17
Set Applicant on NOI / Application when updating non-draft owners

dhaselhan:


services/apps/alcs/src/portal/application-submission/application-owner/application-owner.service.ts
```diff
@@ -290,6 +292,17 @@ export class ApplicationOwnerService {
...
           });
+
+          const fileNumber =
+            await this.applicationSubmissionService.getFileNumber(
+              submissionUuid,
+            );
+          if (fileNumber) {
+            await this.applicationService.updateApplicant(
+              fileNumber,
+              applicantName || 'Unknown',
```
mhuseinov, line 303: nitpick: 'Unknown' is perfect candidate for constant

mhuseinov (approved):

services/apps/alcs/src/portal/application-submission/application-owner/application-owner.service.ts
dhaselhan: Good call, there were actually like 7 of them

dhaselhan: Made a FALLBACK_APPLICANT_NAME const and used everywhere in the backend

PR #899, opened 2023-08-21
Submission Base Field Import

lstod:
A copy & paste of feature/ALCS-1013 that has a clean commit history and has not strayed from the known path of git

### DIFF
--- a/bin/migrate-oats-data/migrate.py
+++ b/bin/migrate-oats-data/migrate.py
@@ -324,7 +324,7 @@ if __name__ == "__main__":
 
                     process_noi_documents(batch_size=import_batch_size)
             case "app-sub-import":
-                console.log("Beginning OATS -> ALCS app-prep import process")
+                console.log("Beginning OATS -> ALCS app-sub import process")
                 with console.status(
                     "[bold green]App submission import (application_submission table update in ALCS)..."
                 ) as status:
--- a/bin/migrate-oats-data/migrate.py
+++ b/bin/migrate-oats-data/migrate.py
@@ -20,6 +20,10 @@ from noi import (
     process_nois,
     clean_nois,
 )
+from submissions import(
+    process_alcs_app_submissions,
+    clean_application_submission,
+)
 from constants import BATCH_UPLOAD_SIZE
 
 import_batch_size = BATCH_UPLOAD_SIZE
@@ -142,6 +146,20 @@ def import_command_parser(subparsers):
     )
     import_command.set_defaults(func=import_batch_size)
 
+def app_sub_import_command_parser(import_batch_size, subparsers):
+    app_sub_import_command = subparsers.add_parser(
+        "app-sub-import",
+        help=f"Import App submission into ALCS (update application_submission table) in specified batch size: (default: {import_batch_size})",
+    )
+    app_sub_import_command.add_argument(
+        "--batch-size",
+        type=int,
+        default=import_batch_size,
+        metavar="",
+        help=f"batch size (default: {import_batch_size})",
+    )
+    app_sub_import_command.set_defaults(func=import_batch_size)
+
 
 def setup_menu_args_parser(import_batch_size):
     parser = argparse.ArgumentParser(description="OATS ETL utility")
@@ -156,6 +174,7 @@ def setup_menu_args_parser(import_batch_size):
     document_noi_import_command_parser(import_batch_size, subparsers)
     noi_document_import_command_parser(import_batch_size, subparsers)
     import_command_parser(subparsers)
+    app_sub_import_command_parser(import_batch_size, subparsers)
 
     subparsers.add_parser("clean", help="Clean all imported data")
 
@@ -210,6 +229,9 @@ if __name__ == "__main__":
                     console.log("Processing application prep:")
                     process_alcs_application_prep_fields(batch_size=import_batch_size)
 
+                    console.log("Processing application submission:")
+                    process_alcs_app_submissions(batch_size=import_batch_size)
+
                     console.log("Done")
             case "clean":
                 with console.status("[bold green]Cleaning previous ETL...") as status:
@@ -219,6 +241,7 @@ if __name__ == "__main__":
                     clean_application_documents()
                     clean_noi_documents()
                     clean_documents()
+                    clean_application_submission()
                     clean_applications()
                     clean_nois()
 
@@ -300,6 +323,19 @@ if __name__ == "__main__":
                     )
 
                     process_noi_documents(batch_size=import_batch_size)
+            case "app-sub-import":
+                console.log("Beginning OATS -> ALCS app-prep import process")
+                with console.status(
+                    "[bold green]App submission import (application_submission table update in ALCS)..."
+                ) as status:
+                    if args.batch_size:
+                        import_batch_size = args.batch_size
+
+                    console.log(
+                        f"Processing app-sub import in batch size = {import_batch_size}"
+                    )
+
+                    process_alcs_app_submissions(batch_size=import_batch_size)
 
     finally:
         if connection_pool:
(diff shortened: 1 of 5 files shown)


dhaselhan:
Can you give this a more descriptive name than Gitclean/alcs 1013 ?

bin/migrate-oats-data/migrate.py
```diff
@@ -300,6 +323,19 @@ def setup_menu_args_parser(import_batch_size):
                     )
 
                     process_noi_documents(batch_size=import_batch_size)
+            case "app-sub-import":
+                console.log("Beginning OATS -> ALCS app-prep import process")
```
mhuseinov, line 327: Looks like a typo 'ALCS app-prep' since you are importing submissions

bin/migrate-oats-data/submissions/__init__.py
```diff
@@ -0,0 +1 @@
+from .app_submissions import *
```
mhuseinov, line 1: Import only those items that are supposed to be accessed from outside the module. In this case, there are only
"process_alcs_app_submissions" and "clean_application_submission" so it should be the only one in import. Otherwise, it is possible to call "get_update_query_for_other" anywhere the submission module is imported.

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,262 @@
...
+        else:
+            other_data_list.append(data)
+
+    return nfu_data_list, nar_data_list, other_data_list, exc_data_list, inc_data_list
+
+
+def get_update_query(unique_fields,unique_values):
+    # unique_fields takes input from called function and appends to query
+    query = """
+                INSERT INTO alcs.application_submission (
```
mhuseinov, line 189: I'd recommend using COPY instead of insert if possible. Copy is more performant, especially for batches

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,262 @@
...
+                    successful_updates_count = (
+                        successful_updates_count + applications_to_be_updated_count
+                    )
+                    last_application_id = dict(rows[-1])["alr_application_id"]
+
+                    print(
+                        f"retrieved/updated items count: {applications_to_be_updated_count}; total successfully updated submissions so far {successful_updates_count}; last updated application_id: {last_application_id}"
+                    )
+                except Exception as e:
+                    # this is NOT going to be caused by actual data update failure. This code is only executed when the code error appears or connection to DB is lost
```
mhuseinov, line 69: The comment is incorrect. This function performs insert instead of update

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+def get_update_query_for_inc():
+    unique_fields = """,
+                    incl_excl_hectares
+    """
+    unique_values = """,
+                    %(alr_area)s
+    """
+    return get_update_query(unique_fields,unique_values)
```
mhuseinov, line 243: Looks like you do not need to distinguish between INCL and EXCL since they are the same

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+def get_update_query_for_other():
```
mhuseinov, line 246: the same as above between NARU and Others. I think you do not need this functions at all since you could just call get_update_query() in prepare_app_sub_data()

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,262 @@
...
+                    false,
+                    'oats_etl',
+                    %(applicant)s
+                    {unique_values}
+                )
+    """
+    return query.format(unique_fields=unique_fields, unique_values=unique_values)
+
+def get_update_query_for_nfu():
+    unique_fields = """,
```
mhuseinov, line 211: You do not need a pre-formatted string variable. You could use unique_fields = ", nfu_hectares" instead. the same goes for unique_values.

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,262 @@
...
+        elif data["alr_change_code"] == ALRChangeCode.INC.value:
+            # data = mapOatsToAlcsLegislationCode(data)
+            inc_data_list.append(data)
+        else:
+            other_data_list.append(data)
+
+    return nfu_data_list, nar_data_list, other_data_list, exc_data_list, inc_data_list
+
+
+def get_update_query(unique_fields,unique_values):
```
mhuseinov, line 186: rename to get_insert_query or get_copy_query based on whatever query you will end up with after refactoring

bin/migrate-oats-data/submissions/sql/app_submission.sql
```diff
@@ -0,0 +1,26 @@
+WITH
```
mhuseinov, line 1: Just to clarify so, are OATS components equivalent to ALCS Submissions? And is it 100% overlap with imported applications?

bin/migrate-oats-data/submissions/app_submissions.py
lstod: COPY is not intended for inserting constant values

bin/migrate-oats-data/submissions/app_submissions.py
lstod: Future tickets will require distinction, column ex. excl_why_land and excl_share_government_borders

bin/migrate-oats-data/submissions/app_submissions.py
lstod: NARU has several unique fields to be mapped later

bin/migrate-oats-data/submissions/app_submissions.py
lstod: when there are multiple values later on it is easier to keep them pre-formatted

mhuseinov: Agree regarding multiple values. However, since, at this point, you are not adding anything, it is better to keep this simple and add features as we go. That way, it is easier to keep code clean and consistent.

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,250 @@
...
+                    successful_inserts_count = (
+                        successful_inserts_count + applications_to_be_inserted_count
+                    )
+                    last_application_id = dict(rows[-1])["alr_application_id"]
+
+                    print(
+                        f"retrieved/inserted items count: {applications_to_be_inserted_count}; total successfully inserted submissions so far {successful_inserts_count}; last inserted application_id: {last_application_id}"
+                    )
+                except Exception as e:
+                    # this is NOT going to be caused by actual data insert failure. This code is only executed when the code error appears or connection to DB is lost
```
mhuseinov, line 69: Actually, it is not true. Unlike the "update", "insert" will return an error if it fails to insert a record due to FK constraint or field type mismatch and etc.

bin/migrate-oats-data/submissions/app_submissions.py
```diff
@@ -0,0 +1,250 @@
...
+                cursor.execute(
+                    f"{application_sql} WHERE acg.alr_application_id > {last_application_id} ORDER BY acg.alr_application_id;"
+                )
+
+                rows = cursor.fetchmany(batch_size)
+
+                if not rows:
+                    break
+                try:
+                    applications_to_be_inserted_count = len(rows)
```
mhuseinov, line 56: also, could you make sure that you rename variables to represent what they actually are. For example here you are inserting the submission and not applications.

mhuseinov (approved):

PR #909, opened 2023-08-24
Decision Fixes

dhaselhan:
* Update styling for POFO/PFRS/ROSO Input Components to match new designs
* Update document table to have upload button on the right like other sections


sandratoh (approved):
LGTM 👍

PR #910, opened 2023-08-24
NOI Decisions V2 Pt2

dhaselhan:
* Update UI to match NOI
* Update number generation for decisions to include NOI and Application Decisions
* Rename DecisionDocumentDto -> ApplicationDecisionDocumentDto

### DIFF
--- a/services/apps/alcs/src/providers/typeorm/migrations/1692902429290-update_dec_number_generation_for_nois.ts
+++ b/services/apps/alcs/src/providers/typeorm/migrations/1692902429290-update_dec_number_generation_for_nois.ts
@@ -1,0 +1,54 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class updateDecNumberGenerationForNois1692902429290
+  implements MigrationInterface
+{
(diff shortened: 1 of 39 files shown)


services/apps/alcs/src/providers/typeorm/migrations/1692902429290-update_dec_number_generation_for_nois.ts
```diff
@@ -0,0 +1,54 @@
...
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE OR REPLACE FUNCTION alcs.generate_next_resolution_number(p_resolution_year integer)
+          RETURNS integer
+          LANGUAGE plpgsql
+        AS $function$
+          
+            declare next_resolution_number integer;
```
mhuseinov, line 13: This function is almost perfectly symmetric.  I think it will be a good idea to add one or two sentences explaining what is happening here.

mhuseinov (approved):

PR #917, opened 2023-08-28
Send NOI decision released status emails

sandratoh:


services/apps/alcs/src/portal/notice-of-intent-submission/notice-of-intent-submission.service.ts
```diff
@@ -61,6 +61,21 @@ export class NoticeOfIntentSubmissionService {
     @InjectMapper() private mapper: Mapper,
   ) {}
 
+  async getOrFailByFileNumber(fileNumber: string) {
```
dhaselhan, line 64: You should be okay to just use .findOneOrFail and remove your manual checking. You also need to add a test for this function.

sandratoh: Updated!

dhaselhan (approved):
LGTM 👍

PR #928, opened 2023-08-31
Bug Fixes Pt3

dhaselhan:
* Change componentLabels -> componentLabelsStr to prevent the labels from disappearing on both NOI and Application Conditions pages when updating
* Add an extra check to allow 33.1 recons to not require approval to be used as links
* Mark the decision components as touched when validating so they show red error borders
* Load components attached to conditions so they show on the NOI conditions page


sandratoh (approved):
LGTM 👍

PR #931, opened 2023-08-31
Fix portal bugs with edit submissions and reviews

sandratoh:
- Change delete confirmation copy
- Add error toasts to L/FNG Review


dhaselhan (approved):
LGTM 👍

PR #932, opened 2023-08-31
Change L/FNG Review to only call save when components are dirty

dhaselhan:
* Also don't reload application / documents / etc when changing steps


sandratoh (approved):
LGTM 👍

services/apps/alcs/src/providers/typeorm/datasource.cli.orm.config.ts
```diff
@@ -20,4 +20,5 @@ export const connectionSource = new DataSource({
   migrations: [join(__dirname, '**', 'migrations/*{.ts,.js}')],
   namingStrategy: new SnakeNamingStrategy(),
   uuidExtension: 'pgcrypto',
+  maxQueryExecutionTime: 500,
```
dhaselhan, line 23: Move to config

PR #947, opened 2023-09-06
Add Primary Contact and L/FNG Selection for Notifications

dhaselhan:
* Also fix small bug with Parcels using the wrong service


portal-frontend/src/app/features/notifications/edit-submission/edit-submission.component.ts
```diff
@@ -129,8 +136,18 @@ export class EditSubmissionComponent implements OnDestroy, AfterViewInit {
         }
         break;
       case EditNotificationSteps.Transferees:
+        //DO NOTHING
```
mhuseinov, line 139: Do I get it right that this will be addressed in next tickets?

mhuseinov (approved):

portal-frontend/src/app/features/notifications/edit-submission/edit-submission.component.ts
dhaselhan: Nope, that is correct. All transferee actions are saved via the dialog and the page itself has no transient state to save.

PR #950, opened 2023-09-06
update timeline and decision emails

mhuseinov:
- update timeline for applications and notifications
- remove decision date from templates for noi and applications


dhaselhan (approved):
LGTM 👍

PR #969, opened 2023-09-13
Feature/alcs 1117

mhuseinov:
- Introduce "smart batch" that takes into account both batch size in bytes as well as the total count of documents
- Refactor code so all document import scripts share the same core functionality 
- Allow running import commands for different file types in separate terminals
  - Application documents could be run in multiple terminals with batch window


dhaselhan (approved):
LGTM 👍

PR #973, opened 2023-09-15
search change requests and fixes

mhuseinov:
- fix submitted date search
- fix region code search
- change no data
- change search on the first load
- open IMAP in separate tab
- rename search tab


dhaselhan (approved):
LGTM 👍

PR #974, opened 2023-09-15
Add SRW Cards and SRW Board

dhaselhan:
* Add notification card dialog
* Add new board, with new statuses
* Prevent allowing create for NOTIs on any board
* Auto-create GIS Subtask and move card to NOTI board on Submission

### DIFF
--- a/services/apps/alcs/src/alcs/admin/board-management/board-management.controller.ts
+++ b/services/apps/alcs/src/alcs/admin/board-management/board-management.controller.ts
@@ -49,7 +49,7 @@ export class BoardManagementController {
   @UserRoles(AUTH_ROLE.ADMIN)
   async canDelete(@Param('code') code: string) {
     //If it's the vetting board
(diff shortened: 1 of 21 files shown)


services/apps/alcs/src/alcs/admin/board-management/board-management.controller.ts
```diff
@@ -49,7 +49,7 @@ export class BoardManagementController {
   @UserRoles(AUTH_ROLE.ADMIN)
   async canDelete(@Param('code') code: string) {
     //If it's the vetting board
```
mhuseinov, line 51: nitpick the comment does not match the code

mhuseinov (approved):

services/apps/alcs/src/alcs/admin/board-management/board-management.controller.ts
dhaselhan: Fixed

PR #976, opened 2023-09-15
adjust date to retrieve submission statuses for email consumers

mhuseinov:
- adjust date filter for email consumer

### DIFF
--- a/services/apps/alcs/src/alcs/application/application-submission-status/application-submission-status.service.ts
+++ b/services/apps/alcs/src/alcs/application/application-submission-status/application-submission-status.service.ts
@@ -4,7 +4,7 @@ import { InjectRepository } from '@nestjs/typeorm';
 import * as dayjs from 'dayjs';
 import * as timezone from 'dayjs/plugin/timezone';
 import * as utc from 'dayjs/plugin/utc';
-import { In, IsNull, LessThanOrEqual, Repository } from 'typeorm';
+import { And, In, IsNull, LessThan, Not, Repository } from 'typeorm';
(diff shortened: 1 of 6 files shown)


services/apps/alcs/src/alcs/application/application-submission-status/application-submission-status.service.ts
```diff
@@ -152,10 +152,12 @@ export class ApplicationSubmissionStatusService {
           SUBMISSION_STATUS.IN_REVIEW_BY_ALC,
         ]),
         emailSentDate: IsNull(),
-        effectiveDate: LessThanOrEqual(date), // this will get only statuses <= today since the status service converts all days to .startOf('day')
+        effectiveDate: And(Not(IsNull()), LessThan(date)), // this will get only statuses < today+1 since the status service converts all days to .startOf('day')
```
dhaselhan, line 155: Tomorrow might be a clearer way than saying today +1, but why do we want to include statuses from tomorrow?

mhuseinov: They will not be included since it checks effectiveDate < tomorrow start of day.

services/apps/alcs/src/queues/scheduler/application/status-emails/status-emails.consumer.ts
```diff
@@ -33,6 +33,7 @@ export class ApplicationSubmissionStatusEmailConsumer {
       const today = dayjs(new Date())
```
dhaselhan, line 33: Is this still today or is this tomorrow?

dhaselhan (approved):
LGTM 👍

PR #977, opened 2023-09-18
additional log messages

mhuseinov:
I need some additional log messages to verify logger and if consumers are starting on dev env


dhaselhan (approved):
LGTM 👍

PR #979, opened 2023-09-18
Add Intake Page

dhaselhan:
* Add ability to re-send email if it failed originally
* Delete ALCS Noti Submission Service, use Portals instead


services/apps/alcs/src/alcs/local-government/local-government.service.ts
```diff
@@ -51,7 +51,7 @@ export class LocalGovernmentService {
   }
 
   async getByUuid(uuid: string) {
-    return this.repository.findOne({
+    return this.repository.findOneOrFail({
```
mhuseinov, line 54: nitpick: await for stack trace?

mhuseinov (approved):

services/apps/alcs/src/alcs/local-government/local-government.service.ts
dhaselhan: done!

PR #980, opened 2023-09-18
Feature/alcs 1132 etl and search bonus

mhuseinov:
- set email_sent_dates submission to statuses table for all etl imported submissions 
- bonus: remove length limitation for file number on advanced search


dhaselhan (approved):
LGTM 👍

PR #982, opened 2023-09-18
Bug Fixes

dhaselhan:
* Fix SUBD Calculations by setting size
* Add Discussion Reviews to the App TImeline
* Full Size the Add Transferee button

### DIFF
--- a/services/apps/alcs/src/alcs/application/application-timeline/application-timeline.service.ts
+++ b/services/apps/alcs/src/alcs/application/application-timeline/application-timeline.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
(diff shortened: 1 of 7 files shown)


services/apps/alcs/src/alcs/application/application-timeline/application-timeline.service.ts
```diff
@@ -1,6 +1,7 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
+import { ApplicationDecisionMeetingService } from '../../application-decision/application-decision-v1/application-decision-meeting/application-decision-meeting.service';
```
mhuseinov, line 4: Is it correct that you are using decision version 1 ?

mhuseinov (approved):

services/apps/alcs/src/alcs/application/application-timeline/application-timeline.service.ts
dhaselhan: I think this meeting service is only in V1 but will be used even once we move to Decision V2.

PR #984, opened 2023-09-19
migrate from bull to BullMQ

mhuseinov:
- migrate from bull -> BullMQ
   - update all local jobs to support BullMQ
   - update Redis configuration to work with new BullMQ


dhaselhan (approved):
LGTM 👍

PR #985, opened 2023-09-19
Allow updating of Primary Contact Email for SRWs

dhaselhan:


services/apps/alcs/src/alcs/notification/notification-submission/notification-submission.controller.ts
```diff
@@ -64,4 +64,29 @@ export class NotificationSubmissionController {
...
   }
+
+  @UserRoles(...ANY_AUTH_ROLE)
+  @Patch('/:fileNumber')
+  async update(
+    @Param('fileNumber') fileNumber: string,
+    @Body('contactEmail') contactEmail: string,
+    @Req() req,
+  ) {
+    if (!contactEmail) {
```
mhuseinov, line 75: Do you need to make sure that the contactEmail is not a white space? Something like contactEmail.trim() ?

dhaselhan: Its easy to add, might as well :)

mhuseinov (approved):

services/apps/alcs/src/alcs/notification/notification-submission/notification-submission.controller.ts
dhaselhan: Done!

PR #994, opened 2023-09-21
Feature/alcs 1129 - import data from OATS to ALCS -> notice_of_intent table part 1

mhuseinov:
- import into the notice_of_intent table
- fix the import issue with the document
- bonus: open links in a new tab

TBD: The second part will consist of mostly refactoring and cleaning up


dhaselhan (approved):
LGTM 👍

PR #1000, opened 2023-09-22
convert cron expression to utc to properly map to PST or PDT

mhuseinov:
- introduce isDST function
- use utc corn expression to match desired PST or PDT time


dhaselhan (approved):
LGTM 👍

PR #1001, opened 2023-09-22
Only set ALC Response Sent status if Noti Email Sends

dhaselhan:
* Add tests for it
* Update email service to return true/false
* Minor UI fix to Inbox

### DIFF
--- a/services/apps/alcs/src/providers/email/email.service.ts
+++ b/services/apps/alcs/src/providers/email/email.service.ts
@@ -164,7 +164,7 @@ export class EmailService {
         ),
       );
       this.logger.debug({ to, from, subject }, `Email sent`);
-      this.repository.save(
+      await this.repository.save(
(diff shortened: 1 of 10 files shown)


services/apps/alcs/src/providers/email/email.service.ts
```diff
@@ -182,7 +183,7 @@ export class EmailService {
         errorMessage = e.response.data.detail;
       }
 
-      this.repository.save(
+      await this.repository.save(
```
mhuseinov, line 186: Is it critical that data will not be recorded into db if save fails in a catch?

mhuseinov (approved):

services/apps/alcs/src/providers/email/email.service.ts
dhaselhan: I dunno, I can remove this

PR #1007, opened 2023-09-25
fix notification and noi table sorting

mhuseinov:
- fix civic address search
- fix sorting on noi and notifications


sandratoh (approved):
LGTM 👍

PR #1008, opened 2023-09-25
Base Views for Public Search

dhaselhan:
* Add Public Module with Search
* Add new views with public filters


services/apps/alcs/src/portal/public/search/application/public-application-search.service.ts
```diff
@@ -0,0 +1,305 @@
...
+  private compileApplicationDecisionSearchQuery(
+    searchDto: SearchRequestDto,
+    query,
+  ) {
+    if (
+      searchDto.dateDecidedTo !== undefined ||
+      searchDto.dateDecidedFrom !== undefined
+    ) {
+      query = this.joinApplicationDecision(query);
+      // TODO
```
mhuseinov, line 166: I assume this will be addressed in later mrs, correct?

services/apps/alcs/src/portal/public/search/application/public-application-search.service.ts
```diff
@@ -0,0 +1,305 @@
...
+        );
+    }
+    return query;
+  }
+
+  private compileApplicationFileTypeSearchQuery(
+    searchDto: SearchRequestDto,
+    query,
+  ) {
+    if (searchDto.fileTypes.length > 0) {
```
mhuseinov, line 272: do you need to adjust this code according to the commented out decision code above?

mhuseinov (approved):

services/apps/alcs/src/portal/public/search/application/public-application-search.service.ts
dhaselhan: Yes

services/apps/alcs/src/portal/public/search/application/public-application-search.service.ts
dhaselhan: Maybe, I'm not sure yet

PR #1009, opened 2023-09-25
Bugfix/alcs 1038 home page for notifications

mhuseinov:
- assigned to me notifications 
- sorting
- notification subtasks on home page
Bonus:
exclude non-app from advanced search by pid


dhaselhan (approved):
LGTM 👍

PR #1015, opened 2023-09-27
Improve searchbar interaction and copy file ID to advanced search

sandratoh:
- Use ViewChildren and QueryList to access ngif search input field
- Focus input field on search icon toggle
- Select input field if no result is found
- Copy File ID to advanced search page if search text was entered to input field


dhaselhan (approved):
LGTM 👍

PR #1025, opened 2023-09-29
Portal table entry bug fixes

sandratoh:
- Save NARU soil and fill table data on change
- Remove default value set in soil table
- Select input on focus to prevent user error from ngx-mask validation

### DIFF
--- a/portal-frontend/src/app/features/applications/edit-submission/proposal/pfrs-proposal/pfrs-proposal.component.ts
+++ b/portal-frontend/src/app/features/applications/edit-submission/proposal/pfrs-proposal/pfrs-proposal.component.ts
@@ -88,10 +88,10 @@ export class PfrsProposalComponent extends FilesStepComponent implements OnInit,
         this.submissionUuid = applicationSubmission.uuid;
 
         this.alreadyRemovedTableData = {
-          volume: applicationSubmission.soilAlreadyRemovedVolume ?? 0,
-          area: applicationSubmission.soilAlreadyRemovedArea ?? 0,
-          averageDepth: applicationSubmission.soilAlreadyRemovedAverageDepth ?? 0,
-          maximumDepth: applicationSubmission.soilAlreadyRemovedMaximumDepth ?? 0,
+          volume: applicationSubmission.soilAlreadyRemovedVolume ?? undefined,
+          area: applicationSubmission.soilAlreadyRemovedArea ?? undefined,
+          averageDepth: applicationSubmission.soilAlreadyRemovedAverageDepth ?? undefined,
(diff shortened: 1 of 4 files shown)


portal-frontend/src/app/features/applications/edit-submission/proposal/pfrs-proposal/pfrs-proposal.component.ts
```diff
@@ -88,10 +88,10 @@ export class PfrsProposalComponent extends FilesStepComponent implements OnInit,
         this.submissionUuid = applicationSubmission.uuid;
 
         this.alreadyRemovedTableData = {
-          volume: applicationSubmission.soilAlreadyRemovedVolume ?? 0,
-          area: applicationSubmission.soilAlreadyRemovedArea ?? 0,
-          averageDepth: applicationSubmission.soilAlreadyRemovedAverageDepth ?? 0,
-          maximumDepth: applicationSubmission.soilAlreadyRemovedMaximumDepth ?? 0,
+          volume: applicationSubmission.soilAlreadyRemovedVolume ?? undefined,
```
mhuseinov, line 91: This will prevent the user from storing empty data if it was previously populated

sandratoh: Just tested, and the form still saves the value if '0' was entered. Also, using `undefined` is the default behaviour of 'Soil to be Removed / Fill to be Placed' table. For some reason, that table was set to `undefined`, and 'Soil Already Removed / Fill Already Placed' table was '0' previously.

portal-frontend/src/app/shared/soil-table/soil-table.component.html
```diff
@@ -50,6 +51,7 @@
             separatorLimit="9999999999"
             matInput
             formControlName="volume2"
+            onfocus="this.select()"
```
mhuseinov, line 54: I wonder if this will cause validation to show up as soon as user clicks on the fields?

portal-frontend/src/app/features/applications/edit-submission/proposal/pfrs-proposal/pfrs-proposal.component.ts
mhuseinov: Have you tried using "empty" (delete the value completely from the field), not 0?

mhuseinov (approved):

portal-frontend/src/app/features/applications/edit-submission/proposal/pfrs-proposal/pfrs-proposal.component.ts
sandratoh: Tested together and there seems to be something in place to handle undefined value already 👍

PR #1026, opened 2023-09-29
notice of intent submission proposal and soil default fields

mhuseinov:
- NOIS proposal fields and soil fields with default values
- validation script


lstod (approved):
💯

PR #1027, opened 2023-10-03
Feature/alcs 1185 improvements

mhuseinov:
- introduce new logger for etl
- replace all 'print' and 'old logger' with new logger functionality


lstod (approved):
🐎

PR #1036, opened 2023-10-06
Update app, noi and srw review layout

sandratoh:


portal-frontend/src/app/features/applications/application-details/application-details.component.scss
```diff
@@ -60,7 +60,7 @@
 
   .other-attachments {
     display: grid;
-    grid-template-columns: max-content max-content max-content;
```
dhaselhan, line 63: This was required for the small screen size layout to expand and overflow with a scrollbar. Does it still work with 1fr?

dhaselhan (approved):
LGMT 👍  Responsive to be fixed in future MR

PR #1046, opened 2023-10-10
Add default visibility flags to portal documents

dhaselhan:
* If document is certain type add public visibility


services/apps/alcs/src/portal/notification-document/notification-document.controller.ts
mhuseinov, line 1: Are there any tests that need to be updated for this controller?

mhuseinov (approved):

services/apps/alcs/src/portal/notification-document/notification-document.controller.ts
dhaselhan: I just added a test to the app doc controller for public visibility, I wasn't sure it was worth it to copy paste that test to the two other document controllers. Do you think I should copy and paste it?

dhaselhan: Done! Tests added.

PR #1052, opened 2023-10-12
Fix standalone noi-import 

lstod:
allows noi import to run without applications being imported first


bin/migrate-oats-data/noi/notice_of_intent_init.py
```diff
@@ -9,6 +14,8 @@
 def init_notice_of_intents(conn=None, batch_size=10000):
     logger.info(f"Start {etl_name}")
     with conn.cursor() as cursor:
+        exclusion_table_create(cursor, conn)
```
mhuseinov, line 17: Will this cause an issue when running the following: python migrate.py import? Will it introduce duplicates of records to exclude or fail?

mhuseinov (approved):

PR #1060, opened 2023-10-16
Bugfix/task 1189 make civic address search case insensitive

urmikgov:


dhaselhan:
Suggestion: Update the MR to have a clearer description like: Lowercase civic address in search. 
Question: It also looks like this fix is only for Applications, does it need to be applied to NOIs and SRWs also?

urmikgov:
@dhaselhan not sure about NOIs and SRWs , it just gave me same number of applications but i need to checkin about this with sandra

dhaselhan (approved):
LGTM 👍

PR #1061, opened 2023-10-16
Add validation for SRW Survey Plan Numbers

dhaselhan:


services/apps/alcs/src/portal/notification-submission/notification-submission-validator.service.ts
```diff
@@ -228,6 +228,16 @@ export class NotificationSubmissionValidatorService {
           ),
         );
       }
+
+      for (const plan of surveyPlans) {
+        if (!plan.surveyPlanNumber || !plan.controlNumber) {
+          errors.push(
+            new ServiceValidationException(
+              `${notificationSubmission.typeCode} proposal has survey plans missing survey/control plan numbers`,
```
sandratoh, line 236: Is it `typeCode` or `type.code`?

dhaselhan: They will be the same value, so mostly interchangeable.

I changed the other places to keep the mocking consistent in the tests (mock type.code)

sandratoh (approved):

PR #1068, opened 2023-10-17
ALCS PROD DB obfuscation

mhuseinov:
- ALCS PROD DB obfuscation
- update READ.ME


bin/migrate-oats-data/menu/commands/obfuscation.py
lstod, line 1: Is there a scenario where we would want to run alcs and oats obfuscation independently?

lstod (approved):

bin/migrate-oats-data/menu/commands/obfuscation.py
mhuseinov: not really, hopefully, this is a time task and we do not need to repeat it

PR #1075, opened 2023-10-18
Feature/alcs 1180 Add legacy id to NOIs

urmikgov:


services/apps/alcs/src/alcs/notice-of-intent/notice-of-intent.entity.ts
```diff
@@ -157,6 +157,8 @@ export class NoticeOfIntent extends Base {
   })
   alrArea?: number | null;
 
+  legacyId?: string | null;
```
dhaselhan, line 160: Missing Column and Automap annotions.

services/apps/alcs/src/alcs/search/notice-of-intent/notice-of-intent-search-view.entity.ts
```diff
@@ -92,6 +93,9 @@ export class NoticeOfIntentSubmissionSearchView {
   @ViewColumn()
   decisionDate: Date | null;
 
+  @ViewColumn()
+  legacyId?: string;
```
dhaselhan, line 97: Since you added a new ViewColumn, you'll need to use TypeORMs generate feature to create SQL to re-create the view with the new column.

dhaselhan (approved):
LGTM 👍

PR #1076, opened 2023-10-18
Fix oats -> alcs document processing

lstod:
imported correct functions, added comments & updated console log

### DIFF
--- a/bin/migrate-oats-data/menu/commands/import_all.py
+++ b/bin/migrate-oats-data/menu/commands/import_all.py
@@ -42,21 +42,22 @@ def import_all(console, args):
 
         # OATS -> ALCS documents = process_documents & process_documents_noi. Must take place in order to have usable rows in ALCS documents
         # ALCS documents -> application_documents/noi_documents = process_application_documents & process_noi_documents
+
+        # documents_noi brings noi labelled documents into ALCS documents
         console.log("Importing NOI specific OATS documents into ALCS:")
         process_documents_noi(batch_size=import_batch_size)
-        # documents_noi brings noi labelled documents into ALCS documents
 
+        # documents brings application labelled documents into ALCS documents
         console.log("Importing OATS app_documents into ALCS:")
         process_documents(batch_size=import_batch_size)
-        # documents brings application labelled documents into ALCS documents
 
+        # process_application_documents takes the imported documents from process_documents
         console.log("Processing ALCS application documents:")
         process_application_documents(batch_size=import_batch_size)
-        # process_application_documents takes the imported documents from process_documents
 
+        # process_noi_documents takes the imported documents from process_documents_noi
         console.log("Processing ALCS NOI documents:")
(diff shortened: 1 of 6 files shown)


bin/migrate-oats-data/menu/commands/import_all.py
```diff
@@ -38,18 +40,23 @@ def import_all(console, args):
...
+        # documents brings application labelled documents into ALCS documents
 
-        console.log("Processing application documents:")
+        console.log("Processing ALCS application documents:")
         process_application_documents(batch_size=import_batch_size)
+        # process_application_documents takes the imported documents from process_documents
 
-        console.log("Processing NOI documents:")
+        console.log("Processing ALCS NOI documents:")
         process_noi_documents(batch_size=import_batch_size)
```
mhuseinov, line 58: Do I get it right that the process_noi_documents is responsible for linking alcs.documents to notice_of_intent_documents?

bin/migrate-oats-data/menu/commands/import_all.py
```diff
@@ -38,18 +40,23 @@ def import_all(console, args):
         console.log("Init NOIs:")
         init_notice_of_intents(batch_size=import_batch_size)
 
-        # TODO Liam question: which process_documents_noi or process_noi_documents is the correct one to keep?
-        # console.log("Processing NOI specific documents:")
-        # process_documents_noi(batch_size=import_batch_size)
+        # OATS -> ALCS documents = process_documents & process_documents_noi. Must take place in order to have usable rows in ALCS documents
+        # ALCS documents -> application_documents/noi_documents = process_application_documents & process_noi_documents
+        console.log("Importing NOI specific OATS documents into ALCS:")
+        process_documents_noi(batch_size=import_batch_size)
```
mhuseinov, line 46: Do I get it right that the process_documents_noi is responsible for importing oats.documents that belong to NOI to alcs.documents?

bin/migrate-oats-data/menu/commands/import_all.py
mhuseinov: also, the comment should be above the code that it describes

bin/migrate-oats-data/menu/commands/import_all.py
lstod: That is correct

bin/migrate-oats-data/menu/commands/import_all.py
lstod: Also correct. Will change comment order

bin/migrate-oats-data/menu/commands/import_all.py
mhuseinov: Lets then update the naming to: process_documents_noi -> import_oats_noi_documents, process_noi_documents-> link_alcs_documents_to_noi_documents. And the same style for applications. That way it will be clear and we do not need comments at all

mhuseinov (approved):
🎯

PR #1084, opened 2023-10-20
search criteria fix

urmikgov:


services/apps/alcs/src/alcs/search/search.controller.spec.ts
```diff
@@ -346,7 +346,7 @@ describe('SearchController', () => {
 
     expect(
       mockNoticeOfIntentAdvancedSearchService.searchNoticeOfIntents,
-    ).toBeCalledTimes(0);
+    ).toBeCalledTimes(1);
```
dhaselhan, line 349: suggestion: Can you update the test description to match what you have changed?

services/apps/alcs/src/alcs/search/search.controller.spec.ts
```diff
@@ -346,7 +346,7 @@ describe('SearchController', () => {
 
     expect(
```
dhaselhan, line 347: suggestion: Now that the test is focused on Non Applications, you should remove the Notice of Intent expects as they are not pertinent.

dhaselhan (approved):
LGTM 👍

PR #1089, opened 2023-10-23
Update Decisions + Add Copying

dhaselhan:
* Remove certain outcomes
* Remove linked outcome resolution for App Decisions
* Add ability and unit tests for copying an existing decision

### DIFF
--- a/services/apps/alcs/src/alcs/notice-of-intent-decision/notice-of-intent-decision-v2/notice-of-intent-decision-v2.service.ts
+++ b/services/apps/alcs/src/alcs/notice-of-intent-decision/notice-of-intent-decision-v2/notice-of-intent-decision-v2.service.ts
@@ -6,6 +6,7 @@ import { MultipartFile } from '@fastify/multipart';
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { In, IsNull, Repository } from 'typeorm';
+import { v4 } from 'uuid';
 import {
   DOCUMENT_SOURCE,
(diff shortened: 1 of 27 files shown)


services/apps/alcs/src/alcs/notice-of-intent-decision/notice-of-intent-decision-v2/notice-of-intent-decision-v2.service.ts
```diff
@@ -297,21 +304,71 @@ export class NoticeOfIntentDecisionV2Service {
...
+        (component) => component.uuid,
+      );
+      return new NoticeOfIntentDecisionCondition({
+        ...condition,
+        uuid: undefined,
+        components: savedDecision.components.filter((component) => {
+          const oldUuid = newToOldComponentsUuid.get(component.uuid);
+          return (
+            !!oldUuid &&
+            conditionsComponents &&
```
sandratoh, line 363: Nitpick: Do we need `&& conditionsComponents`? Application service doesn't have it

sandratoh (approved):
Had one nitpick comment but LGTM 👍

services/apps/alcs/src/alcs/notice-of-intent-decision/notice-of-intent-decision-v2/notice-of-intent-decision-v2.service.ts
dhaselhan: This was the slight difference between using ! and using ? on the .components of the decision. I've changed them both to use ? so that they at least match.

PR #1093, opened 2023-10-24
Update Portal Decisions

dhaselhan:
* Replace outcome -> description
* Update file listing to new UX
* Select last tab if app is in final status
* Map fileSize as it was not sent previously
* Repeat a bunch of times for each page

### DIFF
--- a/portal-frontend/src/app/features/public/notice-of-intent/public-notice-of-intent.component.ts
+++ b/portal-frontend/src/app/features/public/notice-of-intent/public-notice-of-intent.component.ts
@@ -21,6 +21,7 @@ export class PublicNoticeOfIntentComponent implements OnInit, OnDestroy {
   documents: PublicDocumentDto[] = [];
   parcels: PublicParcelDto[] = [];
   decisions: ApplicationPortalDecisionDto[] = [];
+  selectedIndex = 0;
 
   constructor(private publicService: PublicService, private route: ActivatedRoute) {}
 
@@ -39,6 +40,9 @@ export class PublicNoticeOfIntentComponent implements OnInit, OnDestroy {
       const { submission, documents, parcels, decisions } = res;
 
       this.submission = submission;
+      if (submission.status.code === SUBMISSION_STATUS.ALC_DECISION) {
+        this.selectedIndex = 2;
+      }
       this.documents = documents;
       this.parcels = parcels;
       this.decisions = decisions;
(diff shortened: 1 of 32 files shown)


portal-frontend/src/app/features/public/notice-of-intent/public-notice-of-intent.component.ts
```diff
@@ -39,6 +40,9 @@ export class PublicNoticeOfIntentComponent implements OnInit, OnDestroy {
       const { submission, documents, parcels, decisions } = res;
 
       this.submission = submission;
+      if (submission.status.code === SUBMISSION_STATUS.ALC_DECISION) {
+        this.selectedIndex = 2;
```
sandratoh, line 44:
```suggestion
        this.selectedIndex = 1;
```
NOI only has 2 tabs, previous non-public version was set to 1. Also you may want to change from `SUBMISSION_STATUS` to ` NOI_SUBMISSION_STATUS`

portal-frontend/src/app/services/public/public-notification.dto.ts
```diff
@@ -1,6 +1,7 @@
 import { ApplicationStatusDto } from '../application-submission/application-submission.dto';
```
sandratoh, line 1: Cleanup: Remove replaced import

portal-frontend/src/app/features/public/notification/public-notification.component.ts
```diff
@@ -35,6 +37,10 @@ export class PublicNotificationComponent implements OnInit, OnDestroy {
     if (res) {
       const { submission, documents, parcels } = res;
 
+      if (submission.status.code !== NOTIFICATION_STATUS.ALC_RESPONSE) {
+        this.selectedIndex = 0;
+      }
```
sandratoh, line 42: Nit: Was there a reason why public notification logic is flipped from non-public notification and all other submission types? (`selectedIndex` set to 0 first, then checked for `===`)

dhaselhan: Notifications are only public if they are currently in the ALC Response status, where as Notifications and Applications become public once they are accepted as having all items. I thought it made sense to reverse the logic to prevent the scrolling.

portal-frontend/src/app/features/public/notice-of-intent/public-notice-of-intent.component.ts
dhaselhan: Fixed! Good call the NOI Submission Status, led to fixing a few other issues

portal-frontend/src/app/services/public/public-notification.dto.ts
dhaselhan: Done!

sandratoh (approved):
LGTM 👍

PR #1096, opened 2023-10-25
Use add/remove Validators for PID

dhaselhan:
* Set validators and remove validators was clearing the format validator as well as required
* Update code to use add/remove so the format validator remains on parcel / form changes


sandratoh (approved):
LGTM 👍

PR #1100, opened 2023-10-26
Portal Proposal Fixes

dhaselhan:
* Add error class to parcel error so it gets scrolled to
* Update logic for toggling the file uploads for ROSO and PFRS to also update the other ones correctly
* Add backend validation for SUBD Proposals including lot size


sandratoh (approved):
LGTM 👍

PR #1115, opened 2023-10-31
NOI COALESCE on update for fee_paid_date

mhuseinov:
- do not override ALCS values with OATS NULL
- validation script


lstod (approved):
💯

PR #1121, opened 2023-10-31
Add Email Template for Covenants

dhaselhan:
* TURP and COVE use same government template so renamed to no-review template


sandratoh (approved):
LGTM 👍

PR #1123, opened 2023-10-31
Covenant Public View

dhaselhan:
* Add Public Covenant Proposal
* Send Transferees in the main request for Covenants

### DIFF
--- a/portal-frontend/src/app/services/public/public-application.dto.ts
+++ b/portal-frontend/src/app/services/public/public-application.dto.ts
@@ -4,6 +4,7 @@ import {
   NaruSubtypeDto,
   ProposedLot,
 } from '../application-submission/application-submission.dto';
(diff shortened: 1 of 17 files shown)


portal-frontend/src/app/services/public/public-application.dto.ts
```diff
@@ -4,6 +4,7 @@ import {
   NaruSubtypeDto,
   ProposedLot,
 } from '../application-submission/application-submission.dto';
+import { CovenantTransfereeDto, OwnerTypeDto } from '../covenant-transferee/covenant-transferee.dto';
```
sandratoh, line 7: Nit: Were the two new imports used anywhere?

sandratoh (approved):
One small nit re:imports, but LGTM 👍

portal-frontend/src/app/services/public/public-application.dto.ts
dhaselhan: Good catch, nothing gets passed Sandra 🥅

PR #1124, opened 2023-10-31
Feature/alcs 1227

mhuseinov:
- added mapping for missing application field
- do not override ALCS data with OATS null for specific fields
- validation script


lstod (approved):
⚡

PR #1127, opened 2023-11-01
Upgrade Angular

dhaselhan:
Upgrade Angular in both front ends to 16
Upgrade Node in Docker to 20
Apply fixes for upgrade issues


alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -489,7 +489,7 @@ export class BoardComponent implements OnInit, OnDestroy {
       data,
     });
 
-    dialogRef.afterClosed().subscribe((isDirty) => {
+    dialogRef.beforeClosed().subscribe((isDirty) => {
```
mhuseinov, line 492: is this an improvement, does not look like angular update

dhaselhan: This was from angular update, it was firing this after the dialog closed and we navigated to another page, bringing us right back to this page.

mhuseinov (approved):

PR #1130, opened 2023-11-02
Upgrade Backend Packages

dhaselhan:
* Ugrade NestJS and other packages
* Fix Typeorm CLI
* Remove Multitenant as it causes issues and we don't need it
* Move from @automapper to automapper, the newly maintained fork


services/apps/alcs/src/common/authorization/portal-auth-guard.service.ts
```diff
@@ -25,21 +25,20 @@ export class PortalAuthGuard implements CanActivate {
...
+    reflector: Reflector,
     private clsService: ClsService,
     private userService: UserService,
   ) {
     this.keyCloakGuard = new KeyCloakRoleGuard(
       singleTenant,
       keycloakOpts,
       logger,
-      multiTenant,
+      null!,
```
mhuseinov, line 41: Is this an intensional change?

dhaselhan: Yes, its a bit nasty but it seems like the multi tenant package of nest keycloak broke at some point and fails to import. The good news is that we don't use multi-tenant so not passing it doesn't seem to impact anything.

mhuseinov (approved):

PR #1132, opened 2023-11-02
Application Submission Status Migration

lstod:
Fixed the commit history and omitted all the files causing conflicts


bin/migrate-oats-data/menu/command_parser/application_command_parser.py
```diff
@@ -38,3 +38,32 @@ def app_sub_import_command_parser(import_batch_size, subparsers):
         metavar="",
         help=f"batch size (default: {import_batch_size})",
     )
+
+
+def app_status_import_command_parser(import_batch_size, subparsers):
```
mhuseinov, line 43: why do you need separate status command import?

bin/migrate-oats-data/migrate.py
```diff
@@ -56,6 +57,13 @@
                 application_submission_import(console, args)
             case "obfuscate":
                 start_obfuscation(console)
+            case "app-status-import":
+                application_status_import(console, args)
+            case "app-all-import":
+                application_import(console, args)
+                app_prep_import(console, args)
+                application_submission_import(console, args)
+                application_status_import(console, args)
```
mhuseinov, line 66: why do you need app-all-import? how is it different from app-import? Also, the migrate.py should not hold the logic of the sequential calling of functions. Migrate.py should only call the main function that abstracts all required child function calls.

bin/migrate-oats-data/applications/__init__.py
```diff
@@ -1,6 +1,23 @@
 from .base_applications import *
 from .app_prep import process_alcs_application_prep_fields
-from .submissions import process_alcs_app_submissions
+from .submissions import (
+    process_alcs_app_submissions,
```
mhuseinov, line 4: you do not need all these functions here, the applications module should only expose one root function for importing all statuses

bin/migrate-oats-data/applications/submissions/sql/statuses/application_decision_released.sql
```diff
@@ -0,0 +1,10 @@
+UPDATE alcs.application_submission_to_submission_status AS app_statuses
+SET
+    effective_date = app.decision_date
```
mhuseinov, line 3: since effective_date is timezone sensitive and the decision_date is UTC the timezone conversion should happen in sql. For example effective_date = app.decision_date AT TIME ZONE 'US/Pacific'. Please apply this to all other status update scripts including NOI

bin/migrate-oats-data/applications/submissions/statuses/application_status_type_code.py
```diff
@@ -0,0 +1,35 @@
+def get_app_status_type(cursor):
```
mhuseinov, line 1: where these function used?

lstod: Left over from batching script. File no longer used, will delete

bin/migrate-oats-data/applications/submissions/sql/statuses/application_decision_released.sql
mhuseinov: not relevant since we using alcs DB here

mhuseinov (approved):

PR #1144, opened 2023-11-08
switch from covert_time to add_time_zone

mhuseinov:
Oats stores dates the way users selected them so it is safe to assume that the date is actually in PST and not in UTC


lstod (approved):
🕐

PR #1145, opened 2023-11-08
Decision outcome applications,noi public search

urmikgov:


portal-frontend/src/app/features/public/search/notice-of-intent-search-table/notice-of-intent-search-table.component.ts
```diff
@@ -34,7 +34,11 @@ export class NoticeOfIntentSearchTableComponent implements OnDestroy {
   @Input() statuses: ApplicationStatusDto[] = [];
   @Output() tableChange = new EventEmitter<TableChange>();
 
-  displayedColumns = ['fileId', 'ownerName', 'type', 'portalStatus', 'lastUpdate', 'government'];
+  displayedColumns = ['fileId', 'ownerName', 'type', 'portalStatus', 'outcome', 'lastUpdate', 'government'];
+  outcomeMapping: {[code:string] : string } = {
```
dhaselhan, line 38: Rather than having 3 constants, can you move this to the public search dto and use a single constants?
I also think the type: Record<string,string> is easier to understand than {[code:string] : string }.

portal-frontend/src/app/features/public/search/public-search.component.ts
```diff
@@ -53,6 +60,7 @@ export class PublicSearchComponent implements OnInit, OnDestroy {
 
   notifications: NotificationSearchResultDto[] = [];
   notificationTotal = 0;
+  outcome = '';
```
dhaselhan, line 63: What is this used for? I see it being set here as empty but I don't see it used in the HTML or passed to another component.

portal-frontend/src/app/services/search/search.dto.ts
```diff
@@ -23,6 +23,7 @@ export interface SearchResponseDto {
   totalApplications: number;
   totalNoticeOfIntents: number;
   totalNotifications: number;
+  outcome: string;
```
dhaselhan, line 26: I don't think the main search result returns an outcome, did you mean to add this to the BaseSearchResultDto above instead?

services/apps/alcs/src/alcs/search/search.dto.ts
```diff
@@ -16,6 +16,7 @@ export class SearchResultDto {
   fileNumber: string;
   boardCode?: string;
   label?: ApplicationTypeDto;
+  outcome?: string
```
dhaselhan, line 19: I don't think you updated the ALCS search, only the portal public search, this should be undone.

services/apps/alcs/src/providers/typeorm/migrations/1699461885734-add_outcome_public_search_noi.ts
```diff
@@ -0,0 +1,16 @@
...
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`DELETE FROM "alcs"."typeorm_metadata" WHERE "type" = $1 AND "name" = $2 AND "schema" = $3`, ["VIEW","public_notice_of_intent_submission_search_view","alcs"]);
```
dhaselhan, line 12: The down here is only deleting the view, but not re-creating the previous version as it should. Did you delete too much?

dhaselhan:
I don't see any updates to the mobile view, SearchListComponent, does it also need to be updated?

services/apps/alcs/src/alcs/search/search.dto.ts
urmikgov: ticket just mentions public search

services/apps/alcs/src/providers/typeorm/migrations/1699461885734-add_outcome_public_search_noi.ts
urmikgov: i just ran npm run alcs:typeorm migration:revert

portal-frontend/src/app/features/public/search/public-search.component.ts
urmikgov: we use it in mapSearchResults at     this.outcome = searchResult.outcome;

portal-frontend/src/app/features/public/search/notice-of-intent-search-table/notice-of-intent-search-table.component.ts
urmikgov: we dont have that in srw

portal-frontend/src/app/services/search/search.dto.ts
urmikgov: thats being imported in public search dto compoonent

portal-frontend/src/app/features/public/search/notice-of-intent-search-table/notice-of-intent-search-table.component.ts
dhaselhan: The goal is to DRY up the code, since its repeated 3 times, I think we should have it once even if it doesn't involved SRWs.

portal-frontend/src/app/features/public/search/public-search.component.ts
dhaselhan: Okay, I'm a bit confused on how the main search response has an outcome. I think this is a mistake to add it to SearchResponseDto, as a search result with 10 applications cannot have a single outcome. The outcome should only be present on each result object such as ApplicationSearchResultDto and NoticeOfIntentSearchResultDto.

services/apps/alcs/src/portal/public/search/notice-of-intent/public-notice-of-intent-search-view.entity.ts
```diff
@@ -63,8 +64,10 @@ import { LinkedStatusType } from '../public-search.dto';
           qb
             .from(NoticeOfIntentDecision, 'decision_date')
             .select('MAX("date")', 'date')
+            .addSelect('outcome_code', 'outcome')
             .addSelect('notice_of_intent_uuid', 'notice_of_intent_uuid')
-            .groupBy('notice_of_intent_uuid'),
+            .groupBy('notice_of_intent_uuid')
+            .addGroupBy('outcome'),
```
dhaselhan, line 70: Adding a second group by clause will cause bugs. If a notice of intent decision has two different outcomes, grouping by both uuid and outcome will result in two rows, instead of the expected result of just the newest decision.

services/apps/alcs/src/portal/public/search/public-search.controller.ts
```diff
@@ -247,6 +250,7 @@ export class PublicSearchController {
       ownerName: notification.applicant,
       class: 'NOTI',
       status: notification.status.status_type_code,
+      outcome: null
```
dhaselhan, line 253: Since SRW's don't have outcomes, can you make the field optional so we don't need to set null here?

portal-frontend/src/app/services/search/search.dto.ts
dhaselhan: I added a comment above with more detail, but I believe this still need to be removed.

services/apps/alcs/src/portal/public/search/public-search.controller.ts
```diff
@@ -246,7 +249,7 @@ export class PublicSearchController {
       localGovernmentName: notification.localGovernmentName,
       ownerName: notification.applicant,
       class: 'NOTI',
-      status: notification.status.status_type_code,
+      status: notification.status.status_type_code
```
dhaselhan, line 252: This needs a prettier format I think. Looks like on some files your code is using a trailing comma but not all of them. For some reason on this file it looks like it IS NOT adding trailing commas.
  "trailingComma": "all" is the prettier setting in the .prettierrc

dhaselhan (approved):
LGTM 👍

dhaselhan (commented):
Just noticed there are still two migrations, do you need both? Can you remove one?

PR #1147, opened 2023-11-08
DevOps Updates

dhaselhan:
* Remove env so we don't print all our secrets
* Update services docker file to load secrets


lstod (approved):
🥇

PR #1148, opened 2023-11-08
update parcel validation scripts to respect alr_application_land_ind;…

mhuseinov:
- respect alr_application_land_ind in the validation script
- use only date part for comparison in validation script


lstod (approved):
💯

PR #1162, opened 2023-11-17
Date received all items for application and NOI

lstod:
Note this will not work with obfuscated data, to test either remove the line for 'AND proxy' or use a different who_created value in the DB

### DIFF
--- a/bin/migrate-oats-data/applications/migrate_application.py
+++ b/bin/migrate-oats-data/applications/migrate_application.py
@@ -7,6 +7,7 @@ from .submissions import (
 )
 from .base_applications import process_applications, clean_applications
 from .app_prep import process_alcs_application_prep_fields
+from .app_rx_items import update_application_date_rx_all_items
 from .application_submission_status_email import (
     process_application_submission_status_emails,
     clean_application_submission_status_emails,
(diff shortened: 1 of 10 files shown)


bin/migrate-oats-data/applications/migrate_application.py
```diff
@@ -15,6 +16,7 @@
 
 def process_application_etl(batch_size):
     process_alcs_application_prep_fields(batch_size)
+    update_application_date_rx_all_items(batch_size)
```
mhuseinov, line 19: Does 'rx' stand for 'received'?

bin/migrate-oats-data/applications/sql/application-prep/app_rx_all_items.sql
```diff
@@ -0,0 +1,11 @@
+SELECT
```
mhuseinov, line 1: This query will duplicate applications on joining the accomplishments is that expected?

bin/migrate-oats-data/applications/migrate_application.py
lstod: correct

bin/migrate-oats-data/applications/sql/application-prep/app_rx_all_items.sql
lstod: I believe it is intened in order to capture both AKC and SAL

mhuseinov (approved):

PR #1165, opened 2023-11-20
LFNG Return Fixes

dhaselhan:
* Clear submission date when returning to applicant
* Clean up logic for banners and show ALC banner when return to LFNG
* Remove old applicant field and mark owner optional

### DIFF
--- a/services/apps/alcs/src/portal/application-submission-review/application-submission-review.controller.ts
+++ b/services/apps/alcs/src/portal/application-submission-review/application-submission-review.controller.ts
@@ -44,7 +44,6 @@ export class ApplicationSubmissionReviewController {
   constructor(
     private applicationSubmissionService: ApplicationSubmissionService,
     private applicationSubmissionReviewService: ApplicationSubmissionReviewService,
-    private applicationService: ApplicationService,
     private applicationDocumentService: ApplicationDocumentService,
     private localGovernmentService: LocalGovernmentService,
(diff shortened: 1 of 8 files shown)


services/apps/alcs/src/portal/application-submission-review/application-submission-review.controller.ts
```diff
@@ -368,23 +370,10 @@ export class ApplicationSubmissionReviewController {
...
-        SUBMISSION_STATUS.SUBMITTED_TO_LG,
-        null,
-      );
-
-      await this.applicationSubmissionStatusService.setStatusDate(
-        applicationSubmission.uuid,
-        SUBMISSION_STATUS.IN_REVIEW_BY_LG,
-        null,
-      );
+      //Clear Submission Date
```
mhuseinov, line 373: So you are not cleaning the dateSubmittedToAlcs status date?

mhuseinov (approved):

services/apps/alcs/src/portal/application-submission-review/application-submission-review.controller.ts
dhaselhan: Good point! I should be clearing the submitted to ALC date when the ALC returns to L/FNG.

dhaselhan: Updated to make more sense, we now clear the submitted date at the same time/place where we clear the status.

PR #1167, opened 2023-11-20
Init OATS->ALCS Users

lstod:


bin/migrate-oats-data/common/alcs_init_users.py
```diff
@@ -0,0 +1,123 @@
...
+            encoding="utf-8",
+        ) as sql_file:
+            count_query = sql_file.read()
+            cursor.execute(count_query)
+            count_total = dict(cursor.fetchone())["count"]
+        logger.info(f"Total users data to insert: {count_total}")
+
+        failed_inserts = 0
+        successful_inserts_count = 0
+        last_submission_id = "-"
```
mhuseinov, line 34: nitpick: last_user_uuid

bin/migrate-oats-data/common/alcs_init_users.py
```diff
@@ -0,0 +1,123 @@
...
+                    bceid_guid,
+                    email,
+                    display_name,
+                    identity_provider,
+                    preferred_username,
+                    audit_created_by
+                )
+                VALUES (
+                    %(created_guid)s,
+                    '11@11',
```
mhuseinov, line 97: is email a required field? If yes have you confirmed with Ansley default value?

mhuseinov:
Could you also include import user in to both noi and application imports since those will be heavily dependent on it

bin/migrate-oats-data/common/alcs_init_users.py
lstod: is a not null. Yet to confirm, copied obfuscation

bin/migrate-oats-data/common/alcs_init_users.py
lstod: will change to guid

mhuseinov (approved):

lstod:
> Could you also include import user in to both noi and application imports since those will be heavily dependent on it
Done. noi and application clean will not clean users

PR #1168, opened 2023-11-21
Add INCG to L/FNG inbox statuses

dhaselhan:
* Allow  L/FNG to see INCG even when Submitted Date is not set.
* Add !important to beat material


sandratoh (approved):
LGTM 👍

PR #1177, opened 2023-11-22
show local parcel data on delete

urmikgov:


portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-details.component.ts
```diff
@@ -156,7 +156,7 @@ export class ParcelDetailsComponent extends StepComponent implements OnInit, Aft
       .beforeClosed()
       .subscribe((result) => {
         if (result) {
-          this.loadParcels();
+          this.parcels = this.parcels.filter(parcel => parcel.uuid !== parcelUuid);
```
trslater, line 159: Won't this then bypass the other stuff `loadParcels` does?

urmikgov: ohh, we load parcels on init of the page, but when we delete a parcel we want to keep it local since we havent made any changes by clicking save

trslater (approved):

PR #1196, opened 2023-12-04
owner deleting fix

urmikgov:


portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
```diff
@@ -92,13 +93,17 @@ export class ParcelOwnersComponent {
...
     dialog.beforeClosed().subscribe((updatedUuid) => {
-      if (updatedUuid) {
+      if (updatedUuid?.ownerDeleted) {
+        this.onOwnersDeleted.emit(updatedUuid.deletedOwnerId);
+      }
+      else if (updatedUuid) {
         this.onOwnersUpdated.emit();
       }
     });
     dialog.afterClosed().subscribe((result) => {
```
dhaselhan, line 103: Is there a reason we need both onBeforeClosed and onAfterClosed? I recommend you only use one.

portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
```diff
@@ -92,13 +93,17 @@ export class ParcelOwnersComponent {
       });
     }
     dialog.beforeClosed().subscribe((updatedUuid) => {
-      if (updatedUuid) {
+      if (updatedUuid?.ownerDeleted) {
+        this.onOwnersDeleted.emit(updatedUuid.deletedOwnerId);
+      }
+      else if (updatedUuid) {
         this.onOwnersUpdated.emit();
```
dhaselhan, line 100: onOwnersUpdated told the parent component to re-query the owners from the backend. Is there a reason this is not sufficient at updating the table?

portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
```diff
@@ -92,13 +93,17 @@ export class ParcelOwnersComponent {
...
       }
     });
     dialog.afterClosed().subscribe((result) => {
       if (result.ownerDeleted) {
-        this.dataSource.data = this.dataSource.data.filter((owner: ApplicationOwnerDto | NoticeOfIntentOwnerDto) => owner.uuid !== result.deletedOwnerId);
+        const updatedOwners = this.dataSource.data.filter((o: ApplicationOwnerDto | NoticeOfIntentOwnerDto) => o.uuid !== result.deletedOwnerId);
```
dhaselhan, line 105: It's not great practice to mimic what the API call is doing in the front end, as this way if the backend were to fail or not delete the owner, it would appear in the front end as though it worked. Can you re-load the owners after the delete instead?

portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
urmikgov: i just kept afterclosed

portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
urmikgov: removed it, kept the logic as is

portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
```diff
@@ -91,13 +91,11 @@ export class ParcelOwnersComponent {
...
     }
-    dialog.beforeClosed().subscribe((updatedUuid) => {
-      if (updatedUuid) {
+    dialog.afterClosed().subscribe((result) => {
+      if (result) {
         this.onOwnersUpdated.emit();
       }
-    });
-    dialog.afterClosed().subscribe((result) => {
-      if (result.ownerDeleted) {
```
dhaselhan, line 100: Do we still need this fix and special logic for delete? I thought the above change would fix it.

portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
dhaselhan: perfect!

portal-frontend/src/app/shared/owner-dialogs/parcel-owners/parcel-owners.component.ts
urmikgov: no, removed it

dhaselhan: Can you also clean up
`this.dialogRef.close({ ownerDeleted: true, deletedOwnerId: this.existingUuid});`
Someone might see that logic, and assume the code to handle it was accidentally deleted and see it as a bug, rather than code that isn't used.

urmikgov: yeah, makes sense, removed that logic too

dhaselhan (approved):
LGTM 👍

PR #1210, opened 2023-12-06
public search srw fix

urmikgov:


services/apps/alcs/src/portal/public/search/public-search.controller.ts
```diff
@@ -143,8 +143,9 @@ export class PublicSearchController {
       !searchDto.dateDecidedFrom &&
       !searchDto.dateDecidedTo &&
       !searchDto.decisionMakerCode &&
-      !searchDto.decisionOutcome &&
+      !(searchDto.decisionOutcome && searchDto.decisionOutcome.length > 0) &&
       !isStringSetAndNotEmpty(searchDto.civicAddress);
+    console.log(searchDto.decisionOutcome)
```
trslater, line 148: question: did you mean to leave this log here?

urmikgov: ooops, removing

PR #1244, opened 2023-12-20
Update Primary Contact step

trslater:


portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -40,6 +44,8 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
   showVirusError = false;
   hasCrownParcels = false;
 
+  primaryContactType = new FormControl<boolean | null>(null, [Validators.required]);
```
sandratoh, line 47: The usage and naming of this variable are quite confusing.

- Besides initially setting its value in the component file, `primaryContactType` is only used in the html file, and it was confusing to see its inverse used in (`(change)="onSelectPrimaryContactType(!primaryContactType.value)"`) when the method `onSelectPrimaryContactType` has `selectedThirdPartyAgent` as its parameter
- Seeing the name `primaryContactType`. I was expecting it to be the actual type of primary contact, like individual landowner, government staff, or third party agent. Instead, it was a boolean that tracks whether or not an existing owner from the owner list is the primary contact. That is basically the same thing as the opposite of `selectedThirdPartyAgent`

All of the above makes it seem like `primaryContactType` is a replica of `selectedThirdPartyAgent`, so is there a reason why you made a new variable instead of using the existing variable?

portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -275,4 +287,74 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
...
+    if (selectedThirdPartyAgent) {
+      this.onSelectAgent();
+    } else if (this.isGovernmentUser) {
+      this.onSelectGovernment();
+      if (this.parcelOwners.length == 1) {
+        this.onSelectOwner(this.parcelOwners[0].uuid);
+      }
+    } else {
+      this.selectedThirdPartyAgent = false;
+      if (this.parcelOwners.length == 1) {
```
sandratoh, line 321: Same thing as above re: strict equality

portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -275,4 +287,74 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
...
+  }
+
+  switchPrimaryContactType(selectedThirdPartyAgent: boolean | null) {
+    if (selectedThirdPartyAgent) {
+      this.onSelectAgent();
+    } else if (this.isGovernmentUser) {
+      this.onSelectGovernment();
+      if (this.parcelOwners.length == 1) {
+        this.onSelectOwner(this.parcelOwners[0].uuid);
+      }
```
sandratoh, line 318: Use strict equality `===`. Also, you might want to check the logic here, because if a government user is signed in and the parcel has only one owner, it will run everything in this block, which is basically running `onSelectOwner` twice

sandratoh (changes requested):
Questions also apply to NOI

portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.ts
sandratoh: Suggestion:
* Rename`primaryContactType` to something that better reflect boolean type
* Update/flip related methods' logics and parameters so implementation is more readable

portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -275,4 +287,70 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
...
+  switchPrimaryContactType(isExistingOwner: boolean | null) {
+    if (!isExistingOwner) {
+      this.onSelectAgent();
+    } else if (this.parcelOwners.length == 1) {
+      this.onSelectOwner(this.parcelOwners[0].uuid);
+    } else if (this.isGovernmentUser) {
+      this.onSelectGovernment();
+    } else {
+      this.selectedThirdPartyAgent = false;
+    }
```
sandratoh, line 320: If a government user is submitting an applicant on behalf of a single parcel owner and they select 'Yes' to the toggle question, based on this logic, the parcel owner will be the primary contact instead of the expected result of the government employee.

Also use strict equality `===` instead

portal-frontend/src/app/features/notice-of-intents/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -278,4 +289,70 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
...
+    }
+  }
+
+  private switchPrimaryContactType(isExistingOwner: boolean | null) {
+    if (!isExistingOwner) {
+      this.onSelectAgent();
+    } else if (this.parcelOwners.length == 1) {
+      this.onSelectOwner(this.parcelOwners[0].uuid);
+    } else if (this.isGovernmentUser) {
+      this.onSelectGovernment();
```
sandratoh, line 319: Same as application

portal-frontend/src/app/features/notice-of-intents/edit-submission/primary-contact/primary-contact.component.ts
```diff
@@ -238,6 +247,8 @@ export class PrimaryContactComponent extends FilesStepComponent implements OnIni
         this.selectedLocalGovernment = selectedOwner.type.code === OWNER_TYPE.GOVERNMENT;
       }
 
+      this.isExistingOwner.setValue(!this.selectedThirdPartyAgent);
+
```
sandratoh, line 251: Same as application

sandratoh (approved):
LGTM 👍

portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.html
```diff
@@ -1,67 +1,96 @@
...
+          (change)="onSelectPrimaryContactType(isExistingOwner.value)"
+        >
+          <mat-button-toggle [value]="true">Yes</mat-button-toggle>
+          <mat-button-toggle [value]="false">No</mat-button-toggle>
+        </mat-button-toggle-group>
+      </div>
+
+      <!-- Owner -->
+      <div *ngIf="selectedThirdPartyAgent === false && parcelOwners.length > 0">
+        <label *ngIf="owners.length"
```
mhuseinov, line 37: Should this be parcelOwners.length rather than owners.length since the ngFor bellow refers to parcelOwners?

trslater: Yes! Good catch. This was initially copied from step 1 where we want all owners.

mhuseinov (approved):

PR #1245, opened 2023-12-20
Init App Decisions

lstod:


bin/migrate-oats-data/applications/decisions/application_decisions_init.py
```diff
@@ -0,0 +1,202 @@
...
+            "oats_alr_appl_decision_id": row.get("alr_appl_decision_id"),
+        }
+        data_list.append(mapped_row)
+
+    return data_list
+
+
+def _map_outcome_code(row):
+    if row.get("rescinded_date", None):
+        return "RESC"
```
mhuseinov, line 169: there is a AlcsDecisionOutcomeCodeEnum in common that could be used for this mapping

bin/migrate-oats-data/applications/decisions/application_decisions_init.py
```diff
@@ -0,0 +1,202 @@
...
+    elif row.get("outright_refusal_ind", None) == "N":
+        return "APPR"
+    elif row.get("outright_refusal_ind", None) == "Y":
+        return "REFU"
+    return None
+
+
+def _map_resolution_number_and_year(row):
+    year_and_number = str(row["resolution_number"])
+    # some data appears to only have a year, if this is the case the number will be set to 0
```
mhuseinov, line 179: Could you make a note in google doc and in jira regarding this

bin/migrate-oats-data/applications/decisions/application_decisions_init.py
```diff
@@ -0,0 +1,202 @@
...
+def _map_is_subject_to_conditions(row):
+    return row.get("is_subject_to_conditions", "False") == True
+
+
+@inject_conn_pool
+def clean_application_decisions(conn=None):
+    logger.info("Start application_decisions cleaning")
+    with conn.cursor() as cursor:
+        cursor.execute(
+            f"DELETE FROM alcs.application_decision WHERE audit_created_by = '{OATS_ETL_USER}' AND audit_updated_by is NULL"
```
mhuseinov, line 198: So I've been wondering about this. What is the purpose of checking the audit_updated_by is NULL if we are not going to actually clean anything on production, and this is mostly for our local development? Is it used to avoid data that was manually changed in local db?

mhuseinov (approved):

bin/migrate-oats-data/applications/decisions/application_decisions_init.py
lstod: That was the original idea, to avoid deleting manually changed data as it can cause FK constraints in certain circumstances. I am not sure it is critical, we could probably look into removing it.

mhuseinov: Nope, all good, just wanted to clarify if it is related somehow to prod data or a result of local development. I think it is a good idea to have it to avoid random errors on local

PR #1249, opened 2023-12-21
Feature/alcs 1364

mhuseinov:
- implement ELT for notice_of_intent_decision_component
- added oatsAlrApplComponentId to notice_of_intent_decision_component
- removed partial index on notice_of_intent_decision_component to allow multiple components of the same type on back-end
- move ag_cap and ag_cap_source mapping into common
- fix material description mapping for noi

### DIFF
--- a/bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_init.py
+++ b/bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_init.py
@@ -3,6 +3,8 @@ from common import (
     setup_and_get_logger,
     add_timezone_and_keep_date_part,
     ALRChangeCode,
+    OatsToAlcsAgCap,
+    OatsToAlcsAgCapSource,
 )
 from db import inject_conn_pool
 from psycopg2.extras import RealDictCursor, execute_batch
@@ -139,10 +141,16 @@ def _prepare_oats_alr_applications_data(row_data_list):
             ALRChangeCode.EXT.value,
         ]:
             mapped_row = {
-                "ag_cap": row.get("agri_capability_code"),
+                "ag_cap": str(OatsToAlcsAgCap[row["agri_capability_code"]].value)
+                if row.get("agri_capability_code")
+                else None,
                 "ag_cap_consultant": row.get("agri_cap_consultant"),
                 "ag_cap_map": row.get("agri_cap_map"),
-                "ag_cap_source": row.get("capability_source_code"),
+                "ag_cap_source": str(
+                    OatsToAlcsAgCapSource[row["capability_source_code"]].value
+                )
+                if row.get("capability_source_code")
+                else None,
                 "alr_area": row.get("component_area"),
                 "audit_created_by": OATS_ETL_USER,
                 "end_date": add_timezone_and_keep_date_part(
--- a/bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_init.py
+++ b/bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_init.py
@@ -1,0 +1,191 @@
+from common import (
+    OATS_ETL_USER,
+    setup_and_get_logger,
+    add_timezone_and_keep_date_part,
+    ALRChangeCode,
+)
+from db import inject_conn_pool
+from psycopg2.extras import RealDictCursor, execute_batch
+
+etl_name = "init_notice_of_intent_decision_components"
+logger = setup_and_get_logger(etl_name)
+
+
+@inject_conn_pool
+def init_notice_of_intent_decision_components(conn=None, batch_size=1000):
+    """
+    This function is responsible for initializing the notice_of_intent_decision_components in ALCS.
+
+    Args:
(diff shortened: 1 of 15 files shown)


bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_soil_details_update.py
```diff
@@ -0,0 +1,146 @@
...
+            data_list[SoilChangeCode.ADD.value].append(data)
+        elif data["soil_change_code"] == SoilChangeCode.REMOVE.value:
+            data_list[SoilChangeCode.REMOVE.value].append(data)
+
+    return data_list
+
+
+def map_soil_fields(data):
+    soil_type = data.get("material_desc", "")
+    soil_origin = data.get("material_desc", "")
```
trslater, line 135: Won't these two variables have the exact same value?

mhuseinov: good catch, that is a leftover from soil mapping of noi component

bin/migrate-oats-data/noi/notice_of_intent_submissions/notice_of_intent_soil_fields.py
```diff
@@ -149,7 +149,7 @@ def _prepare_oats_alr_applications_data(row_data_list):
 
 
 def map_soil_fields(data):
-    soil_type = data.get("material_description", "")
+    soil_type = data.get("material_desc", "")
```
trslater, line 152: This seems like a duplicate of the function from [.../noi_decisions/noi_components/notice_of_intent_decision_components_soil_details_update.py](https://github.com/bcgov/alcs/pull/1249/files#diff-0f292c4c25c2b8b0ca95bee3d7efa19b6f6d5faafb57a269846275285bebd3ca)

bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_init.py
```diff
@@ -0,0 +1,199 @@
...
+            }
+            data_list.append(mapped_row)
+
+    return data_list
+
+
+def _map_component_type_code(row):
+    alr_change_code = row.get("alr_change_code")
+    if alr_change_code == ALRChangeCode.SCH.value:
+        return "PFRS"
```
lstod, line 179: Future note to change this to enum when doing applications

lstod (approved):

bin/migrate-oats-data/noi/notice_of_intent_submissions/notice_of_intent_soil_fields.py
mhuseinov: correct, since the application and decision components in OATS are the same but different in ALCS. This function is intentionally duplicated since they are responsible for different ALCS tables. The only piece that is currently shared is Enum since it has constraints on DB level and it is safe to reuse that code

lstod: Should basically be a carbon copy but with different join and insert

bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_init.py
mhuseinov: agree for applications since there will be more than 3 values and it will refer to different table in DB

bin/migrate-oats-data/noi/noi_decisions/noi_components/notice_of_intent_decision_components_init.py
```diff
@@ -0,0 +1,199 @@
...
+            "noi/sql/notice_of_intent_decisions/components/notice_of_intent_decision_components_insert.sql",
+            "r",
+            encoding="utf-8",
+        ) as sql_file:
+            query = sql_file.read()
+
+            while True:
+                cursor.execute(
+                    f"""{query} 
+                    WHERE oaac.alr_appl_component_id > {last_decision_component_id} ORDER BY oaac.alr_appl_component_id;"""
```
trslater, line 56: I pointed this out to @lstod and I can't remember what we decided, but doesn't this do a lot of duplicate work? It first selects all rows, but only takes the first batch, then reselects all the rows from the end of that batch, etc. Would it make sense to use a limit clause or add an upper bound of `last_decision_component_id + batch_size`?

mhuseinov: it does not actually read the queries from db, rows = cursor.fetchmany(batch_size) bellow it does the retrieving from DB

trslater: My concern is that it needs to run the whole query on the database first, which does fetch all of the rows, but then only imports the batch size into Python. This isn't really the place for this discussion though (I probably shouldn't have brought it up), so I'm just going to mark it resolved. It might be worth looking into as a potential optimization.

trslater (approved):
LGTM 🎉

PR #1253, opened 2023-12-27
Improvements for notifications buttons, uploader, review, and submit confirmation

trslater:
- Scroll to owner info on owner select (parcel step)
  - Missed from ALCS-1270
- Make save and step buttons sticky
- Remove redundant max file size notices
- Remove space to align with others
- Only show edit button when parcels > 1
- Show units for total area of SR
- Expand submit confirmation text
- Add contact link to submit confirmation
- Fix submit confirmation mobile button layout

### DIFF
--- a/portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
+++ b/portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
@@ -306,8 +306,9 @@
       <div>
         <mat-radio-group
           color="primary"
-          [value]="selectedOwner?.uuid" required
-          [ngClass]="{ 'error': ownerInput.errors && ownerInput.errors['required'] }"
+          [value]="selectedOwner?.uuid"
+          required
+          [ngClass]="{ error: ownerInput.errors && ownerInput.errors['required'] }"
           formControlName="selectedOwnerVal"
         >
           <mat-radio-button
@@ -321,7 +322,7 @@
           </mat-radio-button>
         </mat-radio-group>
 
-        <div class="warning-banner" *ngIf="selectedOwner">
+        <div id="owner-info" class="warning-banner" *ngIf="selectedOwner">
           <div class="owner-details">
             <div class="row">
               <div class="key"><strong>First Name:</strong></div>
@@ -331,8 +332,8 @@
               <div class="key"><strong>Last Name:</strong></div>
               <div class="value">{{ selectedOwner.lastName }}</div>
             </div>
-            <div class="row"> 
-              <div class="key"><strong>Ministry or Department:</strong> </div>
+            <div class="row">
+              <div class="key"><strong>Ministry or Department:</strong></div>
               <div class="value">{{ selectedOwner.organizationName }}</div>
             </div>
             <div class="row">
--- a/portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
+++ b/portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
@@ -315,14 +315,14 @@
             *ngFor="let owner of filteredOwners"
             [value]="owner.uuid"
             class="radio-option"
-            (click)="onCrownOwnerSelected($event, owner, owner.isSelected)"
+            (click)="onCrownOwnerSelected(owner, owner.isSelected)"
             required
           >
             {{ owner.firstName + ' ' + owner.lastName }}
           </mat-radio-button>
         </mat-radio-group>
 
-        <div id="owner-info" class="warning-banner" *ngIf="selectedOwner">
+        <div #ownerInfo class="warning-banner" *ngIf="selectedOwner">
           <div class="owner-details">
             <div class="row">
               <div class="key"><strong>First Name:</strong></div>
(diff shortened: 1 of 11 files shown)


portal-frontend/src/app/features/notifications/edit-submission/edit-submission.component.scss
urmikgov, line 1: This needs to be done on both applications and nois.

portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
```diff
@@ -321,7 +322,7 @@ <h6 class="owner-search-option" *ngIf="filteredOwners.length">OR</h6>
           </mat-radio-button>
         </mat-radio-group>
 
-        <div class="warning-banner" *ngIf="selectedOwner">
+        <div id="owner-info" class="warning-banner" *ngIf="selectedOwner">
```
dhaselhan, line 325: Since this component can exist multiple times per page, you cannot use an ID. IDs must be unique in the DOM.

portal-frontend/src/app/features/notifications/notification-details/parcel/parcel.component.html
```diff
@@ -5,7 +5,7 @@ <h3 class="flex-item">1. Identify Parcel(s) Under The Statutory Right of Way</h3
       <h4>Parcel {{ parcelInd + 1 }}</h4>
       <div>
         <button
-          *ngIf="showEdit"
+          *ngIf="showEdit && parcels.length > 1"
```
dhaselhan, line 8: This is inside a loop of parcels, so this code should only execute if there is at least one parcel. What does this fix?

portal-frontend/src/app/features/notifications/notification-details/parcel/parcel.component.html
```diff
@@ -14,7 +14,7 @@ <h4>Parcel {{ parcelInd + 1 }}</h4>
           Edit Parcel {{ parcelInd + 1 }}
         </button>
         <button
-          *ngIf="showEdit"
+          *ngIf="showEdit && parcels.length > 1"
```
dhaselhan, line 17: Same as above.

portal-frontend/src/app/features/notifications/notification-details/proposal-details/proposal-details.component.html
```diff
@@ -13,7 +13,7 @@
 
   <div class="subheading2 grid-1">Total area of the SRW</div>
   <div class="grid-double">
-    {{ _notificationSubmission.totalArea }}
+    {{ _notificationSubmission.totalArea }} <ng-container *ngIf="_notificationSubmission.totalArea">ha</ng-container>
```
dhaselhan, line 16: nitpick: Use a span just to be consistent with other places we do this, not a big deal since the end functionality works basically the same.

portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
trslater: Good call. Technically, only one parcel is expanded at a time and the inner HTML isn't rendered, but it's still not great. What do you think about adding the parcel number, e.g., "owner-info-1"?

portal-frontend/src/app/features/notifications/notification-details/parcel/parcel.component.html
trslater: It's just one of the requirements:
> only show Edit Parcel button when there is more than one parcel

portal-frontend/src/app/features/notifications/notification-details/proposal-details/proposal-details.component.html
trslater: I suppose for consistency this makes sense, but we end up with a pointless span then.

portal-frontend/src/app/features/notifications/notification-details/parcel/parcel.component.html
dhaselhan: This code wouldn't help with that requirement since it shouldn't render this section if there are no parcels. You should not add code that does not do anything.

dhaselhan: Ohhh, I read that as >  0, I see what you are saying.

portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
dhaselhan: I would use an Angular mechanism for getting the element instead of plain JS.
Check this post for a good example: https://stackoverflow.com/a/54418622

portal-frontend/src/app/features/notifications/notification-details/proposal-details/proposal-details.component.html
dhaselhan: And that's why its a nitpick, which one is the lesser of two evils? 🙈

portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
trslater: That's way better. I was trying to avoid re-adding existing functionality, so reused `scrollToElement` from `portal-frontend/src/app/shared/utils/scroll-helper.ts`.

dhaselhan (approved):
LGTM 👍

urmikgov (approved):

urmikgov:
approved

PR #1262, opened 2024-01-04
Add hideFromPortal Flag

dhaselhan:
* Remove hideFromPortal = false from inbox and public search results
* Prevent public from loading at all
* Add UI to App/NOI detail page to change toggle


services/apps/alcs/src/portal/inbox/notice-of-intent/inbox-notice-of-intent-view.entity.ts
```diff
@@ -38,7 +38,11 @@ import { LinkedStatusType } from '../inbox.dto';
         'status',
       )
       .from(NoticeOfIntentSubmission, 'noi_sub')
-      .innerJoin(NoticeOfIntent, 'noi', 'noi.file_number = noi_sub.file_number')
+      .innerJoin(
+        NoticeOfIntent,
+        'noi',
+        'noi.file_number = noi_sub.file_number AND noi.hide_from_portal = false',
```
sandratoh, line 44: Nitpick: I liked the gigantic `FALSE` you had in application, more readable

services/apps/alcs/src/providers/typeorm/migrations/1704396938830-refresh_views_for_hide_from_public.ts
```diff
@@ -0,0 +1,165 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class RefreshViewsForHideFromPublic1704396938830
+  implements MigrationInterface
+{
+  name = 'RefreshViewsForHideFromPublic1704396938830';
+
```
sandratoh, line 7: What is going on? 😵

alcs-frontend/src/app/features/notice-of-intent/overview/overview.component.ts
```diff
@@ -108,4 +110,34 @@ export class OverviewComponent implements OnInit, OnDestroy {
...
+          this.isHiddenFromPortal = !didConfirm;
+          if (didConfirm && this.noticeOfIntent) {
+            this.noticeOfIntentDetailService.update(this.noticeOfIntent.fileNumber, {
+              hideFromPortal: false,
+            });
+          }
+        });
+    }
+  }
 }
```
sandratoh, line 143: How come we're updating to `hideFromPortal: true` when we want to remove the restriction, and it is the opposite when we want to add the restriction? Wouldn't we want it to be `hideFromPortal` to be false when we want to be able to view it?

services/apps/alcs/src/providers/typeorm/migrations/1704396938830-refresh_views_for_hide_from_public.ts
dhaselhan: What do you mean? The name of the migration is RefreshViewsForHideFromPublic and the number after is the timestamp.

alcs-frontend/src/app/features/notice-of-intent/overview/overview.component.ts
dhaselhan: Ah good catch! I think I got the messages backwards but the logic is correct.

services/apps/alcs/src/portal/inbox/notice-of-intent/inbox-notice-of-intent-view.entity.ts
dhaselhan: Going to ignore this as i don't want to regenerate and update the migrations for a nitpick.

alcs-frontend/src/app/features/notice-of-intent/overview/overview.component.ts
dhaselhan: Updated! The logic might look backwards in the if/else, but the value changes immediately before the onChange is fired.

services/apps/alcs/src/providers/typeorm/migrations/1704396938830-refresh_views_for_hide_from_public.ts
sandratoh: I mean how come we need to do a big refresh? Does it happen periodically with the db?

alcs-frontend/src/app/features/notice-of-intent/overview/overview.component.ts
sandratoh: Makes sense! Application looks good now, but NOI still needs update

services/apps/alcs/src/providers/typeorm/migrations/1704396938830-refresh_views_for_hide_from_public.ts
dhaselhan: sorry, maybe I should use the word re-create, since i changed the query I need to drop and re-create the view which I called refreshing.

alcs-frontend/src/app/features/notice-of-intent/overview/overview.component.ts
dhaselhan: NOI updated

sandratoh (approved):
LGTM 👍

services/apps/alcs/src/main.ts
```diff
@@ -60,6 +60,7 @@ const registerGlobalFilters = (app: NestFastifyApplication) => {
 
 const registerCors = (app: NestFastifyApplication) => {
   app.enableCors({
+    maxAge: 3600,
```
mhuseinov, line 63: nice!

mhuseinov (approved):

PR #1263, opened 2024-01-04
Update application proposals, reviews, and dialog

sandratoh:
- Create shared `commission-purposes-panel` component
- Add new shared panel component to soil type application purposes
- Replace panel component usage in noi purposes and clean up repeated code
- Add proposal units to Step 8/View/Public View/ALCS
- Update submit confirmation dialog text and styling

### DIFF
--- a/alcs-frontend/src/app/features/application/applicant-info/application-details/subd-details/subd-details.component.html
+++ b/alcs-frontend/src/app/features/application/applicant-info/application-details/subd-details/subd-details.component.html
@@ -11,7 +11,7 @@
       <div class="grid-2">
         {{ lot.type }}
       </div>
-      <div class="grid-3">{{ lot.size }} <span *ngIf="lot.size">ha</span></div>
+      <div class="grid-3">{{ lot.size }} <span *ngIf="lot.size !== null">ha</span></div>
     </ng-container>
   </div>
   <div class="subheading2 grid-1">What is the purpose of the proposal?</div>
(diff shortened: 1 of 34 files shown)


alcs-frontend/src/app/features/application/applicant-info/application-details/subd-details/subd-details.component.html
```diff
@@ -11,9 +11,7 @@
       <div class="grid-2">
         {{ lot.type }}
       </div>
-      <div class="grid-3">
-        {{ lot.size }}
-      </div>
+      <div class="grid-3">{{ lot.size }} <span *ngIf="lot.size">ha</span></div>
```
dhaselhan, line 14: Lot.size is a number, so this will return false for 0. Is that okay or should this be lot.size !== null?

portal-frontend/src/app/shared/commission-purposes-panel/commission-purposes-panel.component.html
```diff
@@ -0,0 +1,31 @@
+<mat-expansion-panel #panel [expanded]="true">
```
dhaselhan, line 1: Nice work making this its own component!

dhaselhan (approved):
LGTM 👍

alcs-frontend/src/app/features/application/applicant-info/application-details/subd-details/subd-details.component.html
sandratoh: Good point. The backend doesn't actually save the input if user puts a 0 right now, but I will update it as a safeguard.

sandratoh: Updated!

mhuseinov (approved):
👍

PR #1266, opened 2024-01-04
update elt to support 1 project duration field instead of 2

mhuseinov:
- update application related etls to map one project duration field instead of two
- update notice of intent related etls to map one project duration field instead of two


lstod (approved):
LGTM(ok actually dev)!

dhaselhan (approved):
LGTM 👍

PR #1274, opened 2024-01-08
Feature/alcs application reconsideration etl

mhuseinov:
- application reconsideration ETL
- update application modification ETL (no need to populate new or link modifications)
- add coalesce where missing for modifications 
- update ALCS API to support nullable cards on reconsideration


trslater (approved):
LGTM 👍

services/apps/alcs/src/alcs/admin/unarchive-card/unarchive-card.service.ts
```diff
@@ -123,7 +123,7 @@ export class UnarchiveCardService {
       await this.reconsiderationService.getDeletedCards(fileId);
     for (const reconsideration of reconsiderations) {
       result.push({
-        cardUuid: reconsideration.cardUuid,
+        cardUuid: reconsideration.cardUuid ?? '',
```
dhaselhan, line 126: Should this be ?? null? Won't empty string throw an exception since its not UUID or is that intentional?

dhaselhan (approved):
LGTM 👍

services/apps/alcs/src/alcs/admin/unarchive-card/unarchive-card.service.ts
mhuseinov: nope, since this is DTO. The cardId later could be sent as empty to api on delete action, but the API has the check implemented there. I'll look into setting the cardUuid to null and see how it affects UI.

dhaselhan: Chatted with Mekhti and cleared this up, I was confused on which direction we were mapping. Good to go.

PR #1275, opened 2024-01-08
Feature/alcs 1330 etl hide from portal

mhuseinov:
- set hide_from_portal on notice of intent
- set hide_from_portal on application
- move users ETL into separate folder


bin/migrate-oats-data/noi/set_hide_from_portal_on_noi.py
```diff
@@ -0,0 +1,48 @@
...
+            cursor.execute(count_query)
+            count_total = dict(cursor.fetchone())["count"]
+            logger.info(f"Total application to set hide_from_portal: {count_total}")
+
+            update_query = f"""
+                                UPDATE alcs.notice_of_intent AS noi
+                                SET hide_from_portal = TRUE
+                                FROM oats.oats_alr_applications AS oaa
+                                WHERE noi.file_number = oaa.alr_application_id::TEXT
+                                    AND oaa.who_created <> 'PROXY_OATS_APPLICANT';
```
dhaselhan, line 36: Is there a reason you do a loop here for each noi? Wouldn't it be more efficient to execute this once without the file_number in the where clause to set all of them?

dhaselhan (approved):
LGTM 👍

trslater (approved):
LGTM 👍

bin/migrate-oats-data/noi/set_hide_from_portal_on_noi.py
mhuseinov: this is a join of ALCS with OATS data. Since we need to set who created  in ALCS only for 'PROXY_OATS_APPLICANT' that is stored in OATS

PR #1276, opened 2024-01-09
Fixes for Portal Maintenance Mode

dhaselhan:
* Add new route for auth controller so that it triggers maintenance guard for portal
* Add logic to prevent deactivate from infinite looping when navigating to the maintenance page. Save fails triggering a redirect which triggers canDeactivate which calls save again.

### DIFF
--- a/portal-frontend/src/app/features/applications/alcs-edit-submission/alcs-edit-submission.component.ts
+++ b/portal-frontend/src/app/features/applications/alcs-edit-submission/alcs-edit-submission.component.ts
@@ -50,6 +50,7 @@ export class AlcsEditSubmissionComponent implements OnInit, OnDestroy, AfterView
   expandedParcelUuid?: string;
 
   showValidationErrors = false;
+  isDeactivating = false;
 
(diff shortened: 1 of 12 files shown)


portal-frontend/src/app/features/applications/alcs-edit-submission/alcs-edit-submission.component.ts
```diff
@@ -143,7 +146,16 @@ export class AlcsEditSubmissionComponent implements OnInit, OnDestroy, AfterView
 
   // this gets fired whenever applicant navigates away from edit page
   async canDeactivate(): Promise<Observable<boolean>> {
-    await this.saveApplication(this.customStepper.selectedIndex);
+    if (!this.isDeactivating) {
```
mhuseinov, line 149: nitpick: in theory, it should be possible to create a helper function or base component that will have canDiactivate implemented with the try-catch logic and accept save method from the specific component.

mhuseinov (approved):
👍

trslater (approved):
LGTM 😎

PR #1279, opened 2024-01-09
Change logic of disabling / showing hiding fields for App & NOI

dhaselhan:
* Disable all fields on step 1 until they select Fee Simple or Crown
* Remove disabled logic from parcel entry as its not used
* Hide file upload on Step 8 for SUBD if not required
* Disable soil and fill questions on NFU and NARU unless they answer Yes

### DIFF
--- a/portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
+++ b/portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
@@ -220,6 +220,7 @@
       [showErrors]="showErrors"
       [isRequired]="isCertificateOfTitleRequired"
       [showVirusError]="showVirusError"
+      [disabled]="parcelForm.controls.pid.disabled"
     ></app-file-drag-drop>
   </div>
 
@@ -307,9 +308,8 @@
         <mat-radio-group
           color="primary"
           [value]="selectedOwner?.uuid"
-          required
+          [required]="true"
           [ngClass]="{ error: ownerInput.errors && ownerInput.errors['required'] }"
-          formControlName="selectedOwnerVal"
         >
           <mat-radio-button
             *ngFor="let owner of filteredOwners"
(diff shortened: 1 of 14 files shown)


portal-frontend/src/app/features/applications/edit-submission/proposal/subd-proposal/subd-proposal.component.html
```diff
@@ -264,9 +264,9 @@ <h6>Documents needed for this step:</h6>
         (deleteFile)="onDeleteFile($event)"
         (openFile)="openFile($event)"
         [showErrors]="showErrors"
-        [isRequired]="isHomeSiteSeverance.getRawValue() !== 'false'"
+        [isRequired]="isHomeSiteSeverance.getRawValue() !== 'true'"
         [allowMultiple]="true"
-        [disabled]="isHomeSiteSeverance.getRawValue() === 'false'"
+        [disabled]="isHomeSiteSeverance.getRawValue() !== 'true'"
```
mhuseinov, line 269: just curious what is the reason behind changing from === 'false' to !== 'true'

mhuseinov (approved):
👍

portal-frontend/src/app/features/applications/edit-submission/proposal/subd-proposal/subd-proposal.component.html
dhaselhan: The original requirements had the form field disabled if they chose false, now its disabled if they haven't selected anything or its false. So this check now covers null.

portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
```diff
@@ -307,9 +308,8 @@ <h6 class="owner-search-option" *ngIf="filteredOwners.length">OR</h6>
         <mat-radio-group
           color="primary"
           [value]="selectedOwner?.uuid"
-          required
+          [required]="true"
```
trslater, line 311: This will set `required` on the `mat-radio-group` to the boolean `true` rather than a string that says "true", correct? What is the purpose of doing it this way (seems like it is not most other places)?

trslater (approved):
LGTM 👍

portal-frontend/src/app/features/applications/edit-submission/parcel-details/parcel-entry/parcel-entry.component.html
dhaselhan: Having just required without a value was showing a red error on my IDE saying that a value was required. I'm not even sure it makes a difference.

PR #1283, opened 2024-01-10
Second round of QA fixes for primary contact

trslater:
- Radio buttons now outline in red
- Fixed various spacing issues
- Single owner now selects automatically (NOI's only)
- Removed error showing up when government contact selected
- Previously added owners no longer showing up as primary contact option when government contact selected
- Previously saved NOI 3rd-party contact now selected when returning to primary contact page
- Government contact form now showing on first load
- Owner list error no longer shows up even when something is selected


dhaselhan (approved):
LGTM 👍

portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.html
```diff
@@ -1,92 +1,116 @@
...
+<section>
+  <div class="step-description">
+    <h2>Primary Contact</h2>
+    <div>
+      <p>
+        The primary contact becomes the single point of contact and must ensure that landowners are notified of all
+        information exchanged between the primary contact, local or First Nation government, and the ALC.
+      </p>
+    </div>
+    <div><p>*All fields are required unless stated optional or disabled.</p></div>
```
sandratoh, line 10: Nitpick: What is the reason for all the wrapping divs? Same for NOI

sandratoh (approved):
Nitpick comment for HTML files, but LGTM 👍

portal-frontend/src/app/features/applications/edit-submission/primary-contact/primary-contact.component.html
trslater: You are correct. They are useless. I added these, because originally these paragraphs were pushed up against each other. I was trying to understand the layout system, so I looked around at other pages. One of them had the paragraphs wrapped in divs and the paragraphs had margins, so I thought it might be a scoping issue, but it turned out these were added in a custom CSS rule for that page. My issue was that the entire step-description contents were wrapped in a div, which was preventing the flex gap. I just forgot to take those divs out.

PR #1284, opened 2024-01-10
Prevent creating v1 cards

dhaselhan:
* Remove App and NOI creation from all boards
* Hide in Admin UI for Create

### DIFF
--- a/alcs-frontend/src/app/features/admin/board-management/board-management-dialog/board-management-dialog.component.ts
+++ b/alcs-frontend/src/app/features/admin/board-management/board-management-dialog/board-management-dialog.component.ts
@@ -11,7 +11,7 @@ import { CardStatusService } from '../../../../services/card/card-status/card-st
 import { CardType } from '../../../../shared/card/card.component';
 import { BaseCodeDto } from '../../../../shared/dto/base.dto';
 
-const DISABLED_CREATE_CARD_TYPES = [CardType.APP, CardType.COV, CardType.NOI];
+const DISABLED_CREATE_CARD_TYPES = [CardType.APP, CardType.COV, CardType.NOI, CardType.NOTIFICATION];
 
 @Component({
   selector: 'app-decision-condition-types-dialog',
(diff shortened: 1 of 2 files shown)


alcs-frontend/src/app/features/admin/board-management/board-management-dialog/board-management-dialog.component.ts
```diff
@@ -98,7 +100,7 @@ export class BoardManagementDialogComponent implements OnInit {
     this.step++;
     //We don't support creating Notifications
     this.allowedCreateCardTypes = (this.permittedCardTypes.value ?? []).filter(
-      (cardType) => cardType !== CardType.NOTIFICATION
+      (cardType) => !DISABLED_CREATE_CARD_TYPES.includes(cardType),
```
trslater, line 103: Won't this now enable `NOTIFICATION` card type? Is that intended?

alcs-frontend/src/app/features/admin/board-management/board-management-dialog/board-management-dialog.component.ts
```diff
@@ -98,7 +100,7 @@ export class BoardManagementDialogComponent implements OnInit {
     this.step++;
     //We don't support creating Notifications
```
sandratoh, line 101: We should probably update this comment too

alcs-frontend/src/app/features/admin/board-management/board-management-dialog/board-management-dialog.component.ts
dhaselhan: It definitely would and not intended. Good catch! MR updated.

alcs-frontend/src/app/features/admin/board-management/board-management-dialog/board-management-dialog.component.ts
dhaselhan: Comment removed.

trslater (approved):
LGTM 👍

PR #1286, opened 2024-01-11
Prevent green success toast from firing when file is too large

dhaselhan:
* Green toast was replacing the yellow warning toast
* Move Cancel NOI into the button container so it looks correct


mhuseinov (approved):
👍

sandratoh (approved):
LGTM 👍

PR #1290, opened 2024-01-15
Fix radio list error still showing when no owners (ALCS-555)

trslater:
Added a conditional container to that whole section, except the no owner warning (NOI and app).


dhaselhan (approved):
LGTM 👍

PR #1293, opened 2024-01-15
standardize sticky button UI

urmikgov:


portal-frontend/src/app/shared/button-container/button-container.component.scss
```diff
@@ -0,0 +1,49 @@
+@use '../../../styles/functions' as *;
+@use '../../../styles/colors';
+
+
+  .button-container {
```
trslater, line 5: Nitpick: probably shouldn't be indented. Auto-format on save should have caught this.

portal-frontend/src/app/shared/button-container/button-container.component.html
```diff
@@ -0,0 +1,15 @@
+<div class="button-container">
+  <button *ngIf="!draftMode" (click)="onSaveExit()" mat-stroked-button color="accent">Save and Exit</button>
```
trslater, line 2: Since these buttons are identical besides text, consider changing to:

```html
<button (click)="onSaveExit()" mat-stroked-button color="accent">{{ draftMode ? 'Discard all Changes' : 'Save and Exit'</button>
```

dhaselhan:
suggestion: Make the sticky bar more re-usable by having it just be a container that positons / styles itself and use content projection. https://angular.io/guide/content-projection

portal-frontend/src/app/shared/button-container/button-container.component.html
urmikgov: gotcha, made changes

urmikgov:
> suggestion: Make the sticky bar more re-usable by having it just be a container that positons / styles itself and use content projection. https://angular.io/guide/content-projection

I could do that for text, any suggestion for click handlers?
thats one of the reason i chose to use this method

dhaselhan:
I'm not sure what you mean, you can have <button (click)> inside content projection, it shouldn't cause any issues.

urmikgov:
@dhaselhan I would prefer this approach more, since its reducing file sizes. content projection means we have to move all this back to individual components. we can use this component just for these 3 buttons. Let's discuss it tomorrow in standup.

trslater (approved):
LGTM 👍

PR #1296, opened 2024-01-16
populate staff journal user

mhuseinov:
- populate staff journal user


dhaselhan (approved):
LGTM 👍

bin/migrate-oats-data/menu/commands/applications.py
```diff
@@ -21,9 +25,11 @@ def application_import(console, args):
         init_applications(batch_size=import_batch_size)
         init_alcs_users(batch_size=import_batch_size)
         process_application_etl(batch_size=import_batch_size)
+        populate_default_staff_journal_user()
```
lstod, line 28: staff journal for applications gets populated within process_application_etl, should that be brought out to this level?

bin/migrate-oats-data/menu/commands/notice_of_intents.py
```diff
@@ -22,9 +26,11 @@ def notice_of_intent_import(console, args):
         init_notice_of_intents(batch_size=import_batch_size)
         init_alcs_users(batch_size=import_batch_size)
         process_notice_of_intent(batch_size=import_batch_size)
+        populate_default_staff_journal_user()
```
lstod, line 29: I believe the above is also the case for NOI

bin/migrate-oats-data/menu/commands/applications.py
mhuseinov: nope, since the application_staff_journal script creates entries only for applications where this script updates all staff journal records (application, noi, etc.). Also, script designed in a way that it is possible to run it without clearing any data multiple times and nothing will change in output result

PR #1305, opened 2024-01-17
ETL Application_decision_component_lots

lstod:


bin/migrate-oats-data/applications/decisions/app_components/lot/application_decision_component_lot_insert.py
```diff
@@ -0,0 +1,143 @@
...
+            "r",
+            encoding="utf-8",
+        ) as sql_file:
+            count_query = sql_file.read()
+            cursor.execute(count_query)
+            count_total = dict(cursor.fetchone())["count"]
+        logger.info(f"Total Application Component Lots data to insert: {count_total}")
+
+        failed_inserts = 0
+        successful_inserts_count = 0
```
dhaselhan, line 36: suggestion: since both are counts they should both have _count. I see that this is copied and pasted a bunch only some of them are fixed, might want to check if theres others.

bin/migrate-oats-data/applications/decisions/app_components/lot/application_decision_component_lot_insert.py
```diff
@@ -0,0 +1,143 @@
...
+            "index": _create_parcel_index(row, component_id_counts),
+            "type": "Lot",
+
+        }
+        data_list.append(mapped_row)
+
+    return data_list
+
+
+def _create_parcel_index(row, component_id_counts):
```
dhaselhan, line 127: question: Why is this called create_parcel_index if it returns a dictionary? I assumed it would create a DB index or similar.

mhuseinov: I think in this case, the index is 1,2,3.... values in index column of lot table, rather than DB index

lstod: That is correct, there is no index in oats, this is for the column "index"

dhaselhan: Ahh that makes sense, forgot the column was called index!

bin/migrate-oats-data/applications/decisions/app_components/lot/application_decision_component_lot_insert.py
lstod: current file has been updaed

dhaselhan (approved):
LGTM 👍

mhuseinov (approved):
👍

PR #1306, opened 2024-01-17
Update NFU Submission Template

dhaselhan:
* More bug fixes!
* Flatten primary contact to keep the template simpler


mhuseinov (approved):
👍

services/apps/alcs/src/portal/pdf-generation/generate-submission-document.service.ts
```diff
@@ -379,4 +384,12 @@ export class GenerateSubmissionDocumentService {
 
     return pdfData;
   }
+
+  private formatPid(pid?: string | null) {
+    const matches = pid?.match(/(.{3})(.{3})(.{3})/);
```
trslater, line 389: Question: I noticed at some point that there are 7-digit PID's in the database. Do we need to match those as well?

dhaselhan: There shouldn't be any, all PIDs are 9 digits. I'll take a look.

trslater (approved):
LGTM 👍

services/apps/alcs/src/portal/pdf-generation/generate-submission-document.service.ts
dhaselhan: Good catch! Update the code to support any length of PIDs.

PR #1310, opened 2024-01-18
update obfuscation, update batch insert, update readme

mhuseinov:
- do not obfuscate the organization name in OATS
- update readme file with backup and restore instructions for Postgres DB
- update batch insert script to correctly handle duplications of applications and NOIs if they are caused by local government duplications
- add filter to application batch insert


dhaselhan (approved):
LGTM 👍

trslater (approved):
LGTM 👍

PR #1316, opened 2024-01-22
Temporarily allow deleted application types to improve search performance

trslater:
- Bypasses soft deletion of application types


dhaselhan (approved):
LGTM 👍

dhaselhan:
I think we will also need to apply this to Inbox / ALCS Search as well depending on how long the real fix takes.

mhuseinov (approved):
👍

PR #1319, opened 2024-01-22
Apply 1328 fix to inbox search as well (ALCS-1459)

trslater:
- Apply same fix from #1316 to inbox search
- Includes a couple auto-format changes


dhaselhan (approved):
LGTM 👍

mhuseinov (approved):
👍

PR #1320, opened 2024-01-23
covenant proposal details on alcs to not show errors

urmikgov:


trslater (approved):
LGTM 👍

dhaselhan:
I would double check which fields are optional and which aren't. If the field is optional and the user has not entered a value, ALCS should show No Data.

urmikgov:
@dhaselhan i just checked and makes sense to have no data. the bug is about showing errors. made changes

dhaselhan (approved):
LGTM 👍

PR #1322, opened 2024-01-23
Feature/alcs 1487

mhuseinov:
- set source column for application and notice of intent
- set hide from portal from application and notice of intent
- add missing LG to mapping


bin/migrate-oats-data/noi/set_notice_of_intent_visibility.py
```diff
@@ -0,0 +1,160 @@
...
+                                UPDATE alcs.notice_of_intent 
+                                SET "source" = 'APPLICANT'
+                                FROM oats.oats_alr_applications oaa 
+                                WHERE alcs.notice_of_intent.file_number = oaa.alr_application_id::TEXT 
+                                AND oaa.who_created = 'PROXY_OATS_APPLICANT';
+                            """
+            cursor.execute(update_created_by_applicant_query)
+            updated_created_by_applicant_count = cursor.rowcount
+
+            update_created_by_oats_query = f"""
```
dhaselhan, line 126: This query and the one above almost look identical, it seems they are updating the same IDs with slightly different source options. How does this work?

lstod: I believe the key here is who_created = PROXY vs <> PROXY

lstod (approved):

bin/migrate-oats-data/noi/set_notice_of_intent_visibility.py
mhuseinov: exactly the one above selects all records created by proxy, where this one everything that is NOT selected by proxy. PROXY_OATS_APPLICANT in oats means that the record was created by user from Oats Portal

dhaselhan (approved):
LGTM 👍

PR #1335, opened 2024-01-25
updated created_at date for applications and nois

lstod:
- set created_at to created_date
- if no created_date, use when_created


mhuseinov (approved):
👍

trslater (approved):
LGTM 👍

PR #1336, opened 2024-01-25
Update Logic for OATS type Apps

dhaselhan:
* Previously we only had ALCS and Applicant, now we have OATS
* We want to show all V2 features for OATS applications


trslater (approved):
LGTM 👍

lstod (approved):
💯

PR #1338, opened 2024-01-25
Hardcode tab order for easier referencing

dhaselhan:
* Requiring the tabs to be rendered to get their order forces the page to default to application since they are not rendered on initial load, meaning it was always going to applications


mhuseinov (approved):
👍

trslater (approved):
LGTM 👍

PR #1340, opened 2024-01-25
Update Primary contact First Name

lstod:
- update owners so primary contact does not show "None" for first name if primary contact is an organization

### DIFF
--- a/bin/migrate-oats-data/applications/submissions/parcels/primary_contact/application_primary_contact.py
+++ b/bin/migrate-oats-data/applications/submissions/parcels/primary_contact/application_primary_contact.py
@@ -124,7 +124,10 @@ def _map_data(row):
 def _get_name(row):
     first_name = row.get("first_name", "")
     middle_name = row.get("middle_name", "")
-    return f"{first_name} {middle_name}".strip()
+    name = f"{first_name} {middle_name}".strip()
+    if row.get("organization_name") and name == "None":
+        return None
+    return name
 
 
 @inject_conn_pool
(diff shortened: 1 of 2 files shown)


dhaselhan (approved):
LGTM 👍

bin/migrate-oats-data/applications/submissions/parcels/primary_contact/application_primary_contact.py
```diff
@@ -124,7 +124,10 @@ def _map_data(row):
 def _get_name(row):
     first_name = row.get("first_name", "")
     middle_name = row.get("middle_name", "")
-    return f"{first_name} {middle_name}".strip()
+    name = f"{first_name} {middle_name}".strip()
+    if row.get("organization_name") and name == "None":
```
mhuseinov, line 128: Do I get it right that based on the fix, it looks like the name "None" is an actual string from OATS DB?

mhuseinov (approved):
👍

trslater (approved):
LGTM 👍

bin/migrate-oats-data/applications/submissions/parcels/primary_contact/application_primary_contact.py
lstod: That's correct, which is why we were seeing the first name field be populated with "None"

lstod: the organization_name check is probably unnecessary but technically someone's first name could be None

PR #1342, opened 2024-01-25
Bugfix/alcs 1550 duplicated parcels

mhuseinov:
Instead of using application_id (file_number) use application_component_id (submission_id)


dhaselhan (approved):
LGTM 👍

trslater (approved):
LGTM 👍

PR #1352, opened 2024-01-29
update document table with upload date and source

lstod:
- use oats uploaded_date if not present, use when created
- Source is oats_documents document_source_code


mhuseinov (approved):
Why did you decide to run ETL script instead of existing document mapping? Was it pimplier this way?

lstod:
> Why did you decide to run ETL script instead of existing document mapping? Was it pimplier this way?

The existing document mapping would have needed refactoring to work with the 'add_timezone_and_keep_date' function

PR #1355, opened 2024-01-30
update will_import_fill

lstod:
- fix bug where nfu and naru will_import_fill would be null if N
- now N is correctly mapped to False and Y still mapped to True


mhuseinov (approved):
👍

mhuseinov (commented):
could you remove will_import_fill from insert_query.py to make it easier to back track where data came from?

trslater (approved):
LGTM 👍

PR #1362, opened 2024-01-31
Bug Fixes

dhaselhan:
* Reload entire application when updating owners, this prevents Step 1 becoming unsynced from Step 8
* Add type=button to prevent false form submits on Parcel entry
* Add logic to only fire search if they have entered something


mhuseinov (approved):
👍

sandratoh (approved):

PR #1370, opened 2024-01-31
Bug Fixes

dhaselhan:
* Add SRW Subtasks to Subtask Count
* Don't fire onOwnerUpdate as it will fire onFormChange later, it was firing onOwnerUpdate which re-loaded the app before the owner was linked


trslater (approved):
LGTM 👍

mhuseinov (approved):
👍

PR #1371, opened 2024-01-31
removing our friend None entirely

lstod:
Primary contact will now convert None to null as the first name, resulting in No Data


trslater (approved):
LGTM 👍

mhuseinov (approved):
👍

PR #1374, opened 2024-02-01
Another pass on Owners

dhaselhan:
* Save before we edit an owner. If you were to unselect or add owners to a parcel then Click Edit and Save, since the Parcel was never saved it would undo all of this upon reloading.
* Upon adding a new owner we do not want to reload the whole application as the owner has only been soft linked to the parcel, meaning reloading right away prevents the link from being done in the front end. So reload only owners when this happens.
* Fix typo
* Clear selected owner on NOIs when changing between Fee Simple -> Crown


mhuseinov (approved):
👍

urmikgov (approved):

urmikgov:
lgtm

PR #1377, opened 2024-02-01
Use ALCS document Source mapping

lstod:
-added enum to covert oats doc source to ALCS

### DIFF
--- a/bin/migrate-oats-data/documents/document_source_update.py
+++ b/bin/migrate-oats-data/documents/document_source_update.py
@@ -2,6 +2,7 @@ from common import (
     BATCH_UPLOAD_SIZE,
     setup_and_get_logger,
     add_timezone_and_keep_date_part,
+    OatsToAlcsDocumentSourceCode,
 )
 from db import inject_conn_pool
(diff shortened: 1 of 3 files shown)


bin/migrate-oats-data/documents/document_source_update.py
```diff
@@ -124,6 +125,9 @@ def _get_document_source(data):
     alcs_source = data.get("source", "")
     if alcs_source != 'oats_etl' and alcs_source:
         return alcs_source
-    # Source is ALC/APP/LG as per 1558
-    else:
-        return source
\ No newline at end of file
+    elif source:
+        source = str(
+                OatsToAlcsDocumentSourceCode[source].value
```
mhuseinov, line 130: Is it possible that the source is not in the oats enum, or are there only those 3 values in the documents table?

lstod: oats_document_source_codes table only has 3 entries and the column is not null-able in the documents table

mhuseinov (approved):
👍

urmikgov:
approved

PR #1380, opened 2024-02-02
Improve public application submission search view performance 

trslater:
- Removes application type null check (temporary)
- Removes redundant join
- View is now created via SQL directly to allow finer-grained control


services/apps/alcs/src/providers/typeorm/migrations/1706911371276-improve_pub_app_sub_search_view.ts
```diff
@@ -0,0 +1,125 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class ImprovePubAppSubSearchView1706911371276 implements MigrationInterface {
+    name = 'ImprovePubAppSubSearchView1706911371276'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`DELETE FROM "alcs"."typeorm_metadata" WHERE "type" = $1 AND "name" = $2 AND "schema" = $3`, ["VIEW","public_application_submission_search_view","alcs"]);
+        await queryRunner.query(`DROP VIEW "alcs"."public_application_submission_search_view"`);
```
dhaselhan, line 8: Your MR changed Public Application view but the migration seems to be dropping / re-creating a bunch of other ones such as notice_of_intent_submission_search_view

trslater: That is strange. Perhaps this was missed from a previous change? Any suggestions?

dhaselhan (approved):
LGTM 👍

mhuseinov (approved):
👍

PR #1383, opened 2024-02-05
Added ". Background" to NOI and alias_name fix

mhuseinov:
- added ". Background" to NOI 
- applied fix to alias_name so it will not resolve in None


trslater (approved):
LGTM 👍

PR #1385, opened 2024-02-05
Add Loading Spinners to Public Search

dhaselhan:
* Bind all values so that when component is hidden then rendered it doesn't lose settings
* use on sort change instead of subscribing to the component


trslater (approved):
LGTM 👍

urmikgov:
LGTM

urmikgov (approved):

PR #1386, opened 2024-02-05
Include deleted app types in ALCS advanced search view (ALCS-1589)

trslater:
- Temporary fix: include subs associated with deleted app/NOI/notification types in ALCS advanced search view
- Note: this is already applied to service query using this view


dhaselhan (approved):
LGTM 👍

mhuseinov (approved):
👍

PR #1389, opened 2024-02-05
Parcel Entry Fixes

dhaselhan:
* Don't clear form on search to prevent fields from going red
* Add dynamic mask and length that changes based on PID or PIN


mhuseinov (approved):
👍

PR #1409, opened 2024-02-09
Fix active days not showing in subtask list on home

trslater:
- Add active days to NOI subtask data in home controller
- Display active days on subtask list on home
- Cap displayed active days at 61


dhaselhan (approved):
LGTM 👍

urmikgov (approved):

PR #1411, opened 2024-02-09
documents to show pending on noi

urmikgov:


alcs-frontend/src/app/features/notice-of-intent/documents/documents.component.scss
```diff
@@ -17,3 +17,20 @@
...
+table {
+  position: relative; 
+
+  th mat-header-cell {
+    position: relative; 
+  }
+
+  .subheading {
+    font-size: 11px;
+    line-height: 16px;
```
trslater, line 30: Should these be using the `rem` function?

urmikgov: its kept common from alcs-frontend/src/app/features/application/decision/decision-v2/decision-documents/decision-documents.component.scss

dhaselhan: We use rem in the portal but px in ALCS. ALCS is not responsive whereas the Portal is.

dhaselhan (approved):
LGTM 👍

PR #1415, opened 2024-02-12
Don't send email for In Progress Apps / NOIs

dhaselhan:
* Add checks in ALCS for both
* You can only cancel NOI in Portal when its in progress currently, so delete that email code
* Add check in Portal for Apps
* Add tests


trslater (approved):
LGTM 🚀

sandratoh (approved):
LGTM 👍

PR #1419, opened 2024-02-13
Add app other parcel/NFU/NARU confirmation dialogs

trslater:
- Clean up unused imports
- Add other parcel confirmation dialog
- Add NFU/NARU will import fill confirmation dialog
- Refactored to use boolean toggle values instead of strings
- Added optional `title` to confirmation dialog service


mhuseinov (approved):
👍

PR #1422, opened 2024-02-13
update advanced parcel mapping for applications

lstod:
- set has other parcels to false (done because searching for where not exists in list takes a really loooong time)
- set has other parcels to true where multiple parcels exist for a submission
- set description when true to "Not migrated from OATS"


trslater (approved):
LGTM 🚀

services/apps/alcs/src/providers/typeorm/migrations/1707851589615-add_advanced_parcel_mapping.ts
```diff
@@ -0,0 +1,33 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddAdvancedParcelMapping1707851589615
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+    UPDATE
```
mhuseinov, line 8: I do not think you can set false to everything since we have actual users submitting applications now. This should only apply to etl data

services/apps/alcs/src/providers/typeorm/migrations/1707851589615-add_advanced_parcel_mapping.ts
```diff
@@ -0,0 +1,33 @@
...
+        alcs.application_submission
+    SET
+        has_other_parcels_in_community = 'False';
+    WITH parcel_count as (
+        SELECT count(*), ap.application_submission_uuid
```
mhuseinov, line 13: To make a decision that the other parcel exists for submission you will need to check OATS DB since ETL did not bring over other parcels to ALCS

lstod: placing this on hold until mapping is created. ticket unclear

mhuseinov (approved):
👍

PR #1428, opened 2024-02-14
Preserve Whitespace in Review Tables

dhaselhan:


trslater (approved):
LGTM 🚀

alcs-frontend/src/app/features/application/applicant-info/application-details/application-details.component.scss
```diff
@@ -89,6 +89,7 @@
     padding: 16px;
     margin: 24px 0 40px 0;
     grid-template-columns: minmax(60px, 1fr) minmax(60px, 1fr) minmax(60px, 1fr) minmax(60px, 1fr);
+    white-space: pre-line;
```
mhuseinov, line 92: I wonder if these should be pre-line or, rather, break-spaces. Staff could format text using tabs and white spaces and pre-line will collapse those into 1

mhuseinov (approved):
👍

alcs-frontend/src/app/features/application/applicant-info/application-details/application-details.component.scss
dhaselhan: The issue is that prettier formats the lines to be broken, so even if I delete the space manually prettier will insert one. This makes it look like every field starts with a blank space. I chose pre-line to avoid this.

PR #1430, opened 2024-02-15
500 character limit on project duration and specific activity

urmikgov:


trslater:
Since all of these text fields already have `formCotrolName` or `[formControl]` attributes, couldn't you reuse those instead of creating many new template variables?

portal-frontend/src/app/services/application-submission/application-submission-draft.service.ts
```diff
@@ -35,6 +35,38 @@ export class ApplicationSubmissionDraftService {
   }
 
   async updatePending(fileId: string, updateDto: ApplicationSubmissionUpdateDto) {
+    if (updateDto.northLandUseTypeDescription && updateDto.northLandUseTypeDescription.length > 500) {
```
dhaselhan, line 38: Rather than manually checking in the service layer, you should use the class-validators library by adding an annotation on the DTOs. If you look ed usages of @MaxLength in the existing product you can find some examples.

urmikgov: Will that not change database and add database validation? we chatted we dont want to do extra ETL or database modifications? My first choice was adding maxlength like in other dto files but i chose this option for the sole purpose that it shouldnt make db changes.

dhaselhan: Entity files are the ones that define / impact the database. DTO files are the interfaces between the front / back end and updating them will not impact the database.

urmikgov: Ahh perfect. will do it. thanks :)

urmikgov: resolved

trslater (approved):
LGTM 🚀

dhaselhan (approved):
LGTM 👍

PR #1432, opened 2024-02-15
Bypass validation for OATS Submissions

dhaselhan:
* Skip validation if the submission is from OATS and has never been updated
* Return a response from finish so we can check in the front end
* Only redirect to success page if there is a response


mhuseinov (approved):
💯

trslater (approved):
LGTM 🚀

PR #1433, opened 2024-02-15
Disable organization name obfuscation

mhuseinov:
disable organization name obfuscation since it is not 'personally identifiable information' and as result not protected under Personal Information Protection Act


lstod (approved):
LGTM

trslater (approved):
LGTM 🚀

PR #1434, opened 2024-02-15
execute oats related migrations conditionally

mhuseinov:
Run oats migrations conditionally. This will ensure that developers who work on local env do not need to introduce oats schema and related infrastructure. All hosted envs have "oats" as part of DB


dhaselhan (approved):
LGTM 👍

sandratoh (approved):

PR #1437, opened 2024-02-20
Bug Fixes

dhaselhan:
* Move NOI POFO field to match Portal
* Add focus listener to file uploader to only show the required errors once they focus back from the dialog rather than right when they click


trslater (approved):
LGTM 🚀

mhuseinov (approved):
👍

PR #1438, opened 2024-02-20
Use Portal Healthcheck for Maintenance Mode

dhaselhan:
* Add portal routed health check
* Call portal health check everytime we load the login page
* This will trigger the maintenance mode redirect if the portal is in maintenance mode


mhuseinov (approved):
👍

trslater (approved):
LGTM 🚀

PR #1440, opened 2024-02-20
Open card after creating

dhaselhan:
* Applies to Planning Reviews, Reconds, and App/NOI Modifications
* Update backend controllers to correctly return created objects
* Update dialogs to load card into URL triggering its opening


trslater (approved):
LGTM 🚀

mhuseinov (approved):
👍

PR #1442, opened 2024-02-21
Update portal confirmation dialog UI and add check to structure type change

sandratoh:
- Update confirmation dialog UI to use warning banner in body message and require title
- Add title to all components that used common dialog service
- Show confirmation dialog if previous type selection has filled input


trslater (approved):
LGTM 🚀

dhaselhan (approved):
LGTM 👍

PR #1443, opened 2024-02-21
Post-Launch ETL

lstod:
- left dummy commands for application and noi
- migrate.py is used for post-launch
- pre-launch migrate.py has been changed to migrate_initial
- updated readme to reflect changes


bin/migrate-oats-data/etl_log.txt
mhuseinov, line 1: I think this is a redundant file

lstod: yep looks like it

lstod:
yes that it correct, just left them there to keep the structure.

trslater (approved):
LGTM 🚀

PR #1445, opened 2024-02-21
Setup Typeorm Database Replication

dhaselhan:
* TypeORM supports database replication, by specifying a slave we can specify this for specific queries
* Update application advanced search to use the replica (first test)
* Remove duplicate index on app.file_number, there is already one there from the unique tag
* Add index to submission file number


services/apps/alcs/src/providers/typeorm/migrations/1708556302044-cleanups.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class Cleanups1708556302044 implements MigrationInterface {
+  name = 'Cleanups1708556302044';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DROP INDEX "alcs"."IDX_39c4f5ceb0f5a7a4c819d46a0d"`,
+    );
+    await queryRunner.query(
+      `COMMENT ON COLUMN "alcs"."notice_of_intent"."created_at" IS 'The date at which the noi was created in OATS'`,
```
mhuseinov, line 11: Interesting, so the comment was missing from the table?

mhuseinov (approved):
👍

services/apps/alcs/src/providers/typeorm/migrations/1708556302044-cleanups.ts
dhaselhan: Yes, I kept getting this in my migration and I checked and it wasn't in a previous one. Maybe it was missed.

urmikgov:
@dhaselhan if this requires changes in development.json, can we probably add the changes to be done in readme?

services/config/default.json
```diff
@@ -18,12 +18,21 @@
...
   },
   "DATABASE": {
-    "HOST": "db",
-    "PORT": 5432,
-    "NAME": "postgres",
-    "USER": "postgres",
-    "PASSWORD": "postgres",
-    "MIGRATION_TABLE": "migrations"
+    "MIGRATION_TABLE": "migrations",
+    "MASTER": {
```
dhaselhan, line 22: Combine stuff other than host into the same since we use the same user/ports.

services/apps/alcs/src/alcs/search/application/application-advanced-search.service.ts
```diff
@@ -31,9 +32,18 @@ export class ApplicationAdvancedSearchService {
 
   async searchApplications(
     searchDto: SearchRequestDto,
+    queryRunner: QueryRunner,
   ): Promise<AdvancedSearchResultDto<ApplicationSubmissionSearchView[]>> {
-    let query = await this.compileApplicationSearchQuery(searchDto);
+    let query = this.applicationSearchRepository
+      .createQueryBuilder('appSearch', queryRunner)
+      .leftJoin(
```
dhaselhan, line 39: Remove this left join

sandratoh (approved):

dhaselhan:
> @dhaselhan if this requires changes in development.json, can we probably add the changes to be done in readme?

Our current documentation does not detail the development.json, only saying to grab it from a team member. Since this is a onetime change I do not think it is worth documenting.

PR #1446, opened 2024-02-22
Decision Date Fixes

dhaselhan:
* Don't count the current decision in the date rules
* Don't set a min date if its the first decision


mhuseinov (approved):
👍

trslater (approved):
LGTM 🚀

PR #1447, opened 2024-02-22
ETL boundary amendments

lstod:
- populate boundary amendments
- link boundary amendments to decision components
- added new etl 'user'


bin/migrate-oats-data/common/constants.py
```diff
@@ -1,3 +1,4 @@
 OATS_ETL_USER = "oats_etl"
+OATS_ETL_POST_LAUNCH_USER = "oats_etl_post_launch"
```
mhuseinov, line 2: I do not think we need a "oats_etl_post_launch" user. All ETL data will be treated the same no matter what user was used to import it

services/apps/alcs/src/alcs/application-decision/application-boundary-amendment/application-boundary-amendment.entity.ts
```diff
@@ -52,6 +52,15 @@ export class ApplicationBoundaryAmendment extends Base {
   })
   period: number;
 
+  @Column({
+    select: false,
+    nullable: true,
+    type: 'int8',
+    comment:
+      'This column is NOT related to any functionality in ALCS. It is only used for ETL and backtracking of imported data from OATS. It links oats.oats_components to alcs.application_boundary_condition.',
```
mhuseinov, line 60: It does not look like that OATS has oats.oats_components. Did you mean oats.oats_alr_appl_components

bin/migrate-oats-data/applications/post_launch/application_components/boundary_amendments/application_boundary_amendments_insert.py
```diff
@@ -0,0 +1,130 @@
...
+                rows = cursor.fetchmany(batch_size)
+
+                if not rows:
+                    break
+                try:
+                    records_to_be_inserted_count = len(rows)
+
+                    _insert_records(conn, cursor, rows)
+
+                    successful_updates_count = (
```
mhuseinov, line 52: successful_insert_count should better represent the counter

services/apps/alcs/src/alcs/application-decision/application-boundary-amendment/application-boundary-amendment.entity.ts
lstod: yep, that's correct

bin/migrate-oats-data/common/constants.py
lstod: set back to oats_etl_user

PR #1457, opened 2024-02-26
Manually map in app type

dhaselhan:
* App type is a large join on search, manually join in after search in the mapping functions instead


mhuseinov (approved):
👍

services/apps/alcs/src/alcs/search/application/application-search-view.entity.ts
```diff
@@ -45,11 +37,6 @@ export class SearchApplicationSubmissionStatusType {
       .from(ApplicationSubmission, 'as2')
       .innerJoin(Application, 'a', 'a.file_number = as2.file_number')
       .withDeleted()
```
trslater, line 39: I think this is no longer needed, since it applied to application type.

dhaselhan: Removed

trslater (approved):
LGTM 🚀

PR #1464, opened 2024-02-28
fix subdivision and tur templates

urmikgov:


dhaselhan (approved):
LGTM 👍

trslater (approved):
I'm giving this an auto-approval, because there is no code to review 🚀

PR #1466, opened 2024-02-29
Remove redundant join from pub NOI sub search view

trslater:
`notice_of_intent_decision` was being joined to
itself, but the condition did not relate different
columns, so should have no effect. Removing this 
prevents a nested loop that explodes the number of
rows scanned.

This should speed it up substantially.


lstod (approved):
👍

PR #1476, opened 2024-03-05
Filter out NOI's with draft decisions from pub search

trslater:
- Filter added to pub search NOI view
- Migration for entity change


dhaselhan (approved):
LGTM 👍

mhuseinov (approved):
👍

PR #1478, opened 2024-03-05
Convert pub app search view entity to use query builder

trslater:
Public application search view entity was using the query builder originally, but I had temporarily switched to entering the query manually to save time before launch. This reverts that change, but with the same performance improvements applied.


mhuseinov (approved):
👍

dhaselhan (approved):
LGTM 👍

PR #1486, opened 2024-03-06
Public search scroll to result

urmikgov:


portal-frontend/src/app/features/public/search/public-search.component.ts
```diff
@@ -128,7 +128,8 @@ export class PublicSearchComponent implements OnInit, OnDestroy {
     private codeService: CodeService,
     private statusService: StatusService,
     private toastService: ToastService,
-    private titleService: Title
+    private titleService: Title,
+    private elementRef: ElementRef,
```
trslater, line 132: It's not really clear what element this references. Might be better to use a `ViewChild`

portal-frontend/src/app/features/public/search/public-search.component.ts
```diff
@@ -193,6 +194,14 @@ export class PublicSearchComponent implements OnInit, OnDestroy {
     sessionStorage.setItem(SEARCH_SESSION_STORAGE_KEY, searchDto);
 
     this.isLoading = true;
+    const element = this.elementRef.nativeElement.querySelector(`#${CSS.escape('searchResults')}`);
```
dhaselhan, line 197: Remove the element ref, and use the ID you gave it with scrollToElement util

trslater: Yeah, this is probably better since it keeps behaviour consistent.

portal-frontend/src/app/features/public/search/public-search.component.ts
urmikgov: i have kept it consistent with what we use in decision page

portal-frontend/src/app/features/public/search/public-search.component.ts
urmikgov: made changes

trslater (approved):
LGTM 🚀

dhaselhan (approved):
LGTM 👍

PR #1488, opened 2024-03-07
Add SRW to staff journals

lstod:
- add srw to staff journal
- populate etl author


bin/migrate-oats-data/srw/sql/srw_staff_journal.sql
```diff
@@ -0,0 +1,3 @@
+SELECT osj.journal_date, osj.journal_text, osj.revision_count, osj.staff_journal_entry_id, an."uuid" 
+FROM oats.oats_staff_journal_entries osj
+JOIN alcs.notification an ON an.file_number = osj.alr_application_id::TEXT
```
mhuseinov, line 3: could you add SRW filter here since it is possible that later we will have different types on notification table

bin/migrate-oats-data/srw/srw_staff_journal.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def _map_timezone(data):
+    date = data.get("journal_date", "")
+    journal_date = add_timezone_and_keep_date_part(date)
+    data["journal_date"] = journal_date
+    return data
+
+
+def _add_user(data):
```
mhuseinov, line 139: you could just reference uuid in data["user"]

bin/migrate-oats-data/srw/sql/srw_staff_journal.sql
lstod: Is there a reason we not want those to propagate to staff journals? I would expect every notification with a staff journal associated with it we would want in this table

mhuseinov: The short answer is data integrity. Long answer: the notification table could store not only SRW but any other type. It means that for subsequent ETLs you need to keep in mind that anything other than records linked to SRW could be outdated in ALCS DB or even not relevant to the ALCS version of the feature. As a result, you could end up with "somewhat correct" data which makes it unreliable

trslater (approved):
LGTM 🚀

bin/migrate-oats-data/srw/sql/srw_staff_journal.sql
lstod: Updated with SRW filter

mhuseinov (approved):
:+1

PR #1490, opened 2024-03-07
Logging changes

dhaselhan:
* Don't log all headers, they have the auth key and don't really help anyway
* Prevent keycloak from logging every token expiry


sandratoh (approved):

trslater (approved):
LGTM 🚀

PR #1493, opened 2024-03-11
Configure E2E testing package

trslater:
- Move end-to-end tests
- Clean up env data
- Clean up unused files
- Add types for Node
- Allow .env configurable secrets for local dev
- Allow .env configurable base URL
- Align editor style with other projects
- Add `template.env` and instructions to setup local secrets
- Add some more detailed instructions
- Ignore appropriate files
- Configure projects for browsers
- Frontend (e.g., portal, ALCS) selectable via directory
- Add basic portal inbox test
- Add parameterized login fixture
- Add/config eslint/prettier


e2e/.editorconfig
```diff
@@ -0,0 +1,16 @@
+# Editor configuration, see https://editorconfig.org
```
dhaselhan, line 1: Should we use prettier like all our other packages?

e2e/playwright.config.ts
```diff
@@ -35,16 +36,27 @@ export default defineConfig({
     {
       name: 'chromium',
       use: { ...devices['Desktop Chrome'] },
+      testIgnore: '**',
```
dhaselhan, line 39: Is this temporary? Whats the long term plan here?

e2e/.editorconfig
trslater: I copied this from portal frontend, so does it need to be updated as well?

e2e/playwright.config.ts
trslater: After re-investigating this, I don't think this works for multiple browsers. I'll come up with a better solution.

e2e/.editorconfig
dhaselhan: Not sure, I have no idea what this file does.

trslater: It configures the code style for editors. It's set up the same for portal and ALCS frontends, not for services. Perhaps it is part of Angular scaffolding?

For now I'll leave it, but I'll install prettier as a dev dep and include `.prettierrc` as well.

e2e/playwright.config.ts
trslater: Project dependencies did not work how I thought. I've changed the way this works. Now, projects only select browser, and frontend is selected via directory. Each top-level directory will have its own fixtures file that sets the base URL via the environment (this will allow us to modify the URL for dev/test/local as needed).

e2e/.editorconfig
trslater: Added prettier and eslint

e2e/package.json
```diff
@@ -18,7 +18,17 @@
     "url": "https://github.com/bcgov/alcs/issues"
   },
   "homepage": "https://github.com/bcgov/alcs#readme",
+  "dependencies": {
+    "@playwright/test": "^1.32.0",
+    "@types/node": "^20.11.24",
```
dhaselhan, line 23: nitpick: This doesn't matter since this won't get shipped but types should go in devDependencies

dhaselhan (approved):
LGTM 👍

e2e/package.json
trslater: Could be. I considered it as a main dependency since running the E2E tests is what this package does. I think you could make the case either way.

dhaselhan: The transpiled code will be JS not TS, meaning the types won't be used.

e2e/tests/portal/fixtures.ts
```diff
@@ -0,0 +1,30 @@
...
+
+interface FixtureOptions {
+  userPrefix: string;
+}
+
+interface Fixtures {
+  inboxLoggedIn: Page;
+}
+
+export const test = base.extend<FixtureOptions & Fixtures>({
```
mhuseinov, line 16: Ideally the name of fixture should represent the action that it performs, in this case it is login

trslater: It does, on line 18: `inboxLoggedIn`. The way fixtures work in Playwright is that you extend the main test function with an object that contains the fixtures.

e2e/package.json
trslater: Good point. Changed.

mhuseinov (approved):
👍

PR #1495, opened 2024-03-11
Open portal non-decision documents in new browser tab iframe

sandratoh:
- Add new `utils` method to open file in iframe with `fileName` as browser tab title
- Apply method to all non-decision documents in frontend
- Update backend controller and test files


dhaselhan (approved):
LGTM 👍

trslater (approved):
LGTM 🚀

PR #1500, opened 2024-03-12
login redirect fix

urmikgov:
In the original implementation, the refreshTokens method immediately redirects to /login if the expires time is reached, without considering the refreshExpires time. This means that even if the refresh token is still valid, the user will be redirected to /login prematurely.


dhaselhan:
Can you update your description to explain how this fixes the login issue?

dhaselhan (approved):
LGTM 👍

trslater (approved):
LGTM 🚀

PR #1503, opened 2024-03-12
Fix report path

trslater:
- Upload artifact does not support `working-directory`, so path needs to be relative to repo root


dhaselhan (approved):
LGTM 👍

PR #1504, opened 2024-03-13
added default etl user uuid to srw

mhuseinov:
- add default user to SRW submission
- move default user to constants


lstod (approved):
LGTM 💯

PR #1510, opened 2024-03-14
Add Planning Review Decisions

dhaselhan:
* Update Decision # Generation to include NOIs and PRs

### DIFF
--- a/services/apps/alcs/src/alcs/planning-review/planning-review-decision/planning-review-decision.service.ts
+++ b/services/apps/alcs/src/alcs/planning-review/planning-review-decision/planning-review-decision.service.ts
@@ -1,0 +1,351 @@
+import {
+  ServiceNotFoundException,
+  ServiceValidationException,
+} from '@app/common/exceptions/base.exception';
+import { MultipartFile } from '@fastify/multipart';
+import { Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { In, IsNull, Repository } from 'typeorm';
+import {
+  DOCUMENT_SOURCE,
+  DOCUMENT_SYSTEM,
+} from '../../../document/document.dto';
+import { DocumentService } from '../../../document/document.service';
+import { User } from '../../../user/user.entity';
+import { formatIncomingDate } from '../../../utils/incoming-date.formatter';
+import { PlanningReviewService } from '../planning-review.service';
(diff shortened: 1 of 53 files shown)


services/apps/alcs/src/alcs/planning-review/planning-review-decision/planning-review-decision.service.ts
```diff
@@ -0,0 +1,351 @@
...
+      },
+      relations: {
+        document: true,
+      },
+    });
+
+    if (decisionDocument) {
+      return this.documentService.getDownloadUrl(
+        decisionDocument.document,
+        true, // FIXME: Document does not open inline despite flag being true
```
mhuseinov, line 249: is this still relevant after changes that Sandra made with iframe?

services/apps/alcs/src/providers/typeorm/migrations/1683048988258-generate_next_resolution_number.ts
```diff
@@ -8,33 +8,42 @@ export class generateNextResolutionNumber1683048988258
       `CREATE OR REPLACE FUNCTION alcs.generate_next_resolution_number(p_resolution_year integer)
           RETURNS integer
           LANGUAGE plpgsql
-        AS $function$
-          
-            declare next_resolution_number integer;
-          BEGIN	
+      AS $function$
```
mhuseinov, line 11: I do not think that this is going to be executed since the timestamp for migration did not change

dhaselhan: ah oops, looks like I created a new one and updated the old one

services/apps/alcs/src/alcs/planning-review/planning-review-decision/planning-review-decision.service.ts
dhaselhan: Sandra has not applied the file changes to ALCS yet, only the portal, so I think it applies for now

services/apps/alcs/src/providers/typeorm/migrations/1683048988258-generate_next_resolution_number.ts
dhaselhan: Reverted

mhuseinov (approved):
👍

PR #1513, opened 2024-03-14
Open alcs and portal document inline with standardized method

sandratoh:
- Update alcs CSP to allow frame plugin source from document storage
- Update alcs `openFileInline` logic to open file in new window with file name as browser tab title
- Update portal `openFileWindow` to use the same logic as ALCS `openFileInline` and also renamed it
- Update all portal usage of util method
- Revert backend changes done in https://github.com/bcgov/alcs/pull/1495 (already have filename in frontend parent wrapper)


trslater (approved):
LGTM 🚀

PR #1514, opened 2024-03-15
Decide reponse sent status and date from events rather than status and last status update

trslater:
- Find status response event in events
- Use existence to set response sent status
- Use date for response date
- Precents cancelled status/date taking precedence
- Fix random typo


dhaselhan (approved):
LGTM 👍

PR #1517, opened 2024-03-18
Fix not scrolling to spinner on initial search

trslater:
- Remove duplicate spinner
- Restructure template conditionals so same spinner is used for both initial load and subsequent loads
- Show 'Search Results' title while loading to be consistent with subsequent loads
- Scroll to now always-visible wrapper div
- Use hack to delay scrolling until spinner is visible


dhaselhan (approved):
LGTM 👍

PR #1518, opened 2024-03-18
Four bug fixes for ALCS applicant info

trslater:
- Display correct org text for individual owners
- Show area-wide fill question in ALCS detail
- Align proposal field order with portal order
- Add m^2 to ALCS applicant structures table


dhaselhan (approved):
LGTM 👍

PR #1519, opened 2024-03-18
Add Planning Review to Search, Delete Covenants

dhaselhan:


alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.ts
```diff
@@ -173,7 +173,7 @@ export class FileTypeFilterDropDownComponent implements AfterViewInit {
   }
 
   ngAfterViewInit(): void {
-    this.treeControl.expandAll();
+    //this.treeControl.expandAll();
```
trslater, line 176: I feel like we might as well just clean up this whole function.

trslater (approved):
LGTM 🚀

Leaving it up to you if you want to make the one tiny change I suggested.

alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.ts
dhaselhan: Removed

PR #1521, opened 2024-03-19
SET survey plan

lstod:
if SRWP or SURV file present set survey plan indicator to true

### DIFF
--- a/bin/migrate-oats-data/menu/post_launch_commands/import_all.py
+++ b/bin/migrate-oats-data/menu/post_launch_commands/import_all.py
@@ -27,6 +27,7 @@ def import_all(console, args):
         import_documents(batch_size=import_batch_size)
 
         console.log("Process SRW Updates")
+        # These are updates that need to happen after the SRW document import
         srw_survey_plan_update(batch_size=import_batch_size)
 
         console.log("Done")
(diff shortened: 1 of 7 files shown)


trslater (approved):
LGTM 🚀

bin/migrate-oats-data/menu/post_launch_commands/import_all.py
```diff
@@ -26,4 +26,7 @@ def import_all(console, args):
         console.log("Processing Documents")
         import_documents(batch_size=import_batch_size)
 
+        console.log("Process SRW Updates")
+        srw_survey_plan_update(batch_size=import_batch_size)
+
```
mhuseinov, line 31: Why is this separate from the general SRW import? Could you move it into the function that wraps all SRW-related ETL calls?

lstod: Not possible, it needs to be done after the document import, because it depends on those file types being present. I could move it into the document import

mhuseinov: Thanks for clarifying, a comment highlighting the dependency will be good enough

mhuseinov (approved):
👍

PR #1522, opened 2024-03-19
Update SRW ownership mapping

lstod:
updated logic based on pid and pin being present or not


bin/migrate-oats-data/srw/submission/parcel/srw_parcel_init.py
```diff
@@ -114,7 +114,7 @@ def _map_data(row, insert_index):
 def _map_ownership_type_code(data):
     return (
         OatsToAlcsOwnershipType.CROWN.value
-        if data["pin"]
+        if data.get("pin") or not data.get("pid")
```
mhuseinov, line 117: I thought that the idea is to use the ownership type code from OATS and if it is set to NULL then fallback to this conditional logic based on "pid", "pin"

PR #1524, opened 2024-03-19
Update Parcel ownership mapping

lstod:
- use mapping from applications/nois
- if property_owner_type is missing ise logic based of PIN and PID


mhuseinov (approved):
👍

PR #1527, opened 2024-03-19
PR Home page fixes

dhaselhan:
* Actually fetch subtasks for home page
* Load card and board so we can redirect properly
* Update home-page from non-Applications to Planning Referrals
* Remove max height on card dialog content so it scrolls nicely


trslater (approved):
LGTM 🚀

PR #1528, opened 2024-03-20
Fix ALC intake email updating even when not confirmed

trslater:
- Add `updateOnSave` option to `inline-text` component
- Set `updateOnSave` to false for intake email

### DIFF
--- a/alcs-frontend/src/app/shared/inline-editors/inline-text/inline-text.component.ts
+++ b/alcs-frontend/src/app/shared/inline-editors/inline-text/inline-text.component.ts
@@ -15,6 +15,7 @@ import {
   styleUrls: ['./inline-text.component.scss'],
 })
 export class InlineTextComponent implements AfterContentChecked {
+  @Input() updateOnSave: boolean = true;
   @Input() value?: string | undefined;
   @Input() placeholder: string = 'Enter a value';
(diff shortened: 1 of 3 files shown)


alcs-frontend/src/app/shared/inline-editors/inline-text/inline-text.component.ts
```diff
@@ -41,14 +42,16 @@ export class InlineTextComponent implements AfterContentChecked {
...
+        this.value = this.pendingValue;
+      }
     }
 
     this.isEditing = false;
   }
 
   cancelEdit() {
     this.isEditing = false;
-    this.pendingValue = this.value;
```
dhaselhan, line 52: This may cause other issues, if I edit an inline text field and click cancel then click edit again, does it show the original value or the one I had cancelled previously?

trslater: This line doesn't do anything, because `pendingValue` is already set, and `value` hasn't changed.

trslater: `pendingValue` changes with the input, but it'll be reset by `startEdit` when you edit again.

PR #1529, opened 2024-03-20
Delete V1 Decisions + Add PR Meetings to Commissioner Controller

dhaselhan:
* Move Application Decision Meetings to Application Module, they do not reference decisions but are used in other Application services
* Move the controller to a new meeting module that will serve both App and PR meetings
* Clean up old V1 Decision Code


trslater (approved):
LGTM 🚀

PR #1530, opened 2024-03-20
More PR Card Fixes

dhaselhan:
* Load assignee so it shows the avatar circle
* Only add due date padding when its not the only child
* Use actual PR type instead of fixed type


trslater (approved):
LGTM 🚀

PR #1533, opened 2024-03-20
Fix incorrect info showing in PDF for NOI structures

trslater:
- Fixed `Yes` incorrectly used in template instead of `true`, causing farm use description to never show
- Fixed accessory structure type missed from check for residential use description check
- Cleaned up spaces in template


dhaselhan (approved):
LGTM 👍

PR #1542, opened 2024-03-25
Add Production Routes to API Guard

dhaselhan:
* Production is set to have /api/ added to all routes breaking how the maintenance guard normally checks.


mhuseinov (approved):
👍

PR #1557, opened 2024-03-28
Add correct error class to 'Approval Dependent' field when invalid

trslater:
- Refactor to use boolean value directly
- Add error class when touched and invalid


dhaselhan (approved):
LGTM 👍

PR #1559, opened 2024-03-28
UI Changes

dhaselhan:
* Change detailed page layout for all types to scroll the child content not the window
* Add index column to evidentiary records
* Add scrollable to document sorting to it scrolls
* Add mask to PID on Proposal


trslater (approved):
LGTM 🚀

PR #1561, opened 2024-04-02
Add Right Click Sorting Menu

dhaselhan:
* Add a Move to Top / Bottom menu when right clicking on either Application or PR evidentiary records


mhuseinov (approved):
👍

PR #1564, opened 2024-04-02
Improve app/NOI gen res button visibility, multiline text editing/viewing

trslater:
- Improve app/NOI gen res number button visibility
- Improve app/NOI decision long text field UX
- Remove redundant mixin def
- Preserve new lines app/NOI decision text fields


dhaselhan (approved):
LGTM 👍

PR #1566, opened 2024-04-02
Add some unit tests

dhaselhan:


alcs-frontend/src/app/services/application/application-document/application-document.service.spec.ts
```diff
@@ -51,11 +67,27 @@ describe('ApplicationDocumentService', () => {
...
+          uuid: '1',
+        },
+      ]),
+    );
+
+    const res = await service.getApplicantDocuments('1');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].uuid).toEqual('1');
```
trslater, line 83: The description only says a GET request is made, it doesn't say it needs to return a result or that it must contain a UUID. Are these really necessary?

alcs-frontend/src/app/services/application/application-submission/application-submission.service.spec.ts
```diff
@@ -140,4 +140,31 @@ describe('ApplicationSubmissionService', () => {
     expect(result).toEqual(mockSubmittedApplication);
     expect(mockHttpClient.get).toBeCalledTimes(1);
   });
+
+  it('should call patch when updating a submission', async () => {
+    mockHttpClient.patch.mockReturnValue(of(mockSubmittedApplication));
+
+    const result = await service.update('1', {});
+
+    expect(result).toEqual(mockSubmittedApplication);
```
trslater, line 149: Same as previous.

alcs-frontend/src/app/services/inquiry/inquiry-document/inquiry-document.service.spec.ts
```diff
@@ -51,6 +51,22 @@ describe('InquiryDocumentService', () => {
...
+          code: '1',
+        },
+      ]),
+    );
+
+    const res = await service.fetchTypes();
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].code).toEqual('1');
```
trslater, line 67: And again here.

alcs-frontend/src/app/services/notice-of-intent/decision-v2/notice-of-intent-decision-v2.service.spec.ts
```diff
@@ -62,11 +63,39 @@ describe('NoticeOfIntentDecisionV2Service', () => {
...
+      of([
+        {
+          fileNumber: '1',
+        },
+      ]),
+    );
+
+    const res = await service.fetchCodes();
+
+    expect(res).toBeDefined();
```
trslater, line 77: Same as previous.

alcs-frontend/src/app/services/notice-of-intent/decision-v2/notice-of-intent-decision-v2.service.spec.ts
```diff
@@ -214,4 +243,31 @@ describe('NoticeOfIntentDecisionV2Service', () => {
...
+  it('should show a toast message if load decision fails', async () => {
+    httpClient.get.mockReturnValue(
+      throwError(() => {
+        new Error('');
+      }),
+    );
+
+    const res = await service.loadDecision('uuid');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
```
trslater, line 270: Do we need to check if `get` was called? What if this function fails before `get` is called.

alcs-frontend/src/app/services/notice-of-intent/noi-document/noi-document.service.spec.ts
```diff
@@ -35,13 +35,45 @@ describe('NoiDocumentService', () => {
...
         },
-      ])
+      ]),
+    );
+
+    const res = await service.listAll('1');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].uuid).toEqual('1');
```
trslater, line 51: Same same.

alcs-frontend/src/app/services/notice-of-intent/noi-document/noi-document.service.spec.ts
```diff
@@ -35,13 +35,45 @@ describe('NoiDocumentService', () => {
...
+          code: '1',
+        },
+      ]),
+    );
+
+    const res = await service.fetchTypes();
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].code).toEqual('1');
```
trslater, line 67: And here.

alcs-frontend/src/app/services/notice-of-intent/noi-document/noi-document.service.spec.ts
```diff
@@ -51,11 +83,27 @@ describe('NoiDocumentService', () => {
...
+          uuid: '1',
+        },
+      ]),
+    );
+
+    const res = await service.getApplicantDocuments('1');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].uuid).toEqual('1');
```
trslater, line 99: And here.

alcs-frontend/src/app/services/notification/notification-document/notification-document.service.spec.ts
```diff
@@ -35,13 +35,45 @@ describe('NoiDocumentService', () => {
...
         },
-      ])
+      ]),
+    );
+
+    const res = await service.listAll('1');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].uuid).toEqual('1');
```
trslater, line 51: I'm starting to regret the decision to mark all of these 😆  I'm committed now!

alcs-frontend/src/app/services/notification/notification-document/notification-document.service.spec.ts
```diff
@@ -35,13 +35,45 @@ describe('NoiDocumentService', () => {
...
+          code: '1',
+        },
+      ]),
+    );
+
+    const res = await service.fetchTypes();
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].code).toEqual('1');
```
trslater, line 67: And here.

alcs-frontend/src/app/services/planning-review/planning-referral.service.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    httpClient.get.mockReturnValue(
+      throwError(() => {
+        new Error('');
+      }),
+    );
+
+    const res = await service.fetchByCardUuid('card-uuid');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res).toBeUndefined();
```
trslater, line 58: This one's kind of the opposite problem. Shouldn't we only be checking for the toast?

alcs-frontend/src/app/services/planning-review/planning-referral.service.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const res = await service.create({
+      planningReviewUuid: '',
+      referralDescription: '',
+      submissionDate: 0,
+    });
+
+    expect(httpClient.post).toHaveBeenCalledTimes(1);
+    expect(res).toBeUndefined();
+    expect(toastService.showErrorToast).toHaveBeenCalledTimes(1);
```
trslater, line 93: Same toast issue.

alcs-frontend/src/app/services/planning-review/planning-referral.service.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+      }),
+    );
+
+    const res = await service.update('', {
+      referralDescription: '',
+      submissionDate: 0,
+    });
+
+    expect(httpClient.patch).toHaveBeenCalledTimes(1);
+    expect(res).toBeUndefined();
```
trslater, line 124: Toast issue.

alcs-frontend/src/app/services/planning-review/planning-referral.service.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    httpClient.delete.mockReturnValue(
+      throwError(() => {
+        new Error('');
+      }),
+    );
+
+    const res = await service.delete('');
+
+    expect(httpClient.delete).toHaveBeenCalledTimes(1);
+    expect(res).toBeUndefined();
```
trslater, line 150: Toast issue.

alcs-frontend/src/app/services/planning-review/planning-review-decision/planning-review-decision.service.spec.ts
```diff
@@ -62,6 +62,34 @@ describe('PlanningReviewDecisionService', () => {
...
+        {
+          fileNumber: '1',
+        },
+      ]),
+    );
+
+    const res = await service.fetchCodes();
+
+    expect(res).toBeDefined();
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
```
trslater, line 77: Seems like it doesn't check if a code has been returned, and the mock seems to return a file number. Is that what is meant by code?

alcs-frontend/src/app/services/planning-review/planning-review-decision/planning-review-decision.service.spec.ts
```diff
@@ -62,6 +62,34 @@ describe('PlanningReviewDecisionService', () => {
...
+  it('should show a toast message if fetch codes fails', async () => {
+    httpClient.get.mockReturnValue(
+      throwError(() => {
+        new Error('');
+      }),
+    );
+
+    const res = await service.fetchCodes();
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
```
trslater, line 89: Again, I feel like only the toast should be expected here.

alcs-frontend/src/app/services/planning-review/planning-review-decision/planning-review-decision.service.spec.ts
```diff
@@ -198,4 +226,31 @@ describe('PlanningReviewDecisionService', () => {
...
+  it('should show a toast message if load decision fails', async () => {
+    httpClient.get.mockReturnValue(
+      throwError(() => {
+        new Error('');
+      }),
+    );
+
+    const res = await service.loadDecision('uuid');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
```
trslater, line 253: Only expect toast again.

alcs-frontend/src/app/services/planning-review/planning-review-document/planning-review-document.service.spec.ts
```diff
@@ -35,7 +35,39 @@ describe('PlanningReviewDocumentService', () => {
...
+          uuid: '1',
+        },
+      ]),
+    );
+
+    const res = await service.listAll('1');
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].uuid).toEqual('1');
```
trslater, line 51: Only check `get`.

alcs-frontend/src/app/services/planning-review/planning-review-document/planning-review-document.service.spec.ts
```diff
@@ -35,7 +35,39 @@ describe('PlanningReviewDocumentService', () => {
...
+          code: '1',
+        },
+      ]),
+    );
+
+    const res = await service.fetchTypes();
+
+    expect(httpClient.get).toHaveBeenCalledTimes(1);
+    expect(res.length).toEqual(1);
+    expect(res[0].code).toEqual('1');
```
trslater, line 67: Only check `get`

trslater:
I got a little carried away making what was essentially the same comment a zillion times, so I decided to get rid of them and add a single comment: in general, it seems like the descriptions are not lining up with expects in lots of places. From my understanding, unit tests should generally test one thing, and not be concerned with the steps to achieve the result being tested.

dhaselhan:
@trslater Thanks for the comment, what would you like me to change? I could update most of the descriptions to say return a result but that seems like boiler plate.

For example:
The description only says a GET request is made, it doesn't say it needs to return a result or that it must contain a UUID. Are these really necessary?

Do you want to rename the test to be 'should make a get request, return a result, and the result should contain a UUID when calling XXX'? Most service calls return a result so it feels redundant to add it to each test case title. If you want me to delete the expect statements I can also do that, but thats only weakening the tests by checking less things.

trslater:
> @trslater Thanks for the comment, what would you like me to change? I could update most of the descriptions to say return a result but that seems like boiler plate.
> 
> For example: The description only says a GET request is made, it doesn't say it needs to return a result or that it must contain a UUID. Are these really necessary?
> 
> Do you want to rename the test to be 'should make a get request, return a result, and the result should contain a UUID when calling XXX'? Most service calls return a result so it feels redundant to add it to each test case title. If you want me to delete the expect statements I can also do that, but thats only weakening the tests by checking less things.

I was more trying to open a discussion than directly requesting changes, since you have a lot more experience than I do. I always strive for mutual learning. From everything I've read, learned in school, etc., the rule of thumb is to have a single assert per unit test. Each test case has an end result we're looking for, and we only need to assert that end result has been achieved. This makes the tests less brittle as we may change the steps along the way, but as long as the end result is the same, we're satisfying the requirement. If we want to test that it returns a value, that would be another test case. But I'm also flexible.

alcs-frontend/src/app/services/application/application-document/application-document.service.spec.ts
```diff
@@ -101,11 +133,28 @@ describe('ApplicationDocumentService', () => {
...
+    );
+
+    const res = await service.update('uuid', {
+      fileName: '',
+      source: DOCUMENT_SOURCE.APPLICANT,
+      typeCode: DOCUMENT_TYPE.CERTIFICATE_OF_TITLE,
+      visibilityFlags: [],
+    });
+
+    expect(httpClient.post).toHaveBeenCalledTimes(1);
```
trslater, line 158: Why not check result here?

alcs-frontend/src/app/services/application/decision/application-decision-v2/application-decision-condition/application-decision-condition.service.spec.ts
```diff
@@ -63,4 +63,63 @@ describe('ApplicationDecisionConditionService', () => {
...
+
+    await service.fetchPlanNumbers('1');
+
+    expect(mockHttpClient.get).toHaveBeenCalledTimes(1);
+  });
+
+  it('should show a toast message if fetchPlanNumbers fails', async () => {
+    mockHttpClient.get.mockReturnValue(
+      throwError(() => {
+        new Error('');
```
trslater, line 82: I'm not sure what a work around for this would be, but I think this could lend false confidence that an error toast is being shown for every type of failure, since we're only checking a specific case (there are a bunch of these).

alcs-frontend/src/app/services/application/application-document/application-document.service.spec.ts
dhaselhan: Good call, added checking of result

alcs-frontend/src/app/services/application/decision/application-decision-v2/application-decision-condition/application-decision-condition.service.spec.ts
```diff
@@ -63,4 +63,63 @@ describe('ApplicationDecisionConditionService', () => {
...
+  it('should make an http get and show a success toast when fetchPlanNumbers succeeds', async () => {
+    mockHttpClient.get.mockReturnValue(
+      of({
+        applicationFileNumber: '1',
+      }),
+    );
+
+    await service.fetchPlanNumbers('1');
+
+    expect(mockHttpClient.get).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
dhaselhan: Test name updated

trslater (approved):
LGTM 🚀

PR #1569, opened 2024-04-02
Inquiry Fixes

dhaselhan:
* Set title on inquiry page
* Add code for updating inquiry region and local government
* Add error message to create dialog
* Add spacing to create parcel table


mhuseinov (approved):
👍

PR #1570, opened 2024-04-03
migrations to create cards for applications and nois

mhuseinov:
- create cards for provided applications
- create cards for provided NOIs
- backup tables for applications and NOIs that got card created and linked


dhaselhan (approved):
LGTM 👍

PR #1571, opened 2024-04-03
Add some Portal Unit Tests

dhaselhan:


mhuseinov (approved):
👍

portal-frontend/src/app/services/document/document.service.spec.ts
```diff
@@ -23,4 +44,27 @@ describe('DocumentService', () => {
   it('should be created', () => {
     expect(service).toBeTruthy();
   });
+
+  it('should make a get request for getUploadUrl', async () => {
+    mockHttpClient.get.mockReturnValue(of({}));
+
+    const res = await service.getUploadUrl('fileNumber', null);
+    expect(service).toBeTruthy();
```
trslater, line 52: Won't this always be true? Also, doesn't test whether GET request is made.

dhaselhan: Woops! Yep, looks like I copied that from the default unit test and forgot to change it to expect get. Fixed.

trslater (approved):
LGTM 🚀

PR #1576, opened 2024-04-03
Update coverage rules

dhaselhan:
* Exclude DTOs, Modules, and Entities
* Exclude Command and Import Folder (Ansley said to keep not delete)


mhuseinov (approved):
👍

PR #1584, opened 2024-04-04
inquiry staff journal import + etl bugfix

lstod:
- import issue notes into alcs staff journals
- fix issue with import all and clean all still touching srws and previous code
- import and clean commands now only though pr and inquires

### DIFF
--- a/bin/migrate-oats-data/documents/post_launch/oats_documents_to_alcs_documents_srw.py
+++ b/bin/migrate-oats-data/documents/post_launch/oats_documents_to_alcs_documents_srw.py
@@ -181,7 +181,7 @@ def document_clean(conn=None):
     logger.info("Start documents cleaning")
     with conn.cursor() as cursor:
         cursor.execute(
-            f"DELETE FROM alcs.document WHERE audit_created_by = '{OATS_ETL_USER}' AND audit_created_at > '2024-02-08';"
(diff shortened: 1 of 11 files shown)


bin/migrate-oats-data/documents/post_launch/oats_documents_to_alcs_documents_srw.py
```diff
@@ -181,7 +181,7 @@ def document_clean(conn=None):
     logger.info("Start documents cleaning")
     with conn.cursor() as cursor:
         cursor.execute(
-            f"DELETE FROM alcs.document WHERE audit_created_by = '{OATS_ETL_USER}' AND audit_created_at > '2024-02-08';"
+            f"DELETE FROM alcs.document WHERE audit_created_by = '{OATS_ETL_USER}' AND audit_created_at > '2024-02-08' AND oats_application_id IS NOT NULL;"
```
mhuseinov, line 184: I think this will delete all files, not only SRW documents

lstod: in theory the only documents imported past that date that have an application_id should be SRWs, but yes it is not exclusive to SRWs

mhuseinov (approved):
👍

PR #1588, opened 2024-04-08
Feature/alcs 1739 drag drop doc upload

mhuseinov:
Drag&Drop document upload in ALCS:
- application documents and decisions
- noi documents and decisions
- planning review documents and decisions
- srw documents and decisions
- inquiry documents and decisions


trslater (approved):
LGTM 🚀

Though, I feel like perhaps we should make this a shared component in the future.

dhaselhan (approved):
LGTM 👍

PR #1591, opened 2024-04-08
Add maintenance banner to ALCS and Portal

sandratoh:
- Add admin configs to control portal banner show status and message
- Update admin UI with:
  - toggle slider for existing and new maintenance configs
  - new input for banner message
  - new confirmation dialog before banner display
  - copy and icon changes
- Add maintenance banner to ALCS and Portal header
- Update backend with shared MaintenanceService and individual controllers to return banner info
- Use observables in ALCS frontend to reload header if user updates maintenance features
- Add migration file to save default maintenance config values


alcs-frontend/src/app/features/admin/configuration/configuration.component.scss
```diff
@@ -1,18 +1,73 @@
...
 
-.editing {
-  max-width: 600px;
+:host::ng-deep {
+  .toggle {
+    mat-label {
+      margin-right: 8px;
+    }
+
+    & .mdc-form-field {
```
dhaselhan, line 42: comment: I'm surprised we are doing so much CSS for Admin when historically its been a dont care

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
```diff
@@ -3,37 +3,92 @@ import {
...
+  private async updateConfig(configName: CONFIG_VALUE, configValue: string) {
+    await this.adminConfigurationService.setConfiguration(configName, configValue);
+    this.loadConfigs();
   }
 
-  async onSave() {
-    await this.adminConfigurationService.setConfiguration(CONFIG_VALUE.PORTAL_MAINTENANCE_MODE, this.maintenanceMode);
-    this.loadConfigs();
-    this.isEditing = false;
+  private getConfigValue(
```
dhaselhan, line 38: I would recommend throwing an error or returning undefined when a config is not found, if we were to rename configs on the backend this would result in no front end errors but not actually be working.

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
```diff
@@ -3,37 +3,92 @@ import {
...
+    configs: { name: CONFIG_VALUE; value: string }[],
+    configName: CONFIG_VALUE,
+    isBoolean: boolean = false,
+  ): string | boolean {
```
dhaselhan, line 42: suggestion: Update this function to use generics so you don't need to do any casting on the calling side.

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
```diff
@@ -3,37 +3,92 @@ import {
...
+      this.maintenanceBanner = this.getConfigValue(configs, CONFIG_VALUE.APP_MAINTENANCE_BANNER, true) as boolean;
+      this.maintenanceBannerMessage = this.getConfigValue(
+        configs,
+        CONFIG_VALUE.APP_MAINTENANCE_BANNER_MESSAGE,
+      ) as string;
+    }
+  }
+
+  onToggleMaintenanceBanner(event: MatSlideToggleChange) {
+    if (event.checked === true) {
```
dhaselhan, line 60: nitpick: If event.checked its a boolean, do you need === true?

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
```diff
@@ -3,37 +3,92 @@ import {
...
+        .open(MaintenanceBannerConfirmationDialogComponent, {
+          data: {
+            message: this.maintenanceBannerMessage,
+          },
+        })
+        .beforeClosed()
+        .subscribe((didConfirm) => {
+          if (didConfirm) {
+            this.updateConfig(CONFIG_VALUE.APP_MAINTENANCE_BANNER, this.maintenanceBanner.toString());
+            this.maintenanceService.setShowBanner(true);
```
dhaselhan, line 71: suggestion: Reverse these calls so that if showing the banner becomes true is only does so after the message is updated.

alcs-frontend/src/app/features/admin/configuration/configuration.component.scss
sandratoh: I think we didn't do much CSS when there were no designs for admin features. This one actually has a Figma design so I styled it according to that

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
sandratoh: Updated!

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
sandratoh: Updated!

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
sandratoh: Updated!

alcs-frontend/src/app/features/admin/configuration/configuration.component.scss
sandratoh: Removed some more unused styling

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
sandratoh: Updated!

dhaselhan (approved):
LGTM 👍

alcs-frontend/src/app/features/admin/configuration/configuration.component.ts
sandratoh: Added missing migration file and then updated method to throw error

dhaselhan (approved):
LGTM 👍

PR #1593, opened 2024-04-09
trim redundant space, update DistrictOfBarriere applications

mhuseinov:
- remove redundant spaces in local government names
- update District Of Barriere applications and submissions


dhaselhan (approved):
LGTM 👍

PR #1595, opened 2024-04-10
Caches users in Redis

dhaselhan:
* Cache users in redis for loading them for each request

### DIFF
--- a/services/apps/alcs/src/common/authorization/roles-guard.service.ts
+++ b/services/apps/alcs/src/common/authorization/roles-guard.service.ts
@@ -1,3 +1,4 @@
+import { RedisService } from '@app/common/redis/redis.service';
 import {
   CanActivate,
   ExecutionContext,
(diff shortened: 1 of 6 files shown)


services/apps/alcs/src/common/authorization/roles-guard.service.ts
```diff
@@ -87,10 +89,46 @@ export class RolesGuard implements CanActivate {
...
+    );
+    if (idirUser) {
+      return JSON.parse(idirUser);
+    }
+
+    const user = await this.userService.getByGuid(userGuids);
+    if (user) {
+      const serialized = JSON.stringify(user);
+      if (user.bceidGuid) {
+        await redisClient.setEx(
```
mhuseinov, line 119: should we manually clean the user on logout, just in case?

mhuseinov (approved):
👍

PR #1596, opened 2024-04-10
Bugfix/alcs 1826 - planning review advanced search

mhuseinov:
Advanced search
- fix search by planning review type
- fix sorting and paging
- fix navigation to planning review from advanced search result

### DIFF
--- a/services/apps/alcs/src/alcs/search/planning-review/planning-review-advanced-search.service.ts
+++ b/services/apps/alcs/src/alcs/search/planning-review/planning-review-advanced-search.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-import { Repository, SelectQueryBuilder } from 'typeorm';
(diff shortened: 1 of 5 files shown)


services/apps/alcs/src/alcs/search/planning-review/planning-review-advanced-search.service.ts
```diff
@@ -100,6 +100,8 @@ export class PlanningReviewAdvancedSearchService {
     query = this.compileSearchByNameQuery(searchDto, query);
     query = this.compileDecisionSearchQuery(searchDto, query);
     query = this.compileDateRangeSearchQuery(searchDto, query);
+    query = this.compileDateRangeSearchQuery(searchDto, query);
```
trslater, line 103: This looks like an exact copy of the line above? Is this intentional?

mhuseinov: good catch, probably a merge conflict, fixed

trslater (approved):
LGTM 🚀

PR #1600, opened 2024-04-15
replace "more actions" button with "view details"

mhuseinov:
- replace "more actions" button with "view details" on application and noi


trslater (approved):
LGTM 🚀

PR #1601, opened 2024-04-15
Feature/alcs 1691 pid pin UI

mhuseinov:
- pid pin search ui improvement
- unit tests
- remove redundant logs from refresh token


trslater (approved):
LGTM 🚀

PR #1602, opened 2024-04-15
add field labels on textarea fields in decision

mhuseinov:
added missing field labels on textarea in decision component for application and noi


dhaselhan (approved):
LGTM 👍

PR #1603, opened 2024-04-16
Change Additional Structures form to use ID instead of Index

dhaselhan:
* Re-enable remove from any position
* Switch everything from index to ID
* Update tests


mhuseinov (approved):
👍

PR #1606, opened 2024-04-17
fix migration & entity issues with documents, inquiries & planning review

lstod:
- Added Unique to entity
- Inquiry and inquiry-parcel entities were swapped
- fix inquiry_document typo to inquiry_parcel and create new migration

### DIFF
--- a/services/apps/alcs/src/alcs/inquiry/inquiry-document/inquiry-document.entity.ts
+++ b/services/apps/alcs/src/alcs/inquiry/inquiry-document/inquiry-document.entity.ts
@@ -8,11 +8,13 @@ import {
   ManyToOne,
   OneToOne,
   PrimaryGeneratedColumn,
+  Unique,
 } from 'typeorm';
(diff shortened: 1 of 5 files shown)


services/apps/alcs/src/alcs/inquiry/inquiry-document/inquiry-document.entity.ts
```diff
@@ -8,11 +8,13 @@ import {
   ManyToOne,
   OneToOne,
   PrimaryGeneratedColumn,
+  Unique,
 } from 'typeorm';
 import { DocumentCode } from '../../../document/document-code.entity';
 import { Document } from '../../../document/document.entity';
 import { Inquiry } from '../inquiry.entity';
 
+@Unique('UNIQUE_OATS_ISSUE_IDS', ['oatsDocumentId', 'oatsIssueId'])
```
mhuseinov, line 17: Will this break regular document uploads from ALCS since we are not populating these values?

mhuseinov (approved):
👍

PR #1607, opened 2024-04-17
When checking document access only check NON draft submissions

dhaselhan:
* For ALCS users add a bypass that any user with ALCS roles is allowed to view the application to allow them to see documents attached to drafts that were copied over


trslater (approved):
LGTM 🚀

PR #1608, opened 2024-04-18
add check if columns already exist in DB

lstod:
Migration to fix double column adding issue due to typo in drop statement of a now deleted migration


trslater (approved):
LGTM 🚀

PR #1609, opened 2024-04-18
Obfuscate Planning Review & Inquiries

lstod:
- Overwrite PII and other sensitive info
- add obfuscate back into migrate function
-  did not touch oats, appears not needed as it is being handled

### DIFF
--- a/bin/migrate-oats-data/prod_db_obfuscation/alcs/inquiry/inquiry.py
+++ b/bin/migrate-oats-data/prod_db_obfuscation/alcs/inquiry/inquiry.py
@@ -1,0 +1,83 @@
+from common import setup_and_get_logger
+from db import inject_conn_pool
+from ...utils import log_process, get_update_column_query
+from faker import Faker
+
+logger = setup_and_get_logger("prod_data_obfuscation")
+
+
+def obfuscate_inquiry_related_data():
+    _update_inquiries()
+    _update_inquiry_parcel()
+
+
+@log_process(logger, "update_with_random_alcs_inquiries")
+@inject_conn_pool
+def _update_inquiries(conn=None):
+    fake = Faker()
+    fake_latin = Faker("la")
+
+    try:
+        with conn.cursor() as cursor:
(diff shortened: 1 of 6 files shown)


bin/migrate-oats-data/prod_db_obfuscation/alcs/planning_review/planning_review.py
```diff
@@ -0,0 +1,136 @@
...
+
+    try:
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT uuid FROM alcs.planning_review_decision")
+            record_ids = cursor.fetchall()
+
+            for r_id in record_ids:
+                sql_query = f"""
+                            UPDATE alcs.planning_review_decision 
+                            SET {get_update_column_query("decision_description")}
```
mhuseinov, line 92: the same as above. In general, if there are nulls in non-obfuscated data, try to keep all nulls as is. Otherwise, it could lead to missing bugs in the finished product with actual data from production

bin/migrate-oats-data/prod_db_obfuscation/alcs/inquiry/inquiry.py
```diff
@@ -0,0 +1,238 @@
...
+        with conn.cursor() as cursor:
+            cursor.execute("SELECT uuid FROM alcs.inquiry")
+            record_ids = cursor.fetchall()
+
+            for r_id in record_ids:
+                sql_query = f"""
+                            UPDATE alcs.inquiry 
+                            SET 
+                            {get_update_column_query("summary")}
+                            WHERE uuid = %s AND summary IS NOT NULL;
```
mhuseinov, line 35: I can see that you've opted for splitting the update into multiple queries, but for the future, you could also look into this: "column1 = CASE WHEN column1 IS NOT NULL THEN new_value1 ELSE column1 END," It will achieve the same result and a bit easier to read.

mhuseinov (approved):
👍

PR #1610, opened 2024-04-22
Remove erroneous f

dhaselhan:
* Remove f from primary contact section


trslater:
I only see deletions. Were there meant to be replacement fixed templates?

dhaselhan:
> I only see deletions. Were there meant to be replacement fixed templates?

I'm removing an f from the files that should not be there, thats why each file goes down 5 bytes. I am not deleting the templates.

trslater:
> > I only see deletions. Were there meant to be replacement fixed templates?
> 
> I'm removing an f from the files that should not be there, thats why each file goes down some bytes. I am not deleting the templates.

OK. The GitHub UI is not that clear, but it looks like they're being deleted, but obviously, you would know. Just thought I'd double-check.

dhaselhan:
> OK. The GitHub UI is not that clear, but it looks like they're being deleted, but obviously, you would know. Just thought I'd double-check.

If you look on the left side of files you can see orange marks (edited) instead of red (deleted).
image

PR #1614, opened 2024-04-22
Add interceptor to redirect to login on 401

dhaselhan:
* Clean up existing interceptor to be more interceptor, less service


mhuseinov (approved):
👍

PR #1616, opened 2024-04-22
Show commissioner flag when applicant is present

dhaselhan:
* For some reason there was code hiding C flag if applicant was also checked
* Add missing space to , after G


trslater (approved):
LGTM 🚀

PR #1617, opened 2024-04-23
Search Performance Improvements

dhaselhan:
* Parallelize all searches to happen at the same time
* Remove unnecessary join for type (was already added in mapping layer)
* Add debug performance timing


trslater (approved):
LGTM 🚀

PR #1618, opened 2024-04-23
Downgrade NGX Mask to avoid input Issues

dhaselhan:
* Currrently open issue breaking NGX Mask when it has decimal separators https://github.com/JsDaddy/ngx-mask/issues/1349
* Downgrade to 17.0.4 which does not have the issue


trslater (approved):
LGTM 🚀

PR #1619, opened 2024-04-23
Change inline number input value to check for null

dhaselhan:
'0' evalutes to false so ngIf=value on '0' returns false. Check for null directly instead.


mhuseinov (approved):
👍

PR #1620, opened 2024-04-24
Add .pdf to updated submissions missing the extension

dhaselhan:


services/apps/alcs/src/providers/typeorm/migrations/1713977117461-fix_subd_pdfs.ts
```diff
@@ -0,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class FixSubdPdfs1713977117461 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      UPDATE "alcs"."document"
+      SET "file_name" = CASE
+        WHEN "file_name" NOT LIKE '%.pdf' THEN CONCAT("file_name", '.pdf')
```
mhuseinov, line 8: is it possible that the file type is something other than PDF?

dhaselhan: Good thinking, update the SQL to also only apply to pdf mime type

mhuseinov (approved):
👍

PR #1621, opened 2024-04-24
Compare rounded lot vs. parcel size to avoid floating non-equality

trslater:
- Floating point error can cause lot total to register as different than parcel size when they should be equal
- Rounding to the 5 decimals allowed by UI to account for thi


dhaselhan (approved):
LGTM 👍 
Should we add a unit test?

trslater:
> LGTM 👍 Should we add a unit test?

Probably a good idea. I'll do that.

PR #1627, opened 2024-04-25
Feature/alcs 1815 - update local government on NOI

mhuseinov:
- update local government on NOI submission update
- send email if local government changed


trslater (approved):
LGTM 🚀

PR #1628, opened 2024-04-25
Move logger to Winston, add s3 stream

dhaselhan:
* Move from Pino -> Winston for easier s3 stream transport setup
* Add s3 stream to store logs in s3 stream bucket
* Configure stream settings to rotate logs daily


mhuseinov (approved):
👍

PR #1630, opened 2024-04-25
improve get submission for noi and application

mhuseinov:
- split retrieval of submission and owner into 2 queries 

The step above led to performance improvement from 23 seconds to 0.23 seconds on submisisonService.get(...


dhaselhan (approved):
LGMT 👍

services/apps/alcs/src/alcs/application/application-submission/application-submission.service.ts
```diff
@@ -25,23 +25,33 @@ export class ApplicationSubmissionService {
...
+            },
           },
         },
-        owners: {
-          type: true,
-        },
+      },
+    );
+
+    submission.owners = await this.applicationOwnerRepository.find({
```
dhaselhan, line 47: nitpick: Add a comment mentioning this is done separately for performance reasons

PR #1638, opened 2024-04-30
Add Region dropdown to SRW Intake Page

dhaselhan:


alcs-frontend/src/app/features/notification/intake/intake.component.ts
```diff
@@ -101,6 +119,17 @@ export class IntakeComponent implements OnInit {
       });
   }
 
+  async updateRegion($event: string | string[] | null) {
```
trslater, line 122: Why does this function accept an array `$event` if there is a condition to prevent anything from happening if `$event` is an array?

dhaselhan: The inline-dropdown component supports multi select, so its interface is string | string[] | null. The check exists to make Typescript happy that I am sure it is not a string[].

trslater: Gotcha.

trslater (approved):
LGTM 🚀

PR #1640, opened 2024-04-30
Fix File Type on Public Search

dhaselhan:
* Was broken for all types!


services/apps/alcs/src/portal/public/search/notice-of-intent/public-notice-of-intent-search.service.ts
```diff
@@ -358,4 +362,18 @@ export class PublicNoticeOfIntentSearchService {
 
     promises.push(query.getMany());
   }
+
+  private addFileTypeResults(
+    searchDto: SearchRequestDto,
+    promises: Promise<{ fileNumber: string }[]>[],
+  ) {
+    if (searchDto.fileTypes.includes('NOI')) {
```
mhuseinov, line 370: nitpick: you could use enum for types

mhuseinov (approved):
👍

PR #1642, opened 2024-04-30
Move snowplow to its own script

dhaselhan:


portal-frontend/src/assets/snowplow.js
```diff
@@ -0,0 +1,31 @@
...
+  platform: "web",
+  post: true,
+  forceSecureTracker: true,
+  contexts: {
+    webPage: true, performanceTiming: true
+  }
+});
+window.snowplow("enableActivityTracking", 30, 30); // Ping every 30 seconds after 30 seconds
+window.snowplow("enableLinkClickTracking");
+window.snowplow("trackPageView");
```
trslater, line 30: Doesn't the service in your previous already call `trackPageView`? Why do we need it here as well?

dhaselhan: This was the code given to us by the snowplow team to paste into the index file. I am not sure.

dhaselhan: Removed

trslater (approved):
LGTM 🚀

PR #1644, opened 2024-05-01
Add connect-src for snowplow, fix main docker

dhaselhan:


services/config/default.json
```diff
@@ -61,12 +63,14 @@
...
     "PORT": "6379",
-    "PASSWORD": ""
+    "PASSWORD": "redis"
   },
   "EMAIL": {
-    "DEFAULT_ADMINS": [""]
+    "DEFAULT_ADMINS": [
+      ""
+    ]
   },
```
mhuseinov, line 74: These changes could break someone's local setup. Not an issue to fix but it could be avoided

dhaselhan: I'll make a post in #dev before I merge

portal-frontend/Dockerfile
```diff
@@ -13,6 +13,8 @@ RUN npm ci
 # Copy the source code to the /app directory
 COPY . .
 
+ENV NODE_OPTIONS="--max-old-space-size=2048"
```
mhuseinov, line 16: did you run out of heap size?

mhuseinov (approved):
👍

portal-frontend/Dockerfile
dhaselhan: yeah, had to give docker 10gb or Ram even with this set

PR #1652, opened 2024-05-06
Feature/alcs 1856 reset lfng statuses on ETL applications

mhuseinov:
- reset lfng statuses to the most recent one from OATS

Note: 
The reset ETL will not be a part of the import. It will be a standalone action.


trslater (approved):
LGTM 🚀

PR #1653, opened 2024-05-06
Do not count planning referrals without cards for PR header

dhaselhan:
* Prevents showing go to card when the referral has no card


sandratoh (approved):
LGTM 👍

PR #1657, opened 2024-05-06
Retain file type when renaming file

dhaselhan:
* This retains the type so not everything it octet-stream
* Code was already in place for applications


mhuseinov (approved):
👍

PR #1660, opened 2024-05-07
Fix empty gov email causing emails to fail and success toast on fail

trslater:
- Empty strings as emails were causing SRW government emails to fail
- Return email sent status to resend endpoint
- Fail resend endpoint if email not sent

### DIFF
--- a/services/apps/alcs/src/portal/notification-submission/notification-submission.service.ts
+++ b/services/apps/alcs/src/portal/notification-submission/notification-submission.service.ts
@@ -470,7 +470,7 @@ export class NotificationSubmissionService {
       );
 
       if (localGovernment && localGovernment.emails) {
-        ccEmails = localGovernment.emails;
+        ccEmails = localGovernment.emails.filter((email) => email !== '');
       }
     }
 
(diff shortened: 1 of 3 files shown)


services/apps/alcs/src/portal/notification-submission/notification-submission.service.ts
```diff
@@ -406,18 +406,19 @@ export class NotificationSubmissionService {
...
+    const didSend =
+      (await this.emailService.sendEmail({
+        to: [templateData.to],
+        body: templateData.html,
+        subject: `Agricultural Land Commission SRW${submission.fileNumber} (${submission.applicant})`,
+        parentType: PARENT_TYPE.NOTIFICATION,
+        parentId: templateData.parentId,
+        cc: templateData.cc,
+        attachments: [document.document],
+      })) ?? false;
```
dhaselhan, line 421: Rather than use nullish coalescing, you can fix the email service by changing return; to return false; on line 137 of email.service.ts.

dhaselhan (approved):
LGTM 👍

PR #1668, opened 2024-05-08
backup table and RSL

mhuseinov:
- introduce a backup table for lfng reset status
- include RSL status into ETL
NOTE:
the migration script is the same query as in ETL


dhaselhan (approved):
LGTM 👍

PR #1670, opened 2024-05-08
Allow statuses to come in later than results

dhaselhan:
* The input setters are called in any order, this results that even if all of the inputs are populated in the parent they may not be populated in the child yet
* Add logic so that when either the statuses or results are set, we re-hydrated the results


mhuseinov (approved):
👍

PR #1673, opened 2024-05-09
NOI POFO Hotfix/alcs 1930 - read correct fill type from DB, read soil project duration in template

trslater:
- Read `soil_project_duration` column in PDF template instead of `fill_project_duration`
- Read `soil_fill_type_to_place` column in PDF template instead of `soil_type_removed`


dhaselhan:
If you are changing where this variable is stored, are ALCS and all the other view places up to date already to use this column?

dhaselhan (approved):
LGTM 👍

PR #1674, opened 2024-05-09
Send one email per decision

dhaselhan:
* Move decision emails for Apps/NOIs to their own scheduler/queues
* Add new columns to decisions to track emails


services/apps/alcs/src/queues/scheduler/scheduler.service.ts
```diff
@@ -9,16 +9,15 @@ export const MONDAY_TO_FRIDAY_AT_2AM_PDT_IN_UTC = '0 0 9 * * 1-5';
 export const EVERYDAY_MIDNIGHT_PST_IN_UTC = '0 0 8 * * *';
```
trslater, line 9: Does this it runs every month on the 8th at midnight?

services/apps/alcs/src/queues/scheduler/scheduler.service.ts
```diff
@@ -9,16 +9,15 @@ export const MONDAY_TO_FRIDAY_AT_2AM_PDT_IN_UTC = '0 0 9 * * 1-5';
...
 export const EVERYDAY_MIDNIGHT_PDT_IN_UTC = '0 0 7 * * *';
 
-export const EVERY_15_MINUTES_STARTING_FROM_8AM_PST_IN_UTC =
-  '0/15 16-23,0-6 * * *';
-export const EVERY_15_MINUTES_STARTING_FROM_8AM_PDT_IN_UTC =
-  '0/15 15-23,0-5 * * *';
+export const EVERY_15_MINUTES_STARTING_FROM_8AM_PST_IN_UTC = '* * * * *';
```
trslater, line 12: This will run every minute now. Is that intended?

services/apps/alcs/src/queues/scheduler/scheduler.service.spec.ts
```diff
@@ -92,7 +103,7 @@ describe('SchedulerService', () => {
   //     {
   //       repeat: {
   //         pattern: isDST()
-  //           ? MONDAY_TO_FRIDAY_AT_2AM_PDT_IN_UTC 
+  //           ? MONDAY_TO_FRIDAY_AT_2AM_PDT_IN_UTC
   //           : MONDAY_TO_FRIDAY_AT_2AM_PST_IN_UTC,
   //       },
   //     },
```
trslater, line 109: Maybe we should remove these comments?

dhaselhan: That's for a job we built but disabled as per Ansleys request. This test covers the disabled job.
  //Job Disabled for now

trslater (approved):
LGTM 🚀

services/apps/alcs/src/queues/scheduler/scheduler.service.ts
trslater: I just noticed that * is Shift + 8. I wonder if this was meant to be `0 0 * * * *`, but the shift was missed by accident?

PR #1677, opened 2024-05-13
UI Improvements on Public Search

dhaselhan:
* Add close button to all controls
* Update type dropdown to reset typed values unless they are valid and clear filter when blurring


trslater (approved):
LGTM 🚀

PR #1678, opened 2024-05-13
Disable close on click-outside globally for portal

trslater:
- Also cleaned up everywhere `disableClose` is set for individual dialogs


dhaselhan (approved):
LGTM 👍

PR #1680, opened 2024-05-13
Add CC emails options

dhaselhan:
* Add new UI form input to both NOI and Application Decision Release Modal
* Save new CC emails to DB
* Use CC emails when triggering the decision email


trslater (approved):
LGTM 🚀

PR #1682, opened 2024-05-14
ALCS Advanced Search UI Reorg

dhaselhan:
* Re-organize components
* Remove View more section
* Change date picker to open on click and prevent typing

### DIFF
--- a/alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.html
+++ b/alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.html
@@ -13,7 +13,16 @@
     onkeydown="return false;"
     [formControl]="componentTypeControl"
   />
-  <button matSuffix *ngIf="hasItemsSelected" class="clear-button" type="button" (click)="clear()">×
+  <button
+    matSuffix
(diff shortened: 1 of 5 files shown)


alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.html
```diff
@@ -1,13 +1,21 @@
...
     value="{{ getSelectedItems() }}"
-    placeholder="File Type"
+    id="type"
     type="text"
     matInput
     [matAutocomplete]="auto"
     onkeydown="return false;"
     [formControl]="componentTypeControl"
   />
+  <button matSuffix *ngIf="hasItemsSelected" class="clear-button" type="button" (click)="clear()">×
```
trslater, line 16: In a previous PR, you used `<mat-icon>close</mat-icon>` for an 'x' button. Should we do the same here for consistency?

dhaselhan: Good suggestion! Not sure why I didn't. Updated

trslater (approved):
LGTM 🚀

PR #1683, opened 2024-05-14
ALCS Search Fixes

dhaselhan:
* Fix name queries being terribly wrong, must not have tested this myself
* Change variable names for date_submitted so they do not conflict


trslater (approved):
LGTM 🚀

PR #1685, opened 2024-05-14
Improve public search decision date picker

trslater:
- Split datepicker
- Remove placeholder
- Limit dates
- Open picker by clicking anywhere on field
- Prevent direct editing of date text input


mhuseinov (approved):
👍

PR #1686, opened 2024-05-14
Change Portal Status to Multi Select

dhaselhan:
* Generalize the file type filter dropdown to take in more inputs to be re-usable
* Add a new data source for portal status types according to Ticket
* Switch existing component to use a combo of the above

### DIFF
--- a/alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.ts
+++ b/alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.ts
@@ -1,6 +1,6 @@
 import { SelectionModel } from '@angular/cdk/collections';
 import { FlatTreeControl } from '@angular/cdk/tree';
-import { AfterViewInit, Component, EventEmitter, Output } from '@angular/core';
+import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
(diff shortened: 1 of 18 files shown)


alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.ts
```diff
@@ -34,14 +34,29 @@ export class FileTypeFilterDropDownComponent {
...
 
-    fileTypeData.dataChange.subscribe((data) => {
+  ngOnInit(): void {
+    this.fileTypeData.dataChange.subscribe((data) => {
       this.dataSource.data = data;
     });
+
+    if (this.preExpanded.length > 0) {
+      const nodes = this.treeControl.dataNodes.filter((node) => this.preExpanded.includes(node.item.label));
+      debugger;
```
mhuseinov, line 55: looks like a debug leftover

mhuseinov (approved):
👍

alcs-frontend/src/app/features/search/file-type-filter-drop-down/file-type-filter-drop-down.component.ts
dhaselhan: Removed

PR #1688, opened 2024-05-15
Allow removal of subtypes from submission

trslater:


alcs-frontend/src/app/features/notice-of-intent/proposal/preparation/preparation.component.ts
```diff
@@ -65,69 +65,22 @@ export class PreparationComponent implements OnInit {
...
-    }
-
-    return false;
-  }
-
-  async saveSubtypes($event: string | string[] | null) {
-    if (this._noticeOfIntent && $event instanceof Array) {
-      const selectedCodes = $event.map((label) => this.subtypes.find((subtype) => subtype.label === label)!.code);
+  async saveSubtypes(selectedLabels: string | string[] | null) {
+    console.log(selectedLabels);
```
mhuseinov, line 78: looks like a debug leftover

alcs-frontend/src/app/features/notice-of-intent/proposal/preparation/preparation.component.ts
```diff
@@ -59,75 +59,24 @@ export class PreparationComponent implements OnInit {
...
       this.selectableTypes = subtypes.map((subtype) => {
-        let disabled = false;
-
-        disabled = this.isSubtypeDisabled(submission, subtype);
-
         return {
           label: subtype.label,
           value: subtype.label,
-          disabled,
+          disabled: false,
```
mhuseinov, line 69: I think disabled is optional parameter so you could simply ignore it here and it should act as false

trslater: Yeah, it wasn't for `selectableTypes`, but this was just a placeholder type to pass to the dropdown where it is option, so I just removed the flag completely from this component.

mhuseinov (approved):
👍

PR #1689, opened 2024-05-15
Fix alignment of close icons

dhaselhan:
* Move close to be after date picker icon to be consistent with other fields


mhuseinov (approved):
👍

PR #1690, opened 2024-05-15
Update ALCS Labels

dhaselhan:
* Unify parcel labels to be like the portal
* Add missing ha
* Show ha/m on already filled/already placed table as 0 is valid


sandratoh (approved):
LGTM 👍

PR #1692, opened 2024-05-15
Change referral to use text area for two fields

dhaselhan:
* Update text area so it wraps text
* Move edit button to after the text instead of to the right of it


sandratoh (approved):
LGTM 👍

PR #1694, opened 2024-05-16
exclude ora2pg and openshift file from code coverage

mhuseinov:
exclude ora2pg and openshift from code climate report


dhaselhan (approved):
LGTM 👍

PR #1695, opened 2024-05-21
Inbox Search and Code DRY up

dhaselhan:
* move inbox search to new pattern
* DRY up code


services/apps/alcs/src/utils/search/search-intersection.ts
```diff
@@ -0,0 +1,18 @@
...
+) => {
+  const queryResults = await Promise.all(promises);
+
+  const allIds: Set<string>[] = [];
+  for (const result of queryResults) {
+    const fileNumbers = new Set<string>();
+    result.forEach((currentValue) => {
+      fileNumbers.add(currentValue.fileNumber);
+    });
+    allIds.push(fileNumbers);
```
sandratoh, line 14: How come `fileNumbers` uses `.add` and `allIds` uses `.push` when they're both `Set`, is it just because we already initiated an empty set for `allIds`?

services/apps/alcs/src/alcs/search/application/application-advanced-search.service.ts
```diff
@@ -121,39 +118,64 @@ export class ApplicationAdvancedSearchService {
...
-      this.addFileNumberResults(searchDto, promises);
+      const promise = APP_SEARCH_FILTERS.addFileNumberResults(
+        searchDto,
+        this.applicationRepository,
+      );
+      promises.push(promise);
     }
 
     if (searchDto.legacyId) {
       this.addLegacyIDResults(searchDto, promises);
```
sandratoh, line 129: Can you explain again why certain filters aren't being added to APP_SEARCH_FILTERS? Eg: legacyId, regionCode, decisionDate, etc

services/apps/alcs/src/portal/inbox/application/inbox-application.service.spec.ts
```diff
@@ -72,16 +98,26 @@ describe('InboxApplicationService', () => {
...
-    const compileApplicationSearchQuerySpy = jest
-      .spyOn(service as any, 'compileApplicationSearchQuery')
-      .mockResolvedValue(mockQuery);
+  it('should call searchForFileNumbers method correctly', async () => {
+    mockApplicationSubmissionSearchViewRepository.createQueryBuilder.mockReturnValue(
+      mockQuery,
+    );
+
+    const searchForFileNumbersSpy = jest
+      .spyOn(service as any, 'searchForFileNumbers')
```
sandratoh, line 116: Just curious: When do we use `.spyOn` and when do we just use the default mock?

services/apps/alcs/src/utils/search/notice-of-intent-search-filters.ts
```diff
@@ -0,0 +1,138 @@
...
+  addFileTypeResults: (
+    searchDto: InboxRequestDto,
+    noiRepository: Repository<NoticeOfIntent>,
+  ) => {
+    return noiRepository.find({
+      select: {
+        fileNumber: true,
+      },
+    });
+  },
```
sandratoh, line 137: NOI also has decision components, should this be more like `APP_SEARCH_FILTERS.addFileTypeResults`?

services/apps/alcs/src/utils/search/search-intersection.ts
dhaselhan: `allIds` is an Array of Sets which is why it uses `push`

services/apps/alcs/src/alcs/search/application/application-advanced-search.service.ts
dhaselhan: I only moved filters that are shared, legacyId is only searchable in ALCS for example so I left it in the alcs search service.

services/apps/alcs/src/utils/search/notice-of-intent-search-filters.ts
dhaselhan: I don't know, this ticket was not about changing functionality only improving performance / maintainability. This filter works the same as before.

services/apps/alcs/src/portal/inbox/application/inbox-application.service.spec.ts
dhaselhan: We are using spyOn here since we want to mock a single method on the class we area testing. spyOn is used for mocking a single method, createMock is when you are mocking the entire class.

services/apps/alcs/src/utils/search/search-intersection.ts
sandratoh: Ooh I see it now, thanks!

sandratoh (approved):
LGTM 👍

PR #1709, opened 2024-05-23
Add sorting to lots of queries

dhaselhan:
* Sort owners by first name
* Sort transferees by first name
* Sort card subtasks by created date


sandratoh (approved):
LGTM 👍

PR #1711, opened 2024-05-23
Keep organization name when changing from indv -> organization

dhaselhan:
* Also fixed a bug where the corporate summary was not getting deleted

### DIFF
--- a/services/apps/alcs/src/portal/application-submission/application-owner/application-owner.service.ts
+++ b/services/apps/alcs/src/portal/application-submission/application-owner/application-owner.service.ts
@@ -138,22 +138,30 @@ export class ApplicationOwnerService {
       });
     }
 
-    //If attaching new document and old one was defined, delete it
+    //If null or new document, delete old one
     if (
-      updateDto.corporateSummaryUuid &&
+      updateDto.corporateSummaryUuid !== undefined &&
       existingOwner.corporateSummaryUuid !== updateDto.corporateSummaryUuid &&
       existingOwner.corporateSummary
     ) {
-      const oldSummary = existingOwner.corporateSummary;
+      await this.applicationDocumentService.delete(
+        existingOwner.corporateSummary,
+      );
       existingOwner.corporateSummary = null;
-      await this.repository.save(existingOwner);
-      await this.applicationDocumentService.delete(oldSummary);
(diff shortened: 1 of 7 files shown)


services/apps/alcs/src/portal/application-submission/application-owner/application-owner.service.ts
```diff
@@ -138,22 +138,30 @@ export class ApplicationOwnerService {
...
     }
 
-    existingOwner.corporateSummaryUuid =
-      updateDto.corporateSummaryUuid !== undefined
-        ? updateDto.corporateSummaryUuid
-        : existingOwner.corporateSummaryUuid;
+    //Attach new one
+    if (
+      updateDto.corporateSummaryUuid &&
+      updateDto.corporateSummaryUuid !== existingOwner.corporateSummaryUuid
```
sandratoh, line 157: Wouldn't the previous `if` block change `existingOwner.corporateSummaryUuid` to be `null`, so this will always be `true`? Is it intentional to mutate it before this condition?

services/apps/alcs/src/portal/notice-of-intent-submission/notice-of-intent-owner/notice-of-intent-owner.service.ts
```diff
@@ -153,22 +153,30 @@ export class NoticeOfIntentOwnerService {
...
     }
 
-    existingOwner.corporateSummaryUuid =
-      updateDto.corporateSummaryUuid !== undefined
-        ? updateDto.corporateSummaryUuid
-        : existingOwner.corporateSummaryUuid;
+    //Attach new one
+    if (
+      updateDto.corporateSummaryUuid &&
+      updateDto.corporateSummaryUuid !== existingOwner.corporateSummaryUuid
```
sandratoh, line 172: Same thing for NOI

services/apps/alcs/src/portal/application-submission/application-owner/application-owner.service.ts
dhaselhan: Had call with Sandra to explain

sandratoh (approved):
LGTM 👍

PR #1714, opened 2024-05-27
Map legacy_id to inquiries

lstod:
- verified matching legacy_id to issue #
- only update if oats field is not null
- does not overwrite existing alcs value
- only changes oats_etl entries


mhuseinov (approved):
👍

PR #1715, opened 2024-05-27
Bug Fixes

dhaselhan:
* Reset form was clearing the status codes instead of setting them back to [], so manually specify []
* Add search for file number back into inbox search, not sure why I thought inbox didn't have it


trslater (approved):
LGTM 🚀

PR #1717, opened 2024-05-27
Add padding to bottom of Evidentiary Record table

Abradat:
Added padding to the bottom of the Evidentiary Record table in the Commissioner view


dhaselhan (approved):
LGTM 👍

PR #1718, opened 2024-05-28
Application, NOI and SRW document description mapping

lstod:
- Populate optional description field
- SRW was done already but migration will still run and catch anything was missed
- Does not touch inquiry or planning review


mhuseinov (approved):
👍

PR #1719, opened 2024-05-28
set missing LGs on application and submission

mhuseinov:
set missing LG uuid for ALCS application and application submission based on the names of LG in OATS


lstod (approved):
LGTM 🏅

PR #1725, opened 2024-05-29
Return Owner with Corporate Summary on Save/Updated

dhaselhan:
Looks like Application / NOI now work differently and Application uses the returned owner. The returned owner was not being loaded with their corporate summary resulting in it showing as Not Applicable.


sandratoh (approved):
LGTM 👍

PR #1731, opened 2024-05-29
Use lowest decision date as minimum instead of highest

dhaselhan:
* This means subsequent decisions will have to come after the first rather than requiring each decision to come after the ones already done


trslater (approved):
LGTM 🚀

PR #1734, opened 2024-05-30
Fixes for changing Indv/Org in Dialogs

dhaselhan:
* Update transferee to not clear org name
* Update owners to properly clear files when switching to indv


trslater (approved):
LGTM 🚀

PR #1735, opened 2024-05-30
Bugfix/alcs 1988 alr area component on decisions

mhuseinov:
- do not reset all component fields on inline update of alr are in NOI
- update precision from 2 to 5


dhaselhan (approved):
LGTM 👍

PR #1740, opened 2024-06-04
Fix ALCS UI/UX Bugs

Abradat:
Following UI/UX bugs have been fixed:

- ng-select dropdown border, placeholder, and arrow colour fixed to ALCS primary instead of blue
- Fixed cursor remaining in the pointer state after clicking on proposal summary's inline text area


trslater (approved):
LGTM 🚀 

`!important` is not my favourite, but it does seem like it is necessary when battling Angular Material.

PR #1742, opened 2024-06-05
Automap `wasReleased` to get value in frontend

trslater:
Automap `wasReleased` from the decision entity to
the DTO when responding to API requests for decisions,
so that the frontend can read that value.


Abradat (approved):
LGTM!

PR #1749, opened 2024-06-06
Fix NOI advanced search "submitted to" filter not working properly

Abradat:
In NOIs Advanced search, the "Submitted to ALC - End Date" field used the beginning of the mentioned date (12:00 AM) as a filter. For example, If the AOI was submitted on June 6th, 2:00 PM, it would be after 12:00 AM so it would not be in results. To fix, the filter is now set to the beginning of the next day.


trslater (approved):
LGTM 🚀

PR #1750, opened 2024-06-06
Fix corporate summary not being discarded after changing in edit

Abradat:
Fix corporate summary not being discarded after changing from organization to individual on edit dialog


trslater (approved):
LGTM 🚀

PR #1751, opened 2024-06-07
Backport: mostly Jun 5 regression hotfixes

trslater:
- Hotfixes (test) from Jun 5 test cut regression
- A couple small changes from May 30


Abradat (approved):
LGTM!

PR #1753, opened 2024-06-10
ALCS board, comment, staff journal UI tweaks

trslater:
- Remove underline from mention list items
- Change comment box display to outline
- Add space between input and buttons
- Fix sticky board header
- Change staff journal input to outline


Abradat (approved):
LGTM!

PR #1754, opened 2024-06-11
Reorder condition for initial contact selection

trslater:
By putting government user check first, it only
checks the parcel owners length if not a government
user.


Abradat (approved):
LGTM!

PR #1758, opened 2024-06-11
Fix Homepage UI/UX Bugs

Abradat:
- Added '-' placeholder to non-retroactive NOIs in the "Cards Assigned to Me" section
- Adjusted columns' size for the "type" column to have multiple pills. (Increase type and decreased workflow stage widths)


trslater (approved):
LGTM 🚀

PR #1759, opened 2024-06-13
Fix UI/UX Bugs for Application and NOI Info Request

Abradat:
Following changes have been made for applications and NOIs Information Request dialog:

- Aligned text input with other radio button
- Changed the appearance of "Reason" section from filled to outline


trslater (approved):
LGTM 🚀

PR #1762, opened 2024-06-14
Fix NOI Decision Component Empty Fields Errors

Abradat:
Fixed 500 server error on submitting decision component with cleared fields which had value before for NOIs and Applications.

The error was on the ALCS-frontend that instead of sending null values for empty fields used to sent '' value which caused 500 error on the server side.


trslater (approved):
LGTM 🚀

PR #1767, opened 2024-06-18
Restyle all portal/public tabs

trslater:
Styled:
- Inbox results table
- Public search results table
- View submission


Abradat (approved):
LGTM!

PR #1768, opened 2024-06-19
Fix Application Decision Component PATCH not working for null values

Abradat:
Fix Application Decision Components' fields not being updated if the new values are null.


trslater (approved):
LGTM 🚀

PR #1769, opened 2024-06-20
Fix padding around tabs

trslater:
- **Clean up redundant tab formatting**
- **Set correct sizes and weight for all tabs**
- **Add padding between view page return button and tabs**
- **Add padding between tabs and tab content**
- **Fix padding around public search results heading**


Abradat (approved):
LGTM!

PR #1770, opened 2024-06-24
Fix E2E Test for TUR Failing

Abradat:
- Fixed the map area not being able to validate the text
- Fixed the calendar button falling out of viewport and failing to scroll
- Config changes: 
    - retries for local environment 
    - timeout and action timeout for stability
    - number of workers to be 1 on local env


trslater (approved):
LGTM 🚀

PR #1771, opened 2024-06-25
Fix notification trigger statuses

trslater:
- **Set trigger status for notification email statuses**
- **Update existing data with SUBM trigger status**


Abradat (approved):
LGTM!

PR #1775, opened 2024-06-28
Stop Auto Creation of GIS Subtasks When Submitting New Notifications to ALCS

Abradat:
- Stopped auto creation of GIS subtasks when submitting new SRW notifications to ALCS
- Removed previously automatically created GIS subtasks which were not completed yet


trslater (approved):
LGTM 🚀

PR #1780, opened 2024-07-04
Block Null and Zero Values for Released Decisions' ALR Inline Editor for Applications and NOIs

Abradat:
**This PR covers both ALCS-2065 and ALCS-2077 tickets requirements.**

Following changes have been made:

- Added non zero and required validators to detect zero and empty values on ALR inline edit
- Added form control to inline-number component to have validation and prevent user from submission
- Add conditions to decisions components to have validation only on released decisions in NOIs and Applications


trslater (approved):
LGTM 🚀

PR #1784, opened 2024-07-08
Fix Admin L/FNG Bugs

Abradat:
- Fixed error on creating and updating L/FNG admin accounts
- Added toast on successful create/update of L/FNG accounts


trslater (approved):
LGTM 🚀

PR #1786, opened 2024-07-10
Fix Public Search Status Date Showing the Date in the Future

Abradat:
Fixed the followings in the public search:

- Applications: 
    - Have a review discussion in the future that showed the future date although the status is "Received By ALCS"
    - Decision is going to release in the future. However, the status date shows the future date although the status is not "Decision Released"
- NOIs:
    - Decision is going to release in the future. However, the status date shows the future date although the status is not "Decision Released"


trslater (approved):
LGTM 🚀

PR #1787, opened 2024-07-10
Make applicant editable and change label

trslater:
- **Make applicant editible in header**
- **Add fixed height option to inline text field**
- **Restyle detail header with inline editor**
- **Update app sub table as well as app table**
- **Rename applicant to 'Name'**


Abradat (approved):
LGTM!

PR #1797, opened 2024-07-26
Fix Submission PDF Including "Other" Documents with Non-Applicant Source

Abradat:
- Fixed Applicant Submission PDF (Portal download and {DATE}_APP_Submission.pdf on ALCS) including 'other' documents with non-applicant source for applications and NOIs
    - Documents added after submission by ALC or during L/FNG review.


trslater (approved):
LGTM 🚀

PR #1798, opened 2024-07-29
Fix Error on Changing "Approx. Map Area" Field to Empty Value and Saving it

Abradat:
In parcel details step on portal submission, clearing "Approx. Map Area" field value and trying to save it after it has been saved before could cause 500 error on server side. This commit fixes the issue by sending null value instead of empty string to the server side.


trslater (approved):
LGTM 🚀

PR #1808, opened 2024-08-20
Add Mobile Design for Transferees Table for SRWs and Covenants

Abradat:


portal-frontend/src/app/features/notifications/edit-submission/transferees/transferees.component.ts
```diff
@@ -22,12 +23,15 @@ export class TransfereesComponent extends StepComponent implements OnInit, OnDes
   displayedColumns: string[] = ['type', 'fullName', 'organizationName', 'phone', 'email', 'actions'];
 
   private submissionUuid = '';
+  isMobile = false;
+  VISIBLE_COUNT = 5;
```
trslater, line 27: As a constant, I would move this outside of the component and define it with `const`. I'd also change the name to something like `VISIBLE_COUNT_INCREMENT`.

Abradat: Moved and renamed the constants to constants.ts file in shared module. (Same as alcs-frontend)

trslater: Nice. This is even better.

trslater (approved):
LGTM 🚀

PR #1819, opened 2024-08-26
Small fix for previous migration

trslater:
The timezone was incorrect on the previous migration
for the missed Village of Chase inquiries.


Abradat (approved):
LGTM

PR #1821, opened 2024-08-28
Add 'Go to card' button for Commissioners

trslater:
- **Add ability to highlight tile from query param**
- **Add go to card button**
- **Stop opening commissioner detail in new window**
- **Add option to get meetings by file number**
- **Only show go to card button when file has meetings in schedule**


Abradat (approved):
LGTM!

PR #1826, opened 2024-09-04
Fix Village of Chase migration

trslater:
To accomplish this fix, I couldn't just create a new migration, because the old one was failing, so I had to edit the old migration. The migration failed on test, so was only on dev. The files do not exist on dev, so no new records were created. Therefore, I could just delete the migration from the migrations table. I also moved the date fix into the main migration.

The fix is (hopefully) to wrap the queries in conditionals that check for the existence of the OATS schema.


Abradat (approved):
LGTM

PR #1827, opened 2024-09-04
Show Applications Without Discussion Date on Schedule Page

Abradat:
- For Applications, Reconsiderations, and Planning Reviews
- Rename "Scheduled Meetings"  column to "Incoming"
- Rename "None Scheduled" to "None"
- Add incoming-files controller and service
- Cards in the "Incoming column":
    - They do not have discussion date.
    - They are not in the first (Incoming / Prelim Review) or second column (App Prelim Done / To Be Assigned to LUP) on the Kanban board.
    - They are not in decision released or cancelled columns on Kanban board.
- Add sorting:
    - Highest order is priority
    - Second ordering by active days (planning reviews do not have active days and are considered 0)


trslater (approved):
LGTM 🚀

PR #1836, opened 2024-09-06
Fix "For Discussion" Column Showing the Furthest Date Instead of Closest

Abradat:
- If file has only meetings in the past, show the latest one. 
- If file has meetings in the past and the future, show the closest future one. 
- If the file has only meetings in the future, show the closest future one.

### DIFF
--- a/services/apps/alcs/src/alcs/application/application-decision-meeting/application-decision-meeting.service.ts
+++ b/services/apps/alcs/src/alcs/application/application-decision-meeting/application-decision-meeting.service.ts
@@ -131,7 +131,17 @@ export class ApplicationDecisionMeetingService {
   > {
     return await this.appDecisionMeetingRepository
       .createQueryBuilder('meeting')
-      .select('reconsideration.uuid, MAX(meeting.date) as next_meeting')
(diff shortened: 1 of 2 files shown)


services/apps/alcs/src/alcs/application/application-decision-meeting/application-decision-meeting.service.ts
```diff
@@ -131,7 +131,17 @@ export class ApplicationDecisionMeetingService {
   > {
     return await this.appDecisionMeetingRepository
       .createQueryBuilder('meeting')
-      .select('reconsideration.uuid, MAX(meeting.date) as next_meeting')
+      .select('reconsideration.uuid', 'uuid')
+      .addSelect(
+        `
+        CASE
+          WHEN MIN(CASE WHEN meeting.date >= (CURRENT_DATE) THEN meeting.date END) is NOT NULL
```
trslater, line 138: Have you made sure that `CURRENT_DATE` accounts for the time zone? I believe our dates are stored with time zones. This may work find, but I've run into issues with this before, so good to make sure.

trslater (approved):
LGTM 🚀

PR #1842, opened 2024-09-10
Show Application Decisions on Commissioner's Detailed View

Abradat:


alcs-frontend/src/styles/functions.scss
```diff
@@ -0,0 +1,10 @@
+@use 'sass:math';
+
+@function rem($pixels, $context: 16) {
```
trslater, line 3: The other decisions component uses the portal `rem` function. Is there any reason we can't use that instead?

trslater (approved):
LGTM 🚀
LGTM 🚀

PR #1843, opened 2024-09-11
Convert emails to use HTML templates instead of MJML

trslater:
- Change the way email templates are create and rendered
- Rewrite all templates in HTML


Abradat (approved):
LGTM!

PR #1845, opened 2024-09-11
Fix Multiple Files Showing in One Line in Public Search View

Abradat:
Fix the bug for "Cross Sections" and "Reclamation Plan" fields showing multiple files in one line in public view. The issue was for applications and NOIs for the following file types:
- POFO
- ROSO
- PFRS


trslater (approved):
LGTM 🚀

PR #1862, opened 2024-09-18
Create branching header for notifications

trslater:
Header template and masthead are now functions that accept an optional parameter, `isNotification`. If `true`, they display a different details section.


Abradat (approved):
LGTM

PR #1871, opened 2024-10-01
Add NARU siting explanation/figure drawer

trslater:
- **Add SVG siting figures**
- **Add NARU drawer with figures**
- **Add classes for figures, fig caps, and images**
- **Wrap figures to make size more consistent**


Abradat (approved):
LGTM!

PR #1878, opened 2024-10-02
Fix NARU Proposed Residence Issues

Abradat:
- Fix Proposed Residence table column in NARU submission PDF Template
- Fix text alignment in desktop view
- Add NARU proposed residence table to public view.


trslater (approved):
LGTM 🚀

Just trusting docx is correct—leaving it to QA.

PR #1883, opened 2024-10-03
NOI fixes

trslater:
- **Swap NOI prep fill/soil columns**
- **Make sure non-approved mods cannot be used in decisions**


Abradat (approved):
LGTM!

PR #1884, opened 2024-10-03
Resend the response when an email is changed

fbarreta:
When the email is changed, the resend method is called.
This method resent the notification to the new email, but also changes the status of the Notification:
await this.updateStatus(
        submission.uuid,
        NOTIFICATION_STATUS.ALC_RESPONSE_SENT,
      );

Which I believe will solve the sync issue displayed on the screen.


trslater (approved):
LGTM 🚀

PR #1888, opened 2024-10-04
Fixed the cache flags on SRW searches

fbarreta:
Fixed the cache flags on SRW searches.
So if there is a cached search, the SRW query is not repeated anymore with empty params.


Abradat (approved):
LGTM!

PR #1898, opened 2024-10-07
Fix Reconsiderations Not Showing in "Incoming" Column in commissioner…

Abradat:
Fix reconsiderations not showing in "Incoming" column in commissioner view if original card does not exist


trslater (approved):
LGTM 🚀

PR #1904, opened 2024-10-09
Fix ALCS NARU yes/no and no data text

trslater:
- **Properly map true/false to yes/no**
- **Add no data text for missing fields**


fbarreta (approved):
LGTM !

PR #1909, opened 2024-10-10
Fixes for improved NOI structure text

trslater:
- **Update question wording on portal view, public view, alcs app info**
- **Update structure types on ALCS**
- **Update question wording for PDF's**
- **Rename *all* occurences of structure type in ALCS**
- **Update tests with new structure type names**
- **Revert all renaming of structure types**
- **Add new label mapping to shared constants**
- **Map all existing NOI structure types to new labels**

Note: I changed the way new structure type labels were mapped to avoid having to migrate existing structures text.


fbarreta (approved):
LGTM !

PR #1922, opened 2024-10-21
Always show building plans w/ no data if empty in ALCS app info

trslater:
Previously, the building plans were hidden if not provided. Now, it shows"No Data".


Abradat (approved):
LGTM!

PR #1925, opened 2024-10-21
Add NARU Residence Mobile View in Public and Fix Missing Index Numbers

Abradat:
- 2316: Fix missing index numbers on residences on mobile and desktop view
- 2317: Add NARU residence mobile design to public view


trslater (approved):
LGTM 🚀

PR #1938, opened 2024-10-28
Fix selected ETL missed review deletions

trslater:
- When an app is returned from L/FNG, normally the review is deleted
- This was missed in ETL
- Any apps where a *second* L/FNG review was not yet started will be unable to start, since the old one already exists, creating a foreign key constraint validation


Abradat (approved):
LGTM!

PR #1939, opened 2024-10-28
Feature/alcs 2325 Tag and category implementation

fbarreta:
- Backend module refactor (Created tag module)
- Implementation of Tag and Tag Categories on ALCS Admin.


services/apps/alcs/src/alcs/tag/tag.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+import { TagDto } from './tag.dto';
+import { QueryFailedError } from 'typeorm';
+
+@Controller('tag')
+@ApiOAuth2(config.get<string[]>('KEYCLOAK.SCOPES'))
+@UseGuards(RolesGuard)
+export class TagController {
+  constructor(private service: TagService) {}
+
+  @Get('/:pageIndex/:itemsPerPage')
```
Abradat, line 29: In pagination and filtering, query parameters would be used instead of URL parameters. 

`/?pageIndex=1&itemsPerPage=10`

services/apps/alcs/src/alcs/tag/tag.controller.ts
```diff
@@ -0,0 +1,84 @@
...
+  @UserRoles(AUTH_ROLE.ADMIN)
+  async create(@Body() createDto: TagDto) {
+    try {
+      return await this.service.create(createDto);
+    } catch (e) {
+      if (e.constructor === QueryFailedError) {
+        const msg = (e as QueryFailedError).message;
+        throw new HttpException(msg, HttpStatus.CONFLICT);
+      } else {
+        throw e;
```
Abradat, line 50: What will happen if we just throw the error here?

alcs-frontend/src/styles/functions.scss
```diff
@@ -0,0 +1,10 @@
+@use 'sass:math';
+
+@function rem($pixels, $context: 16) {
```
Abradat, line 3: The reason that we did not put rem function in alcs-frontend was that alcs-frontend would be only used on desktop devices so rem would not be useful.

Abradat (changes requested):

alcs-frontend/src/app/features/admin/tag/tag-category/tag-category-dialog/tag-category-dialog.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    this.isLoading = false;
+    this.dialogRef.close(true);
+  }
+
+  onChange() {
+    this.showNameWarning = false;
+  }
+
+  private handleError(e: any) {
```
trslater, line 65: Really not a fan of `any`. It can lead to so many issues. Could we try to resolve this type up front? It seems like you're already expecting it to be a `HttpErrorResponse`, could we use this as the param type?

trslater: There are a few other cases of the exact same thing. If you figure out a good solution, please apply to other cases.

services/.prettierrc
```diff
@@ -1,4 +1,9 @@
...
   "singleQuote": true,
-  "trailingComma": "all"
-}
\ No newline at end of file
+  "useTabs": false,
+  "tabWidth": 2,
+  "semi": true,
+  "bracketSpacing": true,
+  "endOfLine": "auto"
+}
```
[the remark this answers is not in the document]
fbarreta: This endOfLine was just freaking me out on my VSCode and fixed my CR/CRLF warnings hope it won't cause any harm, and added some of the same bracketSpacing as our frontends.

Abradat (approved):
LGTM!

PR #1942, opened 2024-10-30
ALCS-644 Tags/Categories seed and nullable categories

fbarreta:
- Made category on tag non nullable
- Soft delete tags
- Category seed
- Tag seed


Abradat (approved):
LGTM!

PR #1943, opened 2024-10-30
Add structures to soil apps

trslater:
- **Move structure type options to module constant for use by other modules**
- **Rename NOI structure label map to make general for all soil**
- **Add new structures question to DB**
- **Git add new structure question to all services and DTO's**
- **Add new structure question to UI**
- **Show/hide old questions based on answer to new structure question**
- **Add all new questions to DB**
- **Add new conditional structure questions to all services and DTO's**
- **Add missed DTO field**
- **Add structures to all DTO's and services**
- **Expose structure form type for use by other modules**
- **Add structures table and conditional questions to UI**
- **Remove unused imports**
- **Add is new structure confirmation dialog**
- **Remove conditional app type text**
- **Fix NOI structure dialog return type issue**
- **Add wanting when changing structure type on mobile**
- **Implement mobile UI for soil apps**
- **Update POFO/ROSO with changes from PFRS**
- **Add building plans**
- **Fix clearing file uploads**
- **Update POFO/ROSO styles**
- **Fix structure add button layout**
- **Fix POFO/ROSO wording**
- **Add updates to portal view**
- **Fix missing saved submission fields**
- **Fix submission validation**
- **Fix structures type in DTO**
- **Add app soil updates to app info**
- **Update public view with soil changes**
- **Fix app info building plans**
- **Fix building plans being deleting for wrong action**
- **Fix PFRS validation**
- **Fix POFO reduce impact visibility**
- **Fix ROSO sctructure table layout**
- **Display POFO reduce impact field**
- **Pipe new structures data to PDF templates**
- **Add structure changes to PDF's**
- **Fix m2 -> m^2**


services/apps/alcs/src/portal/pdf-generation/generate-submission-document.service.ts
```diff
@@ -495,6 +495,25 @@ export class GenerateSubmissionDocumentService {
...
+    const hasFarmStructure = submission.soilProposedStructures.some(
+      (structure) => structure.type === 'Farm Structure',
+    );
+
+    const hasResidentialStructure = submission.soilProposedStructures.some(
+      (structure) =>
+        structure.type === 'Residential - Principal Residence' ||
+        structure.type === 'Residential - Additional Residence' ||
+        structure.type === 'Residential - Accessory Structure',
+    );
```
fbarreta, line 507: Wouldn't be better to compare this to an Enum like it's used here:
portal-frontend\src\app\features\notice-of-intents\edit-submission\additional-information\additional-information.component.ts
export enum STRUCTURE_TYPES {
  FARM_STRUCTURE = 'Farm Structure',
  PRINCIPAL_RESIDENCE = 'Residential - Principal Residence',
  ADDITIONAL_RESIDENCE = 'Residential - Additional Residence',
  ACCESSORY_STRUCTURE = 'Residential - Accessory Structure',
  OTHER_STRUCTURE = 'Other Structure',
}

fbarreta:
Besides the comment I left (which I believe is optional), overall looks great.
I couldn't find anything that blocks the merge:
-  Everything implements the structure like on NOI
- All the bugfixes are carried and implemented
Just let me know what you think about the comment, and I'll just drop the approval.

(Q&Aing this, is the bigger challenge .... hahah)

services/apps/alcs/src/portal/pdf-generation/generate-submission-document.service.ts
trslater: Yeah, good point! I was thinking there wasn't an enum like this in the services app, but there is. Probably should be applied to NOI's as well.

trslater: I agree it would be better, but I did it this way, because a) how NOI's does it and b) there is no enum for services or an obvious place to put it. I guess I could create an enum for services. I'll just have to figure out a place that makes sense, and should probably apply to NOI's as well.

fbarreta (approved):
G2G !!!

PR #1944, opened 2024-10-30
ALCS-2327 Allow the changes on desktop version to reflect on mobile

fbarreta:
Every time the desktop version changes the structure type, it persists on the FormProposedStructure array, and this array is used to render the mobile component.
It doesn't exist when the area is changed, that's why it was not carried to the mobile view.
Added two methods to replicate the behavior to the Area fields, and carry the changes to the mobile view.

### DIFF
--- a/portal-frontend/src/app/features/notice-of-intents/edit-submission/additional-information/additional-information.component.ts
+++ b/portal-frontend/src/app/features/notice-of-intents/edit-submission/additional-information/additional-information.component.ts
@@ -353,6 +353,12 @@ export class AdditionalInformationComponent extends FilesStepComponent implement
     this.form.markAsDirty();
   }
 
+  private setStructureAreaInput(structure: FormProposedStructure, value: string) {
+    structure.area = value;
+    this.prepareStructureSpecificTextInputs();
+    this.form.markAsDirty();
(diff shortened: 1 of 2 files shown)


portal-frontend/src/app/features/notice-of-intents/edit-submission/additional-information/additional-information.component.ts
```diff
@@ -383,6 +389,16 @@ export class AdditionalInformationComponent extends FilesStepComponent implement
...
   }
 
+  onChangeArea(id: string, event: Event) {
+    const input = event.target as HTMLInputElement;
+    const structure = this.proposedStructures.find((structure) => structure.id === id);
+    if (!structure) {
+      console.error('Failed to find structure');
+      return;
+    }
+    return this.setStructureAreaInput(structure, input.value);
```
trslater, line 399: `setStructureAreaInput` doesn't return a value, so we don't really need to return here.

fbarreta: makes sense !

trslater (approved):
LGTM 🚀

PR #1949, opened 2024-11-01
Add Tags Functionality to Applications and NOIs

Abradat:
- Tags for applications and NOIs
    - Adding and Removing tags from file page
- Admin Tags
    - Removing a tag and checking if an application or NOI is associated with it
    - Removing a category and checking if a tag is associated with it


fbarreta (approved):
G2G !

PR #1951, opened 2024-11-04
QA Fixes: 472-2213

Abradat:
- Fix tags header border behaviour: The border should disappear once user has selected a tag and clicks on another component
- Make tag name bold on confirmation dialog when deleting a tag from application/NOI
- Hide inactive tags from user for selecting in applications and NOIs

### DIFF
--- a/alcs-frontend/src/app/shared/confirmation-dialog/confirmation-dialog.component.html
+++ b/alcs-frontend/src/app/shared/confirmation-dialog/confirmation-dialog.component.html
@@ -2,7 +2,7 @@
   <span class="subheading2">{{ data.title }}</span>
 </div>
 <div mat-dialog-content>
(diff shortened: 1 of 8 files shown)


alcs-frontend/src/app/shared/confirmation-dialog/confirmation-dialog.component.html
```diff
@@ -2,7 +2,7 @@
   <span class="subheading2">{{ data.title }}</span>
 </div>
 <div mat-dialog-content>
-  <p>{{ data.body }}</p>
+  <p [innerHTML]="data.body">{{ data.body }}</p>
```
trslater, line 5: Why do we need to set this twice?

trslater (approved):
LGTM 🚀

PR #1953, opened 2024-11-05
Feature/alcs 1874 tags on advanced search

fbarreta:
Advanced search for tags and categories.


services/apps/alcs/src/alcs/search/notice-of-intent/notice-of-intent-advanced-search.service.ts
```diff
@@ -206,17 +183,14 @@ export class NoticeOfIntentAdvancedSearchService {
 
     const finalResult = intersectSets(allIds);
 
+    console.log(finalResult);
```
trslater, line 186: And this?

services/apps/alcs/src/alcs/search/notice-of-intent/notice-of-intent-advanced-search.service.ts
```diff
@@ -197,6 +173,7 @@ export class NoticeOfIntentAdvancedSearchService {
 
     const allIds: Set<string>[] = [];
     for (const result of queryResults) {
+      console.log('count');
```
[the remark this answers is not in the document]
fbarreta: def not .... hahaha

alcs-frontend/.eslintrc.json
```diff
@@ -1,6 +1,7 @@
 {
   "root": true,
   "ignorePatterns": ["projects/**/*"],
+  "no-octal-escape": 0,
```
Abradat, line 4: I would prefer we move all the eslintrc changes to another since it is out of scope of this change

trslater (approved):
LGTM 🚀

alcs-frontend/src/app/features/search/search.component.html
```diff
@@ -79,6 +79,65 @@ <h4>File Details</h4>
...
+                #tagInput
+                [matChipInputFor]="chipGrid"
+                [formControl]="tagControl"
+                [matAutocomplete]="auto"
+                [matChipInputSeparatorKeyCodes]="separatorKeysCodes"
+                (focus)="autoCompleteTrigger.openPanel()"
+                (click)="clearSearch()"
+                (blur)="checkDirty()"
+              />
+              @if (tags.length > 0) {
```
Abradat, line 105: You're using *ngFor in the bottom and @if here which is the new template syntax. Was there any specific reason for it?

PR #1954, opened 2024-11-05
Remove '[object object]' from bottom of empty attachments table

trslater:


fbarreta (approved):
Although I can't see this changes on github, I believe this is a simple template update.

PR #1956, opened 2024-11-06
Add Tags to NOIs Once Submitted Instead of Subtypes

Abradat:
- Add tags to NOIs once submitted to ALC based on structure types and area-wide filling question
- Deprecated populateNoiSubtype function


fbarreta (approved):
LGTM

PR #1959, opened 2024-11-06
ALCS-1874 Refresh tags params correctly

fbarreta:
The search params were not being refresh correctly on subsequent searches.
Used array mapping to update.

Also found an unrelated bug that was firing some console errors, and fixed it.
image


Abradat (approved):
LGTM

PR #1960, opened 2024-11-06
1156 QA fixes 2

trslater:
- **Undo previous fix adding warning to accessory structures**
- **Fix type mobile switching not updating conditional fields properly**
- **Fix various field state issues**
- **Make structure type counts more reliable**
- **Fix issue with mobile structure type not reflecting on desktop**


fbarreta (approved):
LGTM

PR #1962, opened 2024-11-06
QA Fix: Fix NOI Tags Not Being Saved Correctly Upon Submission

Abradat:
The map function would execute addTag function in an asynchronously that caused tags to be saved incorrectly.


trslater (approved):
LGTM 🚀

PR #1963, opened 2024-11-06
ALCS-2350 Remove borders when there is a residential warning

fbarreta:
Remove <td> borders when there is a residential warning.


Abradat (approved):
LGTM

PR #1973, opened 2024-11-13
Fix NOI structure questions disappearing

trslater:
- **Set structure type form control value when updating type**
- **Patch form after structure questions visibility has been set**
- **Explicitly set return type of `addControl`**


Abradat (approved):
LGTM

PR #1975, opened 2024-11-13
Fix S&F app public search visibility issues

trslater:
- **Show 'what steps...' question on 'No' instead of 'Yes'**
- **Move "what alternative..." question inside structures check**


Abradat (approved):
LGTM!

PR #1981, opened 2024-11-14
Revert "ALCS-2379 Naru text changes"

fbarreta:
Reverts bcgov/alcs#1980


fbarreta:
revert not not needed

PR #1984, opened 2024-11-15
ALCS-2384 Prevent results when tags filters are on

fbarreta:
Hotfix for prevent Notifications, Planning and Inquiries returning results when tags filters are on.


Abradat (approved):
LGTM

PR #1990, opened 2024-11-18
ALCS-2389 Fix table change method call

fbarreta:
The table type was comparing with wrong strings.
Fixed so all the pagination methods are correctly called.

### DIFF
--- a/portal-frontend/src/app/features/home/inbox/inbox-table/inbox-table.component.ts
+++ b/portal-frontend/src/app/features/home/inbox/inbox-table/inbox-table.component.ts
@@ -59,7 +59,6 @@ export class InboxTableComponent {
   }
 
   private getTableType(): string {
-    console.log(this.type);
     switch (this.type) {
       case 'Notice of Intent':
         return 'NOI';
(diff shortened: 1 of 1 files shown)


portal-frontend/src/app/features/home/inbox/inbox-table/inbox-table.component.ts
```diff
@@ -57,4 +57,16 @@ export class InboxTableComponent {
       await this.router.navigateByUrl(link);
     }
   }
+
+  private getTableType(): string {
+    console.log(this.type);
```
Abradat: remove log :D

portal-frontend/src/app/features/home/inbox/inbox-table/inbox-table.component.ts
```diff
@@ -57,4 +57,16 @@ export class InboxTableComponent {
...
+    switch (this.type) {
+      case 'Notice of Intent':
+        return 'NOI';
+      case 'Notification':
+        return 'NOTI';
```
Abradat, line 66: We usually use SRW for notifications. It would be great if we have the same value for integrity across application

Abradat (changes requested):

portal-frontend/src/app/features/home/inbox/inbox-table/inbox-table.component.ts
fbarreta: Maybe it's not the best time to change it. Cause the URL's are dependent on this type, and will start breaking in many other places.

Abradat (approved):
LGTM

PR #1997, opened 2024-11-21
ALCS-2399 Order by label on retrieving codes

fbarreta:
The drop down on add condition should be ordered by label. Previously it was ordered by edited date.

image

image


Abradat (approved):
LGTM!

PR #1999, opened 2024-11-26
Add status and delete to condition types admin UI

trslater:
- **Add is_active to DB for apps/NOI's**
- **Add is_active to condition type DTO's**
- **Add is_active to condition type services**
- **Add is_active to frontend DTO's**
- **Add status column to conditions admin UI**
- **Add status field to condition type dialog**
- **Add delete button/functionality**
- **Use proper NOI DTO instead of apps DTO**
- **Allow delete error message to be set by backend**
- **Throw error if trying to delete condition type associated with undeleted decision**


Abradat (approved):
LGTM

PR #2001, opened 2024-11-28
Feature/alcs 2092 add due date to condition types

fbarreta:
Update conditions types according to this description:

Use a migration to add optional single date field with label = Due Date to the Edit Decision UI of all conditions EXCEPT 'Use End Date' condition type or 'Status Report' condition type


trslater (approved):
LGTM 🚀

PR #2002, opened 2024-11-28
QA Fix: Fix Application and NOIs Decision Condition Type Bugs

Abradat:
- Disable the code field on edit mode
- Remove zero value on conditions without admin fee


trslater (approved):
LGTM 🚀

PR #2005, opened 2024-12-02
Feature/alcs 2099 components script and date removals

fbarreta:
Date fields removal (only UI) from Applications and NOI Decision components.

Script to migrate all component dates to new Use End Date Conditions.


trslater (approved):
LGTM 🚀

PR #2006, opened 2024-12-05
2322 QA Fixes

trslater:
- **Fix undefined app component name**
- **Apply same change to NOI's**
- **Add success toast when deleting condition types**


fbarreta (approved):
LGTM!

PR #2007, opened 2024-12-11
Block the User from Entering the Code that is Already Taken

Abradat:
Following sections on the admin:
- Condition Types (Applications and NOIs)
- CEO Criterion
- Decision Makers
- Boards
- Columns (Card Status)


fbarreta (approved):
LGTM

PR #2008, opened 2024-12-11
Feature/alcs 2098 migration to create end date

fbarreta:
Inserts or updates the Use End of Date on application and noi type.


trslater (approved):
LGTM 🚀

PR #2009, opened 2024-12-11
Feature/alcs 2092 new migration to add fee

fbarreta:
Add migration fee to types that are not 'SRPT' and 'UEND'


trslater (approved):
LGTM 🚀

PR #2019, opened 2024-12-16
ALCS-2332 Added 5 decimal places validation and user hint

fbarreta:
The filed is already saving the right amount of decimals on DB.

- Fixed the NOI input field
- Added user hints on NOI and Application


trslater (approved):
LGTM 🚀

PR #2020, opened 2024-12-16
2312 QA 3

trslater:
- **Allow saving of dates to unsaved conditions**
- **Fix browser error on cancel date dialog**
- **Sort condition dates on setting**
- **Apply fix to 'can't save dates to new conditions' done previously to apps to NOI's**


Abradat (approved):
LGTM!

PR #2026, opened 2024-12-17
ALCS-1850 Change expiry comparision and compare only date portions

fbarreta:
Change expiry comparison from <= to just <
Compare only date portions truncating the time.


trslater (approved):
LGTM 🚀

trslater (approved):
LGTM 🚀

PR #2029, opened 2024-12-18
ALCS-2431 Validate Admin fee and security amount

fbarreta:
Added validation to existing admin fees and security amounts.


Abradat (approved):
LGTM!

PR #2032, opened 2024-12-18
ALCS-2431 Prevents validation flash

fbarreta:
- Prevents validation flash using isLoading state while fetching
- Sets admin fee to null on submit if unchecked


Abradat (approved):
LGTM!

PR #2033, opened 2024-12-18
ALCS-2446 Set single date as default if date is not checked

fbarreta:
Set single date as default when the date is not checked and set.


trslater (approved):
LGTM 🚀

PR #2036, opened 2024-12-19
ALCS-2416 Added  preventDefault on dropdowns

fbarreta:
I couldn't find a particular reason for the dropdown to open.
With some research I read that angular material have some issues with overlays, and these two being the same component could make use of the same overlay. (and it triggers the first dropdown no matter where it's on the screen ... I tested it ... :( )
But the event.preventDefault stops the overlay actions, and triggers only the desired functionality.


trslater (approved):
LGTM 🚀

PR #2037, opened 2024-12-19
ALCS-2422 Fix the controllers with the correct role

fbarreta:
Fixed the controller setting the status searches with the right role.


trslater (approved):
LGTM 🚀

PR #2038, opened 2024-12-19
Move all security amounts to bond conditions

trslater:
- **Add missing bond conditions**
- **Clear all non-bond security amounts**
- **Disable security amount for non-bond condition types**
- **Enable required security amount for bond condition type**


fbarreta (approved):
LGTM

PR #2045, opened 2024-12-31
Add Tags and File Status to Commissioner View

Abradat:
- Add tags to the application header on commissioner's view
- Fix tag and tag category fetch error on commissioner advanced search
- Add application status pill to commissioner's view


trslater (approved):
LGTM 🚀

PR #2046, opened 2025-01-06
ALCS-2453 Handle unique descriptions on condition types

fbarreta:
We have a constraint that all the CODE based entities should have an unique description.

What was done:

Check the existence of a description and display a clear message to the user about the uniqueness of the description.

image


trslater (approved):
LGTM 🚀

PR #2048, opened 2025-01-06
ALCS-2451 Allow to remove check box if the value is zero

fbarreta:
Allow the checkboxes to be disabled if the stored value is zero.

There's also a minor fix on the edit dialog, to prevent the validation from unique description on the actual value.


Abradat (approved):
LGTM!

PR #2049, opened 2025-01-07
Fix Portal Status and File Type Dropdown Bugs

Abradat:
fix the advanced search bug that clicking on "File Type" text on dropdown would open the "Current Portal Status" dropdown


trslater (approved):
LGTM 🚀

PR #2051, opened 2025-01-13
Feature/alcs 2059 decision view summary

fbarreta:
Created summary for Applications and NOI, including conditions statuses, if associated with components.

PS: Merged Tristan's hotfix that fixed dates, necessary to this implementation.

image
image


Abradat (approved):
LGTM!

PR #2052, opened 2025-01-14
ALCS-2426 Navigate to conditions when decision released

fbarreta:
Navigate to conditions when decision released for Applications and NOIs


trslater (approved):
LGTM 🚀

PR #2053, opened 2025-01-14
ALCS-2467 Check decisions

fbarreta:
The validation was checking only for conditions, but there were some orphan rows that were invalidating the fees removal.

I added the decision relation on conditions methods, updated the DTOs, and filtered only the decision related conditions.

Maybe we should capture a technical debt to handle the conditions while deleting a decision.


Abradat (approved):
LGTM

PR #2056, opened 2025-01-15
Allow all but commissioner to archive cards

trslater:
- Unhide archive button for all but commissioner
- Add all but commissioner role to archive endpoint


fbarreta (approved):
LGTM

PR #2060, opened 2025-01-17
ALCS-2474 Fix single date updates

fbarreta:
While saving the whole condition, with already having updated the date in a single date condition, the date row was being updated only with the date portion but deleting the completedDate and the comment, since it was being send to the backend ONLY the date attribute, deleting the attributes added earlier, and changing the status in a wrong way.
The change checks if there's more data on that date and emits everything.


trslater (approved):
LGTM 🚀

PR #2061, opened 2025-01-17
Application Conditions Card

Abradat:
- New board and card status through migration
- New Application Decision Condition Card entity, service, controller
- New dialog for creating condition card in application decision conditions page
- New board card dialog for application conditions
- Modify board and relevant services for new requirements


fbarreta (approved):
LGTM

PR #2066, opened 2025-01-20
Application Condition Card UI/UX features

Abradat:
- Add "Go to Condition" to the card dropdown in the header
- Add 'GO TO BOARD' to toast service after creating a condition card
- Add status pills to the condition card in the "Application Condition Board"
- Fix Tooltip not showing on disabled conditions
- Add decision order to the condition card dialog

### DIFF
--- a/alcs-frontend/src/app/features/board/board.component.ts
+++ b/alcs-frontend/src/app/features/board/board.component.ts
@@ -31,7 +31,7 @@ import {
   RECON_TYPE_LABEL,
   RETROACTIVE_TYPE_LABEL,
 } from '../../shared/application-type-pill/application-type-pill.constants';
-import { CardData, CardSelectedEvent, CardType } from '../../shared/card/card.component';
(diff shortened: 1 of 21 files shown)


alcs-frontend/src/app/features/board/board.component.ts
```diff
@@ -507,7 +507,19 @@ export class BoardComponent implements OnInit, OnDestroy {
 
   private mapApplicationDecisionConditionToCard(
     applicationDecisionCondition: ApplicationDecisionConditionCardBoardDto,
-  ): CardData {
+  ): ConditionCardData {
+    let isExpired = false;
+    let isPastDue = false;
+
+    for (const condition of applicationDecisionCondition.conditions) {
+      if (condition.status === 'EXPIRED') {
```
trslater, line 515: Super nitpicking, but for these situations, I always think it's a little cleaner to just do `isExpired = condition.status === 'EXPIRED'`. I think it's also more robust to use condition status enum directly: `isExpired = condition.status === CONDITION_STATUS.EXPIRED`

PR #2072, opened 2025-01-21
QA Fixes #1: 1871

Abradat:
- Fix application condition board cards drag and drop bug not working
- Show all conditions only when editing the card
- Show the error color on the table border and checkbox color when no condition is selected
- Disabled checkbox for conditions already in use by other cards
- Add 'No Data' text to conditions without End/Due date value


fbarreta (approved):
LGTM

PR #2073, opened 2025-01-22
QA Fixes #2: 1871

Abradat:
Fix the bug that showed wrong "Due Date" on application decision condition card when a condition with multiple date type has no date value


trslater (approved):
LGTM 🚀

PR #2075, opened 2025-01-22
ALCS-2478 Add date to the single date conditions

fbarreta:
This script looks for all the conditions that are from Single Date type, that have no dates associated, and add a date row, allowing the user to manage the date, they can update and also add the completion.

I decided not the make any UI/Backend changes to not introduce any workarounds, and keep the data integrity they way we projected to be.


trslater (approved):
LGTM 🚀

PR #2077, opened 2025-01-22
ALCS-2478 Add dates to NOI conditions

fbarreta:
Same script used for applications, but on NOIs now.

Applications were already tested and validated by Nowshin.


Abradat (approved):
LGTM

PR #2083, opened 2025-01-23
Align revert button to right like edit button

trslater:
Edit draft and revert to draft buttons are separate, so grid settings were only being applied to one. This just adds the same grid column setting to the revert button, and aligns it right, so the button is flush.


Abradat (approved):
LGTM

PR #2084, opened 2025-01-23
ALCS-2478 Create new dates when duplicating conditions

fbarreta:
A hard to find issue ...
But, when we were duplicating a condition the dates were kept with the exact same properties, including the UUID.
So when saving the new condition the dates were updated to the new condition, leaving the old condition without dates.

The fix itself was pretty easy:
```typescript
condition.dates = condition.dates?.map((d) => {
        return new ApplicationDecisionConditionDate({
          ...d,
          uuid: undefined,
        });
      });
```
For the copied dates, I just set the UUID to undefined, and this would make typeorm create a new date row, instead of update the old one.
The same fix was also applied to NOIs.


trslater (approved):
LGTM 🚀

PR #2086, opened 2025-01-23
Archive and Unarchive Application Decision Condition Cards

Abradat:
- Disassociate conditions when the relevant condition card is archived
- Update conditions table design when the card is unarchived


services/apps/alcs/src/alcs/card/card.service.ts
```diff
@@ -225,6 +217,11 @@ export class CardService {
     const subtaskUuids = card.subtasks.map((subtask) => subtask.uuid);
     await this.subtaskService.recoverMany(subtaskUuids);
 
+    if (card.typeCode === CARD_TYPE.APP_CON) {
+      console.log(card.uuid);
```
trslater, line 221: Is this logging supposed to be here?

trslater (approved):
LGTM 🚀

PR #2089, opened 2025-01-25
Move Application Conditions to Other Boards

Abradat:
- Add application condition card type support to boards (CEO, Executive Committee, and all panels)
- Add condition, modification, and reconsideration pills when moving the condition card to other non-condition boards (and vice versa)
- Remove expired and past due pills when moving the condition card to other non-condition boards (and vice versa)


fbarreta (approved):
LGTM

PR #2092, opened 2025-01-28
2448: Fix public doc sorting

trslater:
3 things were needed:
- The table source data needed to be updated via the `data` property rather than replaced
- Sorting needed to be initialized after view init (at this point, all columns were sortable except for type)
- A custom sorting accessor was needed to make the type column work (since its value was a sub property)


Abradat (approved):
LGTM!

PR #2094, opened 2025-01-29
QA Fixes #2: 2406

Abradat:
- Add application decision condition cards to the commissioner schedule's meetings
- Remove the application conditions board from commissioner's view

### DIFF
--- a/services/apps/alcs/src/providers/typeorm/migrations/1738089637628-turn_off_application_condition_board_show_on_schedule.ts
+++ b/services/apps/alcs/src/providers/typeorm/migrations/1738089637628-turn_off_application_condition_board_show_on_schedule.ts
@@ -1,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class TurnOffApplicationConditionBoardShowOnSchedule1738089637628 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+          UPDATE alcs.board
+          SET show_on_schedule = false
+          WHERE code = 'appcon'
+        `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+          UPDATE alcs.board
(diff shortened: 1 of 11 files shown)


services/apps/alcs/src/providers/typeorm/migrations/1738089637628-turn_off_application_condition_board_show_on_schedule.ts
trslater, line 1: This seems potentially dangerous. I know we haven't released to test/prod yet, but I'm just worried that such a wide spread wiping out of values seems risky. Also, in lieu of our chat about migrations, should we find another way to update this data?

services/apps/alcs/src/providers/typeorm/migrations/1738196945240-remove_application_card_type_from_condition_board.ts
trslater, line 1: Same issue as previous here.

services/apps/alcs/src/providers/typeorm/migrations/1738089637628-turn_off_application_condition_board_show_on_schedule.ts
Abradat: This would change the boolean value of the 'appcon' board to prevent showing the board to the commissioner. would it be a big change/wipeout? 

We should manually do it on deployments if we do not want to do it.

services/apps/alcs/src/providers/typeorm/migrations/1738196945240-remove_application_card_type_from_condition_board.ts
Abradat: This would affect a single row from the board_allowed_card_types_card_type table, removing the row matching application conditions and APP card type. would it be a big change/wipeout?

I think we would have to manually do it on deployments if we do not want to do it through migration. Plus, if someone installs/restores the DB locally, they should be reminded of this process manually to prevent the out-of-sync behavior of the application.

trslater (approved):
LGTM 🚀

PR #2095, opened 2025-01-30
ALCS-2081 Added condition cards on homepage

fbarreta:
Condition Cards on homepage

### DIFF
--- a/alcs-frontend/src/app/features/home/assigned/assigned.component.ts
+++ b/alcs-frontend/src/app/features/home/assigned/assigned.component.ts
@@ -106,8 +106,6 @@ export class AssignedComponent implements OnInit {
         .sort((a, b) => a.date! - b.date!),
     ];
 
-    console.log(this.applications);
-
     this.planningReferrals = [
       ...planningReferrals
         .filter((r) => r.card!.highPriority)
--- a/alcs-frontend/src/app/features/home/assigned/assigned.component.ts
+++ b/alcs-frontend/src/app/features/home/assigned/assigned.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { ApplicationModificationDto } from '../../../services/application/application-modification/application-modification.dto';
 import { ApplicationReconsiderationDto } from '../../../services/application/application-reconsideration/application-reconsideration.dto';
-import { ApplicationDecisionConditionDto, ApplicationDto } from '../../../services/application/application.dto';
+import { ApplicationDto } from '../../../services/application/application.dto';
 import { ApplicationService } from '../../../services/application/application.service';
 import { HomeService } from '../../../services/home/home.service';
 import { InquiryDto } from '../../../services/inquiry/inquiry.dto';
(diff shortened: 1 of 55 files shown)


services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition-card/application-decision-condition-card.dto.ts
```diff
@@ -27,6 +27,26 @@ export class ApplicationDecisionConditionCardDto {
   applicationFileNumber?: string | null;
 }
 
+export class ApplicationDecisionConditionHomeCardDto {
```
Abradat, line 30: Why not using the already available "ApplicationDecisionConditionCardDto"? 

The difference is decisionUuid which is optional so if not assigned it won't appear in the DTO

services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.dto.ts
```diff
@@ -93,10 +91,13 @@ export class ApplicationDecisionConditionDto {
...
 
-  @AutoMap(() => ApplicationDecisionConditionCardUuidDto)
-  conditionCard: ApplicationDecisionConditionCardUuidDto | null;
+  @AutoMap(() => ApplicationDecisionConditionHomeCardDto)
+  conditionCard: ApplicationDecisionConditionHomeCardDto | null;
 
   status?: string | null;
+
+  @AutoMap()
+  decision?: ApplicationDecisionDto;
```
Abradat, line 100: the decision DTO already has the conditions info, if we add the decision here, wouldn't it cause a circular dependencies and a lot of unused data?

services/apps/alcs/src/alcs/application-decision/application-decision-v2/application-decision/application-decision.dto.ts
```diff
@@ -272,6 +269,9 @@ export class ApplicationDecisionDto {
 
   @AutoMap(() => Number)
   flagEditedAt: number | null;
+
+  @AutoMap()
+  application: ApplicationDto;
```
Abradat, line 274: Application DTO has various fields and data. Do we need to include all of the data in the decision DTO? Adding it seems like a specific case.

alcs-frontend/src/app/features/home/assigned/assigned.component.ts
```diff
@@ -171,6 +181,18 @@ export class AssignedComponent implements OnInit {
     };
   }
 
+  private mapApplicationCondition(a: ApplicationDecisionConditionDto): AssignedToMeFile {
```
Abradat, line 184: We're using a few fields of the heavy DTO being sent here. Would it be better if we have a customized DTO containing only the required fields?

Abradat (changes requested):
Overall, I think it would be better that we create and maintain DTOs for specific cases.  By keeping them general, we would grow them bigger over time according to our use cases, and it would introduce issues such as performance, dependency, etc.

My suggestion would be to create a customized DTO for this specific use case so we don't touch other DTOs and don't add more fields to them like Application Decision DTO.

"ApplicationDecisionConditionCardBoardDto" can be an example that is customized for a specific use case. Most of the fields such as "decisionOrder", "decisionUuid", "decisionIsFlagged", "isReconsideration", and "isModification" are required for this specific use case, so they're fetched from different services and mapped to this DTO.

Abradat (approved):
LGTM!

PR #2097, opened 2025-01-31
NOI Condition Cards - Backend

Abradat:
NOI Condition Cards backend functionalities:

- NOI Condition Card entity, DTO, service, controller
- Add NOI condition card to relevant services, including card and board
- Migrations for creating new NOI conditions board, card type, and new NOI condition card entity


trslater (approved):
LGTM 🚀

PR #2098, opened 2025-01-31
ALCS-2468 Delete dependencies

fbarreta:
After deleting a decision, also clean the relations with:

- application_decision_component
- application_decision_condition
- application_decision_condition_date
- application_decision_document

and

- notice_of_intent_decision_component
- notice_of_intent_decision_condition
- notice_of_intent_decision_condition_date
- notice_of_intent_decision_document

Change of condition cards.

- Stop soft removing the cards
- Removing the relation from the cards

### DIFF
--- a/services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.service.ts
+++ b/services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.service.ts
@@ -152,7 +152,7 @@ export class ApplicationDecisionConditionService {
 
     await this.repository.manager.transaction(async (transactionalEntityManager) => {
       await transactionalEntityManager.remove(lots);
-      await transactionalEntityManager.softRemove(conditions);
+      await transactionalEntityManager.remove(conditions);
     });
   }
 
--- a/services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.service.ts
+++ b/services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.service.ts
@@ -24,8 +24,6 @@ export class ApplicationDecisionConditionService {
     private conditionComponentPlanNumbersRepository: Repository<ApplicationDecisionConditionComponentPlanNumber>,
     @InjectRepository(ApplicationDecisionConditionToComponentLot)
     private conditionComponentLotRepository: Repository<ApplicationDecisionConditionToComponentLot>,
-    @InjectRepository(ApplicationDecisionConditionDate)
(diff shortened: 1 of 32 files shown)


services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.service.ts
```diff
@@ -150,9 +152,13 @@ export class ApplicationDecisionConditionService {
       },
     });
 
+    conditions.forEach(async (c) => {
+      await this.dateRepository.softRemove(c.dates);
```
trslater, line 156: I think we're hard deleting dates

services/apps/alcs/src/alcs/notice-of-intent-decision/notice-of-intent-decision-condition/notice-of-intent-decision-condition.service.ts
```diff
@@ -137,7 +139,10 @@ export class NoticeOfIntentDecisionConditionService {
   }
 
   async remove(conditions: NoticeOfIntentDecisionCondition[]) {
-    await this.repository.remove(conditions);
+    conditions.forEach(async (c) => {
+      await this.dateRepository.softRemove(c.dates);
```
trslater, line 143: Again, think this should be hard delete.

services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.service.ts
```diff
@@ -150,9 +152,13 @@ export class ApplicationDecisionConditionService {
...
+    });
+
     await this.repository.manager.transaction(async (transactionalEntityManager) => {
-      await transactionalEntityManager.remove(lots);
```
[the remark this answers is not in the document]
fbarreta: I tried to make it consistent with the other transactions we were doing. But, I can revert it to how it was before. (including the dates above ...)

services/apps/alcs/src/alcs/application-decision/application-decision-condition/application-decision-condition.service.ts
```diff
@@ -152,7 +152,7 @@ export class ApplicationDecisionConditionService {
 
     await this.repository.manager.transaction(async (transactionalEntityManager) => {
       await transactionalEntityManager.remove(lots);
-      await transactionalEntityManager.remove(conditions);
```
trslater, line 155: Same here, I'm not sure we should be changing stuff that is already working this way. Soft delete can have some annoying issues.

trslater (approved):
LGTM 🚀

PR #2099, opened 2025-01-31
Throw error if correct documents failed to generate and attach

trslater:
- Applies to apps, NOI's SRW's
- Also update message with more detail


Abradat (approved):
LGTM!

PR #2101, opened 2025-02-04
NOI Condition Card Frontend Component

Abradat:
- Add NOI Condition card frontend component, services, and DTOs.
- Add NOI condition card to the board
- Update tests


trslater (approved):
LGTM 🚀

PR #2102, opened 2025-02-04
2509: Redirect on login failures

trslater:
This does not necessarily resolve 2509, but is an important step.

- **Always redirect to frontend after login**
- **Toast on login when query param present**


Abradat (approved):
LGTM!

PR #2111, opened 2025-02-06
2523: Fix view condition and condition card display issues

trslater:
- **Create single date if single date and missing
   - On loading condition view, create a date if none exist
   - Applies to apps, NOI's
   - Backend service now allows for creating a single date 
   - Added an error if trying to create a date on a condition that doesn't   allow it.**
- **Fix runtime errors from missing decision**
- **Fix condition card display issue when date missing**


Abradat (approved):
LGTM!

PR #2112, opened 2025-02-06
ALCS-2468 Remove the individual date method

fbarreta:
Since we are hard deleting the conditions making it an empty array, there's no need to handle the dates individually anymore. TypeORM is making the deletion when saving the decision entity.


Abradat (approved):
LGTM!

PR #2122, opened 2025-02-12
ALCS-2514: Fix dockerfile warnings

dylanrogowsky-oxd:
Address the warnings in [ALCS-2514](https://apps.nrs.gov.bc.ca/int/jira/browse/ALCS-2514):

```
344 - FromAsCasing: 'As' and 'FROM' keywords' casing do not match (line 6)
345 - FromAsCasing: 'As' and 'FROM' keywords' casing do not match (line 22)
346 - FromAsCasing: 'As' and 'FROM' keywords' casing do not match (line 43)
347 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 55) 
```

### DIFF
--- a/alcs-frontend/Dockerfile
+++ b/alcs-frontend/Dockerfile
@@ -13,7 +13,7 @@ RUN npm ci
 # Copy the source code to the /app directory
 COPY . .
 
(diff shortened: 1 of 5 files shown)


Abradat (approved):
LGTM!

PR #2125, opened 2025-02-13
ALCS-2081 Remove active days for conditions and conditions subtasks

fbarreta:
Remove active days for conditions and conditions subtasks, as defined on a breakout with Ansley and Nowshin.


trslater (approved):
LGTM 🚀

PR #2129, opened 2025-02-14
2546: Pass submitted date to upstream submission methods

trslater:
- This helps make sure notification, PDF, and email all have the same
  submitted date
- It also allows them to execute in any order


fbarreta (approved):
LGTM

PR #2132, opened 2025-02-18
Update "Go to Card" button after creating condition card

Abradat:
Update "Go to Card" button to include condition cards in the dropdown once the condition card is created for applications and NOIs


trslater (approved):
LGTM 🚀

PR #2136, opened 2025-02-19
ALCS-1987 Custom ng-select adjustments

fbarreta:
Match the ng-select disabled appearance with the other inputs.

image


Abradat (approved):
LGTM!

PR #2137, opened 2025-02-19
Hotfix: Fix modification and reconsideration pills on condition cards

Abradat:
Show modification and reconsideration pills on application and NOI condition cards when the file has modification or reconsideration request instead of the decision


fbarreta (approved):
LGTM

PR #2140, opened 2025-02-20
ALCS-2560 Change the search to APP/NOI File number

fbarreta:
Get MODI and RECON pills based on app or noi file number for conditions and subtasks.


Abradat (approved):
LGTM

PR #2144, opened 2025-02-21
ALCS-2390: Remove "Use End Date" Field from Application and NOI Prep

Abradat:
- "Use End Date" from NARU, NFU, POFO, and ROSO applications
- "Use End Date" from POFO and ROSO NOIs
- "Use End Date (Placement of Fill/Removal of Soil)" from PFRS applications and NOIs


fbarreta (approved):
LGTM

PR #2147, opened 2025-02-24
ALCS-2260: QA Fixes #1: Fix position and alignment of quick filters

trslater:
- Move quick filters before "modified by" line
- Adjust margins to better match designs


Abradat (approved):
LGTM!

PR #2151, opened 2025-02-25
ALCS-2260: QA #2: Fix complete filter and update status properly

trslater:
- There was a typo making complete filter fail
- When the status was updated in the single condition component, the change was not being propogated to the list condition component's data, so when the list redrew the single components, it used its old data
   - Added an event emitted to update the list status when updating the single status


Abradat (approved):
LGTM!

PR #2152, opened 2025-02-25
ALCS-2476: Disable deploy concurrency

dylanrogowsky-oxd:
By adding the concurrency key in both the main CD workflow (cd.yml) and the deploy workflow (deploy.yml), we ensure that:

- A new deployment cancels any in-progress deployment for the same environment
- Only one deployment per environment (test, prod) runs at a time
- No outdated deployments remain stuck in a pending/waiting for approval state
- This prevents long-running or stuck (waiting for approval) deployments, which will eventually timeout

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -12,9 +12,15 @@ jobs:
     with:
       environment: test
     secrets: inherit
+    concurrency: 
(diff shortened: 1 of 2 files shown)


trslater (approved):
LGTM 🚀

PR #2158, opened 2025-02-26
ALCS-2078: NOI Decision Condition Financial Instrument

Abradat:
- Add frontend and backend components for NOI decision condition financial instruments
- Fix financial instrument common table's status column rendering received date instead of status date


fbarreta (approved):
LGTM

PR #2159, opened 2025-02-26
ALCS-2579: Make shared upload dialog more robust and performant

trslater:
- Create interfaces for setting data to allow for
  better validation
- Create a global options consts that can be used
  for both add and edit to make sure they are
  consistent
- Pass in parcel/owner services to avoid long
  modal opening times


Abradat (approved):
LGTM!

PR #2161, opened 2025-02-27
ALCS-2579: QA #1: Don't overwrite options object

trslater:
- Silly mistake
- I forgot `Object.assign()` overwrites the target object
- Replaced with spread operators


Abradat (approved):
LGTM!

PR #2162, opened 2025-02-27
Feature/alcs 2029 sort conditions qa#1

fbarreta:
QA changes:

- Add the new condition at the bottom of the array
- Work with a copy of the array on order dialog and only apply to original array, if the save button is clicked


Abradat (approved):
LGTM!

trslater (approved):
LGTM 🚀

PR #2165, opened 2025-02-27
ALCS-2548: Redirect to decision tab for modification and reconsideration cards

Abradat:
Redirect to the decision tab by clicking on the "VIEW DETAIL" button for:
- Application modification and reconsideration cards
- NOI modification cards


trslater (approved):
LGTM 🚀

PR #2168, opened 2025-02-27
ALCS-2533: Reconfigure OWASP ZAP scans

dylanrogowsky-oxd:
- Combine the previous API and Frontend scan workflows into a single workflow
- Add the portal scan
- Point the API scan to the swagger page, which the `action-api-scan` documentation suggests


trslater (approved):
LGTM 🚀

PR #2175, opened 2025-02-28
ALCS-2029 Change move up algorithm

fbarreta:
The move up algorithm was switching the select record with the top one.
I changed the behavior bringing the selected row to the top, and bringing the subsequent ones down.


Abradat (approved):
LGTM!

PR #2187, opened 2025-03-03
Add confirmation dialog for deleting dates in multiple date conditions

Abradat:
The change applies for applications and NOIs


fbarreta (approved):
LGTM

PR #2189, opened 2025-03-04
ALCS-2576 Implement alpha index on condition card

fbarreta:
Implemented the alpha index on the conditions card

- If there's no order recorded, suppress the index

Fix was made while adding a new condition to record the current order.


trslater (approved):
LGTM 🚀

PR #2190, opened 2025-03-04
ALCS-2576 Avoid order gaps when removing conditions

fbarreta:
Fix to avoid gaps when removing a condition from draft.


Abradat (approved):
LGTM

PR #2199, opened 2025-03-06
Rename Financial Instrument Data Type Enums and Add Default to Status Column

Abradat:
- Fix financial instruments' enum data types' names being truncated due to PostgreSQL's 63-byte limit for names. Applies to both applications and NOIs.
- Add default value to application financial instrument's status column


trslater (approved):
LGTM 🚀

PR #2208, opened 2025-03-11
ALCS-2603: Fix move to top/bottom

trslater:
- Ordering was relying on UUID which doesn't exist if conditions haven't been saved
- Changed to use only order property, index not needed
- Also simplified updating order and sorting


Abradat (approved):
LGTM

fbarreta (approved):
LGTM

PR #2210, opened 2025-03-11
ALCS-2603 QA #1: Replace condition that kept sane order numbers

trslater:
- I had taken out a condition that I thought wasn't necessary, because it relied on the index which was causing the initial problem
- The condition prevented conditions that didn't move from having their orders updated
- I've put it back and it is now based on the record order, which should accomplish the same result


Abradat (approved):
LGTM

PR #2212, opened 2025-03-12
Add Decision Meeting Banner to Application Condition Card

Abradat:
Show next decision meeting banner on application condition card when it is on "Ready for review sent / Going to next review discussion" column and not in "Application Conditions" board.


fbarreta (changes requested):

fbarreta (approved):
LGTM

PR #2214, opened 2025-03-12
ALCS-2528 Add Return Application button

fbarreta:
Move Start/Continue Application button outside the tabs, and add Return Application button


Abradat (approved):
LGTM!

PR #2217, opened 2025-03-13
ALCS-2356: Filter tags by category when when adding tags and in advanced search

trslater:
- **Filter tags by category in search**
- **Filter tags by category when adding tags**


Abradat (approved):
LGTM!

PR #2226, opened 2025-03-13
ALCS-2561 Fix re-enabling the fields

fbarreta:
Once the field was changed, it was only enabled again in case of an error.
The desired behavior was to enable once the processing is over.


Abradat (approved):
LGTM!

PR #2229, opened 2025-03-16
ALCS-2601: Adjust nginx config for WAVA scan results

dylanrogowsky-oxd:
Adjust nginx config to resolve 3 WAVA scan issues


Abradat (approved):
LGTM!

PR #2237, opened 2025-03-17
ALCS-2455: Allow editing resolution numbers

trslater:
- Had to update shared inline number component:
   - Add option to hide buttons, so component can be controlled by parent
   - Add option to disable thousands separator, since this doesn't make sense for resolution numbers
   - Add ability to set async validators
   - Use number type internally
      - The component used strings, but the internal Angular component used numbers, so the value was changing from string to number while typing
- Make resolution number editable:
   - Add inline editing component
   - Modify the way decision/pos-decision observables work to avoid flashes of old value
   - Add buttons to control inline editing component
   - Hack to position error messages
   - Created a new backend/frontend services/controller for checking if a resolution number exists
   - Add an async validator to check for existing resolution numbers on change


Abradat (approved):
LGTM!

PR #2249, opened 2025-03-20
ALCS-2455: Don't show resolution number save error when re-editing

trslater:
- When the resolution number editor is open and a release is attempted, it shows an error
- Now, when we save or cancel the resolution editor, and then re-edit it, the error is gone


fbarreta (approved):
LGTM

PR #2252, opened 2025-03-21
ALCS-881 QA #1: Fix user being able to delete mid chain decision after revert to draft

Abradat:
- Applies to applications and NOIs
- Add canBeDeleted field to application and NOI decision DTO
- refactor post decision components to refresh reconsiderations and modifications


fbarreta (approved):
LGTM

PR #2253, opened 2025-03-21
ALCS-2628: Fix paused and active days numbers misalignment on decision

Abradat:
Apply to applications and NOIs


fbarreta (approved):
LGTM

PR #2281, opened 2025-03-25
Feature/alcs 2461 option to send cancel email

fbarreta:
Add option to send or not email on Application and NOI cancelations

### DIFF
--- a/services/apps/alcs/src/alcs/application/application.controller.ts
+++ b/services/apps/alcs/src/alcs/application/application.controller.ts
@@ -10,6 +10,7 @@ import {
   Param,
   Patch,
   Post,
+  Query,
   Req,
   UseGuards,
 } from '@nestjs/common';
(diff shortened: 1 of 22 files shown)


services/apps/alcs/src/alcs/application/application.controller.ts
```diff
@@ -105,11 +106,17 @@ export class ApplicationController {
 
   @Post('/:fileNumber/cancel')
   @UserRoles(...ROLES_ALLOWED_APPLICATIONS)
-  async cancel(@Param('fileNumber') fileNumber): Promise<void> {
+  async cancel(@Param('fileNumber') fileNumber, @Query('sendEmail') sendEmail): Promise<void> {
```
Abradat, line 109: What do you think about having `sendEmail` in the request body? In my view, `sendEmail` is about what action to perform rather than how the action is performed like filters and sorting. I'd be interested in @trslater input as well.

Abradat (approved):
LGTM!

PR #2284, opened 2025-03-25
ALCS-2638: Sanitize fileName when uploading document

Abradat:
- This pull request includes changes to the `saveUploadedFile` method in two different controllers to handle specific characters in file names. The main change is the replacement of en-dash and em-dash characters with hyphens in file names. 
- Applies to applications, NOIs, notifications, planning reviews, and inquiries on ALCS


trslater (approved):
LGTM 🚀

PR #2286, opened 2025-03-25
ALCS-2651 Add display index as parameter

fbarreta:
Do not display the index on "View Conditions" if the order is not set.


trslater (approved):
LGTM 🚀

PR #2293, opened 2025-03-27
ALCS-2542: Implement new inline document links and replace all ALCS links

trslater:
- Implement new document open system
   - Add a new Angular route
   - Add a new component as target for the route
      - The component view is initially a loading spinner with "Loading document..."
      - The component grabs the document UUID, fetches the S3 URL, and loads it into an object
   - Controllers, DTO's, and services needed to be updated to make sure that <file>-documents had a documentUuid, and that there was a shared endpoint for retreiving the S3 URL
- Replace all existing click handlers with router links to the new route


fbarreta (approved):
LGTM

PR #2301, opened 2025-03-28
ALCS-2657 Added disable behavior on incorrect status

fbarreta:
Display the return button when the statuses are:

- IN_REVIEW_BY_LG
- RETURNED_TO_LG
- SUBMITTED_TO_LG

But disable and show tooltip if it is:

- SUBMITTED_TO_LG


trslater (approved):
LGTM 🚀

PR #2303, opened 2025-03-31
ALCS-2212: Improve advanced search name field

trslater:
- Change the way the search string is formatted
- It is now a single string and the full name, first name, and last name
  are compared against this string
- Use different concat method, so there is no space when first or last
  name are missing


fbarreta (approved):
LGTM

PR #2338, opened 2025-04-28
Allow public docs to load when not authenticated

trslater:


services/apps/alcs/src/document/document.service.ts
```diff
@@ -156,6 +158,49 @@ export class DocumentService {
...
+      from
+        alcs.notification_document nd2
+      where 
+        nd2.document_uuid = $1
+      `,
+      [uuid],
+    );
+
+    if (fileDocuments.length > 1) {
+      console.log('issue');
```
PaulGarewal, line 194: Not sure the intention here for the console log if multiple docs are found

PaulGarewal: Other than that looks great to me

trslater: Good catch. I meant to fix this once I figured out what the best course of action was, but forgot. I'll fix it now.

trslater: K, I pushed a change.

PaulGarewal (approved):
Look great to me 😄

PR #2353, opened 2025-05-07
Fix public/portal document links

trslater:
- Check decision docs as well as file docs when checking if document is public
- Fix document file loader component names
- Fix portal file loader layout
- Fix ALC review document links according to Excel document (extension of table in linked issue)

### DIFF
--- a/alcs-frontend/src/app/shared/document-file-loader/document-file-loader.component.ts
+++ b/alcs-frontend/src/app/shared/document-file-loader/document-file-loader.component.ts
@@ -3,7 +3,7 @@ import { ActivatedRoute } from '@angular/router';
 import { DocumentService } from '../../services/common/document/document.service';
 
 @Component({
-  selector: 'app-temp',
+  selector: 'document-file-loader',
   templateUrl: './document-file-loader.component.html',
   styleUrl: './document-file-loader.component.scss',
 })
--- a/portal-frontend/src/app/features/public/application/alc-review/decisions/decisions.component.html
+++ b/portal-frontend/src/app/features/public/application/alc-review/decisions/decisions.component.html
@@ -23,18 +23,18 @@
           <a routerLink="/document/{{ document.documentUuid }}" target="_blank">{{ document.fileName }}</a>
           &nbsp;({{ document.fileSize | filesize }})
         </div>
-        <button class="center" mat-button (click)="openFile(document)">
+        <button class="center" mat-button (click)="downloadFile(document.documentUuid)">
           <mat-icon>file_download</mat-icon>
           Download
         </button>
       </div>
       <div class="document responsive left" *ngFor="let document of decision.documents">
         <div class="document-name">
-          <a routerLink="/document/{{ document.documentUuid }}" target="_blank">{{ document.fileName }}</a>
+          <a (click)="downloadFile(document.documentUuid)">{{ document.fileName }}</a>
         </div>
         <div class="split">
           <div>{{ document.fileSize | filesize }}</div>
-          <button class="center" mat-button (click)="openFile(document)">
+          <button class="center" mat-button (click)="downloadFile(document.documentUuid)">
             <mat-icon>file_download</mat-icon>
             Download
           </button>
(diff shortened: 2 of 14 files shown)


lunamoonmoon (dismissed):
lgtm!

portal-frontend/src/app/features/applications/application-details/incl-details/incl-details.component.html
```diff
@@ -52,29 +53,26 @@
...
       <app-no-data [showRequired]="showErrors" *ngIf="proofOfSignage.length === 0"></app-no-data>
     </div>
 
     <div class="subheading2 grid-1">Report of Public Hearing</div>
     <div class="grid-double">
       <div *ngFor="let file of reportOfPublicHearing">
-        <a routerLink="/document/{{ file.documentUuid }}" target="_blank">
-          {{ file.fileName }}
-        </a>
+        <a class="mobile" (click)="downloadFile(file.documentUuid)">{{ file.fileName }}</a>
```
PaulGarewal, line 74: we change the way we conditionally show mobile/desktop links, is there a reason? (ngIf vs class="mobile") (just some of the first things I am seeing.

portal-frontend/src/app/features/applications/application-details/incl-details/incl-details.component.scss
```diff
@@ -0,0 +1,15 @@
+@use '../../../../../styles/functions' as *;
+
+.mobile {
+  @media screen and (min-width: $tabletBreakpoint) {
```
PaulGarewal, line 4: Is this the best way to determine mobile vs desktop(screen width)? Should there be an OS/Browser detection system?

portal-frontend/src/app/features/applications/application-details/cove-details/cove-details.component.ts
```diff
@@ -55,11 +55,10 @@ export class CoveDetailsComponent {
...
 
-  async openFile(file: ApplicationDocumentDto) {
-    const res = await this.applicationDocumentService.openFile(file.uuid);
-    if (res) {
-      openFileInline(res.url, file.fileName);
-    }
+  async downloadFile(uuid: string) {
+    const { url, fileName } = await this.documentService.getDownloadUrlAndFileName(uuid, false, true);
+
+    downloadFile(url, fileName);
```
PaulGarewal, line 61: Could possibly add some error handling here just in case

portal-frontend/src/app/features/applications/application-details/incl-details/incl-details.component.html
trslater: The only reason was that I started doing the CSS classes first, but some of these components already were using `isMobile`.

portal-frontend/src/app/features/applications/application-details/incl-details/incl-details.component.scss
trslater: Mostly that this is the established way the app worked when I started. Changing it now would be a large task.

I'm not sure it would be advantageous to anyway. Admittedly, I haven't looked into this in a while, so maybe this has changed, but I believe browser/device sniffing is very unreliable and using breakpoints was the standard practice last I checked.

portal-frontend/src/app/features/applications/application-details/cove-details/cove-details.component.ts
trslater: That's probably a good idea.

trslater: Done.

portal-frontend/src/app/features/applications/application-details/incl-details/incl-details.component.scss
PaulGarewal: Ah ok that's a big deal and probably not worth it, good to know as I slowly learn about this app. Resolving 😄

portal-frontend/src/app/features/applications/application-details/incl-details/incl-details.component.html
PaulGarewal: Sounds good, something I picked up on but isn't a big deal at all especially if these components were already using isMobile.

PaulGarewal (approved):
LGTM, appreciate all the explanations 👍

PR #2424, opened 2025-06-26
Set path from environment when updating images

trslater:


raarielgrace:
PR looks good but the playwright tests are failing. I don't believe that happened before.

raarielgrace (approved):

PR #2433, opened 2025-07-04
Fix incorrectly escaped whitespace character

trslater:
Apparently this is a high risk security vulnerability


lunamoonmoon (approved):
So small! looks good

PR #2455, opened 2025-07-21
2393 feat: C&E property component

PaulGreywal:
- Frontend portion of the Property component of ticket 2393 and very small portion of the backend work. 
- Backend work almost at completion, working on save to draft

### DIFF
--- a/alcs-frontend/src/app/features/compliance-and-enforcement/property/property.component.ts
+++ b/alcs-frontend/src/app/features/compliance-and-enforcement/property/property.component.ts
@@ -29,8 +29,9 @@ export function cleanPropertyUpdate(update: UpdateComplianceAndEnforcementProper
   const cleaned: any = {};
   for (const [key, value] of Object.entries(update)) {
     if (
-      value !== null &&
       value !== undefined &&
+      // Allow null values for pid/pin so they can be cleared
+      (value !== null || key === 'pid' || key === 'pin') &&
       // Filter out default values
       !(key === 'ownershipTypeCode' && value === 'SMPL')
     ) {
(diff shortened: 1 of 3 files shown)


alcs-frontend/src/app/features/compliance-and-enforcement/property/property.component.ts
```diff
@@ -0,0 +1,169 @@
...
+export class PropertyComponent implements OnDestroy {
+  $destroy = new Subject<void>();
+
+  isPatching = false;
+  isSubscribed = false;
+
+  PARCEL_OWNERSHIP_TYPE = PARCEL_OWNERSHIP_TYPE;
+  localGovernments: ApplicationLocalGovernmentDto[] = [];
+  filteredLocalGovernments: ApplicationLocalGovernmentDto[] = [];
+  regions = [
```
trslater, line 31: We usually do this sort of thing with an enum. Something like:

```ts
enum REGION {
  ISLA = 'Island',
  ...
}
```

alcs-frontend/src/app/services/compliance-and-enforcement/compliance-and-enforcement.service.ts
```diff
@@ -25,10 +25,11 @@ export class ComplianceAndEnforcementService {
   async create(
     updateDto: UpdateComplianceAndEnforcementDto,
     createInitialSubmitter = false,
+    createInitialProperty = false,
   ): Promise<ComplianceAndEnforcementDto> {
     return await firstValueFrom(
       this.http.post<ComplianceAndEnforcementDto>(
-        `${this.url}?createInitialSubmitter=${createInitialSubmitter}`,
+        `${this.url}?createInitialSubmitter=${createInitialSubmitter}&createInitialProperty=${createInitialProperty}`,
```
trslater, line 32: 100% up to you if you want to bother changing this. This works fine as is, and I obviously did it this way to start, but I recently discovered what I think is a bit nicer way to handle this. Here's what I did for submission docs:

```ts
let params = new HttpParams();

if (fileNumber) {
  params = params.set('fileNumber', fileNumber);
}
for (const typeCode of typeCodes) {
  params = params.append('typeCodes', typeCode);
}

return firstValueFrom(this.http.get<ComplianceAndEnforcementDocumentDto[]>(this.url, { params }));
```

trslater (approved):

PR #2464, opened 2025-07-25
2460 fix: updated axios and jsom to address vulnerabilities 

PaulGreywal:
- Updated axios and jsdom, axios will use form-data 4.0.4 to be secure, while jsdom eliminates form-data dependency entirely
- Tested app to make sure there were no breaking changes, suggestions here will be the most beneficial (especially with document upload features)
- Please checkout this branch and run npm with a cleared cache to test out the changes in both the frontend and backend services.

### DIFF
--- a/alcs-frontend/package-lock.json
+++ b/alcs-frontend/package-lock.json
@@ -3240,9 +3240,9 @@
       "license": "Python-2.0"
     },
     "node_modules/@eslint/eslintrc/node_modules/brace-expansion": {
-      "version": "1.1.11",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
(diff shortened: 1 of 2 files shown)


raarielgrace (dismissed):

services/package.json
```diff
@@ -36,7 +36,7 @@
     "@fastify/static": "^7.0.3",
     "@grpc/grpc-js": "^1.13.4",
     "@grpc/proto-loader": "^0.8.0",
-    "@nestjs/axios": "^3.1.3",
```
trslater, line 39: Might this introduce breaking changes?

alcs-frontend/package.json
```diff
@@ -32,6 +32,7 @@
     "@ng-select/ng-option-highlight": "^12.0.6",
     "@types/validator": "^13.15.2",
     "angular-mentions": "^1.5.0",
+    "jsdom": "^26.1.0",
```
trslater, line 35: Why are we adding a new package? Is this a dependency of another package where we want to change the version?

PR #2472, opened 2025-07-30
Complaint/referral doc QA 1

trslater:
- Fix double prepending ORCS tag prefix
- Remove complainant as source option
- Add table sorting from headers

### DIFF
--- a/services/apps/alcs/src/document/document.service.ts
+++ b/services/apps/alcs/src/document/document.service.ts
@@ -73,7 +73,7 @@ export class DocumentService {
       fileName,
       source,
       system,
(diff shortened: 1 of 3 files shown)


services/apps/alcs/src/document/document.service.ts
```diff
@@ -73,7 +73,7 @@ export class DocumentService {
       fileName,
       source,
       system,
-      tags: this.formatTagsForDb(tags ?? DEFAULT_DOC_TAGS),
```
trslater, line 76: `formatTagsForDb` is already applied within `createDocumentRecord`.

PR #2597, opened 2025-10-31
2585 Security Issues

PaulGreywal:
- updates of packages to address security issues.


portal-frontend/nginx.conf
```diff
@@ -11,6 +11,7 @@ http {
 
     server_tokens off;
     proxy_cookie_path / "/; secure; HttpOnly; SameSite=lax";
+    proxy_cookie_flags ~ secure HttpOnly SameSite=lax;
```
PaulGreywal: This file is where input is needed, I probably won't commit this. It is to address the OWASP ZAP scan, which is associated with the cookies in the keycloak login.

PR #2633, opened 2025-11-19
2475: Add chronology entry QA 1

trslater:
- Fix corners
- Remove default public from docs
- Persist newlines in description view
- Change error handling design
  - Complete always enabled
  - Show errors on click complete


PaulGreywal (approved):
LGTM

PR #2644, opened 2025-11-21
fix: resolved lat long zero bug

PaulGreywal:
- Resolved issue of the user being unable to input zero decimal places in the lat long fields
- Users can now place integers and short decimals, e.g. 49 -> 49.00000, 49.12 -> 49.12000
- Users will still throw an error message if going over 5 decimal places in inputs


alcs-frontend/src/app/features/compliance-and-enforcement/property/property.component.spec.ts
```diff
@@ -75,4 +77,645 @@ describe('PropertyComponent', () => {
     component.parentForm = parentForm;
     expect(parentForm.contains('property')).toBe(true);
   });
+
+  describe('Latitude/Longitude Validation', () => {
```
PaulGreywal: Super long tests needed to pass the sonar coverage

Issue #2606, opened 2025-11-06
Set author of chronology entry


ancharb:
As the officer, I want the chronology draft to automatically recognize me as the author when I create it, but then allow me to change the author if needed.

**Acceptance Criteria**

- [ ] Add a drop-down on the chronology entry draft for the author
  - [ ] on create, fill this with the user's name
  - [ ] only users with the C&E role are available in the drop-down 
  - [ ] if the user interacts with the drop-down, keep their selection moving forward, unless the user interacts with the drop-down again
- [ ] Show the author on the saved View

ancharb:
great work!

Issue #2500, opened 2025-08-26
Close chronology on C&E file


ancharb:
### MUST BE COMPLETED AFTER #2475 

As the C&E officer, I want to provide the reasons why I'm closing the file, then I want to close the chronology so that no one else adds an entry and so that Isha knows to audit and change the status to 'Closed' on Overview tab. If I want to add something to the chronology later, without changing the status back to Open, I should still be able to do that.

**Acceptance Criteria**

- [x] Provide option to 'Close Chronology' but can also be 'Re-Open Chronology' if closed
  - [x] if there are no chronology entries, hide the Close button
  - [x] if a chronology entry is being drafted, disable the Close button
- [x] Provide a confirmation dialog when user clicks 'Close Chronology' with a reminder that user has to add a final chronology entry before they close it
- [x] Once chronology is closed
  - [x] add a final entry to the chronology that has the user's name and date of the closure action
  - [x] disappear the '+ Entry' button, and disable the Edit pencil on existing entries
  - [x] change 'Close Chronology' button label to 'Re-Open Chronology', provide confirmation dialog
  - [x] Disable entry edit buttons
- [x] Once chronology is re-opened
  - [x] disappear the final entry that had the closure info
  - [x] Re-Open button switches back to 'Close Chronology'
  - [x] show the +Entry button

Figma: https://www.figma.com/design/5qXmbgyOL5M8PO7m7oqCeR/ALCS?node-id=6788-88737&t=mp2kTKoILNpIlvWu-1

ancharb:
great work!

PR #2648, opened 2025-11-25
2500/2606: Close chronology on C&E file

trslater:
This PR is for #2500 and #2606.

The way I handled tracking the closed state of the chronology in the DB requires some explaining: instead of adding a new entry for the final entry, I simply track the person closing and the closed date (nullable) directly on the C&E file. A chronology is closed if it has a closed date. It is open if the closed date is null.

Rationale: adding a mostly empty entry with a lot of conditional logic for the UI adds complexity, and adds no real value. It also would require a way to signify this special entry in some way, which again adds complexity.

- Added new columns, including DTO's, services, entities, migrations, and automappers:
  - C&E file:
    - `chronology_closed_by`: user who closed the chronology
    - `chronology_closed_at`: date chronology was closed
  - Chronology entry:
    - `author`
 - Modified confirmation dialog service to have "styles"
   - The first style is `WARN`: this adds a reddish warning box, bold "warning", and info icon around message
   - This has come up quite a few times, so this is an easy, backwards compatible way to add this style in the future, without needing a bespoke confirmation dialog
 - Extra: cleaned up C&E file backend fetch service method

### DIFF
--- a/services/apps/alcs/src/providers/typeorm/migrations/1763683662058-add_chronology_entry_author.ts
+++ b/services/apps/alcs/src/providers/typeorm/migrations/1763683662058-add_chronology_entry_author.ts
@@ -37,7 +37,9 @@ export class AddChronologyEntryAuthor1763683662058 implements MigrationInterface
       FROM
         inserted_users`,
     );
-    await queryRunner.query(`ALTER TABLE "alcs"."compliance_and_enforcement_chronology_entry" ADD "author_uuid" uuid`);
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."compliance_and_enforcement_chronology_entry" ALTER COLUMN "author_uuid" SET NOT NULL`,
+    );
     await queryRunner.query(
       `ALTER TABLE "alcs"."compliance_and_enforcement_chronology_entry" DROP COLUMN "nris_inspection_id"`,
     );


alcs-frontend/src/app/features/compliance-and-enforcement/details/chronology/chronology.component.ts
```diff
@@ -12,7 +12,7 @@ import {
   ComplianceAndEnforcementChronologyEntryDto,
   UpdateComplianceAndEnforcementChronologyEntryDto,
 } from '../../../../services/compliance-and-enforcement/chronology/chronology.dto';
-import { DOCUMENT_SOURCE, DOCUMENT_TYPE } from '../../../../shared/document/document.dto';
+import { DOCUMENT_TYPE } from '../../../../shared/document/document.dto';
```
trslater: This should have been reabsed from develop, but for some reason it wasn't

services/apps/alcs/src/alcs/compliance-and-enforcement/compliance-and-enforcement.service.ts
```diff
@@ -45,18 +45,24 @@ export class ComplianceAndEnforcementService {
     return this.mapper.mapArray(entity, ComplianceAndEnforcement, ComplianceAndEnforcementDto);
   }
 
-  async fetchByUuid(
-    uuid: string,
+  async fetchById(
```
trslater: This is unrelated to ticket, just cleaning this up. There was a lot of duplicated code here, and they had mismatches. `fetchByUuid` was never used.

alcs-frontend/src/app/features/compliance-and-enforcement/details/chronology/entry/entry.component.ts
```diff
@@ -89,6 +93,7 @@ export class ComplianceAndEnforcementChronologyEntryComponent implements OnInit,
     this.form.patchValue(
       {
         date: entry.date !== undefined && entry.date !== null ? moment(entry.date) : null,
+        authorUuid: entry.author.uuid,
```
PaulGreywal: I think authorUuid can be put here and elsewhere, could be included in line 83-84 too. Unless I am misunderstanding something, it should be included in the payload when user's are making the draft and could change the author potentially.

services/apps/alcs/src/providers/typeorm/migrations/1763683662058-add_chronology_entry_author.ts
```diff
@@ -0,0 +1,67 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddChronologyEntryAuthor1763683662058 implements MigrationInterface {
+  name = 'AddChronologyEntryAuthor1763683662058';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "alcs"."compliance_and_enforcement_chronology_entry" ADD "author_uuid" uuid NOT NULL`,
```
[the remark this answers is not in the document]
trslater: This is a tricky one, because we obviously need to be able to have not null columns. Since the whole C&E feature set is new, we could consider this a foundational migration. We'd just need to make sure the DB is empty before deployment.

PR #2695, opened 2025-12-18
2618 Document Upload QA

PaulGreywal:
- fix from last document change to not validate parcel if there is no data in parcel service or the length of the selectable parcels is less than 1


alcs-frontend/src/app/shared/document-upload-dialog/document-upload-dialog.component.ts
```diff
@@ -270,19 +270,26 @@ export class DocumentUploadDialogComponent implements OnInit, OnDestroy {
...
 
     this.selectableParcels = await this.data.parcelService.fetchParcels(this.data.fileId);
 
     if (this.selectableParcels.length < 1) {
+      // No parcels available, we will not require it
+      this.parcelId.clearValidators();
+      this.parcelId.updateValueAndValidity();
       return;
     }
+
```
PaulGreywal: moved validation check and set to outside the if statements (where the parcels are null or not fetchable, thus not requiring validation in those blocks)

PR #2706, opened 2025-12-23
2649: Upgrade portal Angular from 17 to 20

trslater:


trslater:
Looks good to me. Could use a last look from @PaulGreywal

PR #2722, opened 2026-01-08
2649 Remaining security fixes

trslater:
I've gone through all Trivy and dependabot security alerts. Most were NPM fixes. Most of these were fixed by Angular updates or normal dependabot updates. There were a few I needed to update myself. The only non-npm issue was libpng. This should be resolved by the new image being used: node:24-apline. The Node 24 image uses a newer version of Alpine, which should include the patched libpng.

As for ACS, I've upgraded the API Node version to 24 as well. This should fix most issues for the API, but at the time of writing, all Node images had a critical and a few high security vulnerabilities: https://hub.docker.com/_/node/tags

### DIFF
--- a/services/package-lock.json
+++ b/services/package-lock.json
@@ -91,93 +91,6 @@
       }
     },
     "node_modules/@angular-devkit/core": {
-      "version": "17.3.11",
-      "resolved": "https://registry.npmjs.org/@angular-devkit/core/-/core-17.3.11.tgz",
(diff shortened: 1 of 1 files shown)


services/Dockerfile
```diff
@@ -3,7 +3,7 @@
 # https://www.tomray.dev/nestjs-docker-production
 ###################
 
-FROM node:20-alpine AS development
+FROM node:24-alpine AS development
```
trslater: This image uses a newer version of Alpine, which should patch a lot of the vulnerable binaries.

PR #2727, opened 2026-01-12
Fix E2E tests

trslater:


PaulGreywal (approved):
LGTM, one small comment but does not change my approval.

PR #2830, opened 2026-03-05
2518: ETL Submitter Data

trslater:
- **Fail on conflict**
- **Fix num failed**
- **Add import all help text**
- **Remove never used user**
- **Add submitter ETL**
  - CLI command
  - module
  - function
  - SQL files
- **Add C&E file obfuscation**
- **Add submitter obfuscation**

### DIFF
--- a/bin/migrate-nris-data/migrate.py
+++ b/bin/migrate-nris-data/migrate.py
@@ -67,7 +67,7 @@ def obfuscate(ctx, batch_size):
     if ctx.invoked_subcommand is None:
         # Order matters
         ctx.invoke(obfuscate_ce_files)
-        # ctx.invoke(obfuscate_submitters)
+        ctx.invoke(obfuscate_submitters)
 
 
 @obfuscate.command("ce-files")


bin/migrate-nris-data/ce_files/ce_files.py
```diff
@@ -1,12 +1,11 @@
 from common.etl_logger import setup_and_get_logger
 from config import ABS_PATH
-from db import batch_etl, ensure_ce_users
+from db import batch_etl
 
 
 def ce_file_etl(batch_size):
     logger = setup_and_get_logger("ce-files")
 
-    ensure_ce_users(logger)
```
trslater: This didn't end up being used.

bin/migrate-nris-data/db.py
```diff
@@ -109,8 +109,7 @@ def batch_etl(
                 except Exception as err:
                     logger.exception(err)
                     write_conn.rollback()
-                    num_failed = count_total - num_successful
```
trslater: This was total count which wasn't that useful, since we're given total succeeded out of total expected at the end.

PR #3153, opened 2026-08-25
2676 QA 1: Only use property owners in search results

### LINKED ISSUE #2676: Add C&E IDs to top nav search, add C&E tab on Advanced results
As a C&E officer, I want to search by an ID in the top nav so I can find my C&E file.

**Acceptance Criteria**

- [x] include C&E IDs in the top navigation search tool
- [x] if the user searches for a C&E ID that is unique, bring them directly to the file
- [x] add a tab for C&E results on the Advanced Search page -> tying in the other fields to search will happen in a future ticket after the MVP
- [x] if the user searches for an ID that is shared with another type of file, bring them to the advanced search page and scroll them to the results at the bottom
- [x] Add C&E file to advanced search type drop-down

NOTE: all users can find C&E files, but the view of the file (detailed view) will be limited by role as per #2669 

https://www.figma.com/design/5qXmbgyOL5M8PO7m7oqCeR/ALCS?node-id=8438-26297&t=FYWRRQOSNa87aIFh-1

trslater:
The discrepancy between the owner in the C&E search results and details header was due to the search including all responsible party types, where we only want owners.

Also, there was a dependency issue via dependabot with fastify/static


lunamoonmoon (approved):
Looks great!

PR #3156, opened 2026-08-26
Increase file number sequence to 300k

### LINKED ISSUE #2660: Assign new IDs to all ALCS files by incrementing from max NRIS ID
As an officer, I want to get a sense of how old this file is by its ID number, so I can place it in my mental map of files easier.
As an ALCS user, I want all files to share the same ID increment pattern so it's not confusing.

**Additional Context**
- ALCS IDs provide a (rough) historical timeline ex. OATS ids are under 80,000, ALCS IDs started at 100,000, IDs coming from NRIS will span 1 to >200,000 but with many gaps (because NRIS is a shared platform, ALC not the only business area drawing from ID pool)
- Since we will be interspersing the C&E IDs from NRIS into the ALCS ID field, the current function for grabbing new IDs would result in a lower number for a new C&E file. If nothing is done, the C&E numbers will jump back and we want to avoid that because it could confuse the evidentiary record going to the courts.
- So I want to keep incrementing the C&E file ID higher than the max number from NRIS.
- Moving forward -> have unique ids across ALCS even though past ids are non unique

**Decision**: To minimize collisions and complexity (as per #2859) we pick Option 2 - start incrementing **ALL ALCS IDs** above the max number ETL'd from NRIS (~260,000)

**Acceptance Criteria**
- [x] When the user creates a new file of any type in ALCS/Portal after the NRIS ETL, the ALCS ID should be higher than the max ID from NRIS
  - [x] Set min file number to 300k—should give lots of buffer

trslater:


services/apps/alcs/src/providers/typeorm/migrations/1787769695918-bump_file_numer_sequence_ahead.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class BumpFileNumerSequenceAhead1787769695918 implements MigrationInterface {
+  name = 'BumpFileNumerSequenceAhead1787769695918';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      SELECT setval('alcs_file_number_seq', 300000, false)
```
lunamoonmoon, line 8: Will the historical file number be kept under a different field or is it no longer needed at all?

lunamoonmoon (approved):

services/apps/alcs/src/providers/typeorm/migrations/1787769695918-bump_file_numer_sequence_ahead.ts
lunamoonmoon: Tristan addressed, this is for new files only

trslater: I probably worded this poorly. It doesn't affect existing file numbers. This will only set the new starting place for new file numbers.

